PR #71, opened 2014-01-07
Add browserify support

deathcap:
I wanted to use node-minecraft-protocol in the web browser via [browserify](http://browserify.org/), but it fails because of the ursa native library dependency:

    Error: module "../bin/ursaNative" not found from "node-minecraft-protocol/node_modules/ursa/lib/ursa.js"

So I added a browser-specific module entry point, which does not include any ursa-dependent functionality. This means auth/crypto/etc is unavailable in this environment, but you can still encode and decode packets.

(@roblabla has been working on replacing [ursa](https://github.com/Obvious/ursa) with [Forge](https://github.com/digitalbazaar/forge), pure JavaScript: https://github.com/roblabla/node-minecraft-protocol/tree/feature-purejsrsa-forge - which could also solve this unbrowserifyable natives problem. But at least for my purposes, only including lib/protocol when browserified was sufficient. Didn't need any of the decryption functionality.)

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,0 +1,5 @@
+
+module.exports = {
+  protocol: require('./lib/protocol')
+};
+
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   "engines": {
     "node": ">=0.8.16"
   },
+  "browser": "browser.js",
   "devDependencies": {
     "mocha": "~1.8.2",
     "mkdirp": "~0.3.4",


deathcap:
Looking into this again, still interested in this feature, updates since:

* [ursa](https://www.npmjs.com/package/ursa) is now maintained again (transferred ownership)
* node-minecraft-protocol falls back to [node-rsa](https://www.npmjs.com/package/node-rsa) if ursa is unavailable (cool)

unfortunately, even with ursa listed in `optionalDependencies` in the package.json, it fails to browserify:

```
200    3ms     1.17KB  /bundle.js ➞ ./node_modules/watchify ./index.coffee -d --debug
200   87ms      1.2KB  /crosshair4.png
200   87ms     7.08KB  /logo-white.png
404   10ms        12B  /favicon.ico
200  280ms       777B  /
Error: Cannot find module '../bin/ursaNative' from '/Users/admin/games/voxeljs/voxpopuli/node_modules/voxel-clientmc/node_modules/mineflayer/node_modules/minecraft-protocol/node_modules/ursa/lib'
  at /Users/admin/games/voxeljs/voxpopuli/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:43:25
  at load (/Users/admin/games/voxeljs/voxpopuli/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:99:43)
  at /Users/admin/games/voxeljs/voxpopuli/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:105:22
  at /Users/admin/games/voxeljs/voxpopuli/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
  at FSReqWrap.oncomplete (fs.js:99:15)
```

but since node-rsa is now available as a replacement for ursa, and [browserify](https://github.com/substack/node-browserify) supports module substitution for the browser in package.json, tried this:

```javascript
  "browser": {
    "ursa": "node-rsa"
  },
```

of course, node-rsa is not API-compatible with ursa so this doesn't work.. (pull out your `rsa-wrap.js` into a new module, to allow browser-based substitution?). Additionally the `superagent` module has trouble with loading `mime`, attempts to call `fs.readFileSync(file, 'ascii')' which is not available in browserify (and not covered by [brfs](http://npmjs.com/package/brfs) - it only can replace fixed file paths), not sure what is going on there since [superagent](https://www.npmjs.com/package/superagent) is advertised as supporting both Node.js and the browser (with the same API).

The original change in this pull request (only exposing `lib/protocol.js` in the browser, not pulling in any other dependencies) still works. Could this be merged in the meantime until the other features of node-minecraft-protocol (yggdrasil authentication, using superagent; protocol encryption, using ursa; etc.) are made to work the browser (if ever)? I only really am interested in the packet decoding/encoding.

(Tested against the packets-1.8 branch)

roblabla:
Shouldn't browserify ignore optionalDependencies if they fail to install ? This strikes me as a weird bug. But anyway, I'll move my rsa-wrapper in an npm module. ```ursa-purejs``` or something. I should've done that from the start.

I looked at superagent's source and couldn't find the offending line. Do you have the traceback ? I think you might want to open an issue on their side. 

Do note that the raw packet support in packets-1.8 branch is... well, completely broken. The packet format changed quite a lot in 1.8 (and might change again), and writeraw is very version-dependant. 

I'll merge this in the meantime, but I hope to find a better solution eventually. I don't like adding a browser field in the package.json because we don't actually support the whole library inside a browser, just a tiny fraction of it.

deathcap:
Thanks!

(about the yggdrasil/superagent browser issue, posted more details in https://gist.github.com/deathcap/898bec7260a0b0dbf5cd, but probably not worth fixing, since XHR will not be able to contact the Mojang auth servers, anyway, due to cross-origin restrictions - and the only way to fix this, without Mojang enabling CORS on their auth servers, which I doubt they will, is to use a CORS proxy, which introduces security issues since a third-party has to pass through the user credentials. for browser auth, a different solution is needed - in [wsmc](https://gist.github.com/deathcap/898bec7260a0b0dbf5cd), I'm experimenting with having the user login with the official client (authenticating to Mojang's servers as usual, online mode), then get a token to use for browser-based login, to solve these problems.)

roblabla:
It might be worth it asking grum about oauth. He said he wanted to add it a while ago... and then poof, nothing happened since.

PR #110, opened 2015-01-04
Don't write out packets if connection is ended

pcfreak30:
Don't write out packets if connection is ended, and mark as ended on end method

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -143,10 +143,14 @@ Client.prototype.connect = function(port, host) {
 
 Client.prototype.end = function(reason) {
   this._endReason = reason;
+  this.ended = true;
   this.socket.end();
 };
 
 Client.prototype.write = function(packetId, params) {
+  if(this.ended) {
+    return false;
+  }
   if (Array.isArray(packetId)) {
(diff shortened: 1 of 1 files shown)


wtfaremyinitials:
Looks good to me. @roblabla?

roblabla:
Thank you for your contribution. This won't quite work though : you'll want to remove the ```ended``` variable from the [setSocket function](https://github.com/pcfreak30/node-minecraft-protocol/blob/master/lib/client.js#L118), and change every occurrence of that variable to ```self.ended```.

pcfreak30:
From what I could tell, the "end" event is never called, so ended is never set from my debugging. I kept getting write after end errors when calling end() on the bot library. This prevented outbound writes once end is called and set before its actually closed.

roblabla:
end event is called whenever the TCP connection is lost, whether it is because the server shut it down, or because you lost your internet connection. So outbound writes won't work past that point either. Calling socket.end() will not call the end event, I guess, but that doesn't mean the event should still be ignored. 

Besides, the 'error' event also calls endSocket, and you definitly don't want writes after a TCP error ^^.

Issue #30, opened 2013-01-29
add tests for invalid data


roblabla:
I think it would be a good idea to change the tests to be more complete : 
- [ ] Testing the client and the server against invalid packet ID.
- [ ] Testing the client and the server against invalid data types (a string that's too big, an overflowing byte/int...

It may also be a good idea to use http://about.travis-ci.org/docs/user/languages/javascript-with-nodejs/ .

PR #130, opened 2015-03-14
Add travis-ci for automated continuous integration testing

deathcap:
Adds a .travis.yml and build status badge so the test suite runs automatically on ever commit and with each pull request on node 0.10, 0.12, and latest iojs, example: https://github.com/deathcap/node-minecraft-protocol/pull/1 and https://travis-ci.org/deathcap/node-minecraft-protocol/jobs/54399338 . If accepted this project will also have to be enabled on https://travis-ci.org/repositories (log in with GitHub, go to repositories, turn on node-minecraft-protocol)

There was some previous discussion about Travis in https://github.com/PrismarineJS/node-minecraft-protocol/issues/30

As part of this PR I also disabled the client tests. Not sure if these are supposed to work, they fail for me (time out) both locally and on Travis-CI. If fixed they can easily be re-enabled but until then at least the 118 packet and mc-server tests run and pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: node_js
+node_js:
+  - "0.10"
+  - "0.12"
+  - "iojs"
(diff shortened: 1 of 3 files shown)


roblabla:
yupyup, I :+1: this wholeheartedly. Only thing is, we're using circleci for the rest of the prismarinejs projects, so I think we'll be using it here too. Also, we're planning on rewriting the whole test suite down the road to get rid of the java dependency (the idea is to use a bukkit plugin to generate a file full of valid packet, one with occasional invalid packets, and one with complete garbage data, and use this file as the testbed instead of the real server.

roblabla:
Closing this, should really use Circle-CI.

PR #132, opened 2015-03-18
adding command line switches to examples/proxy.js for dumping messages

thejoshwolfe:
I found this useful for investigating the 0x8 messages for opening chests, which appears to be broken over in mineflayer.


roblabla:
Oooh, nice improvements. Merging :D

Issue #136, opened 2015-03-23
0.13.1 broke require('minecraft-protocol/lib/client')


deathcap:
Using minecraft-protocol 0.13.0 I was able to require certain parts of the library as follows:

```javascript
var Client = require('minecraft-protocol/lib/client')
    , protocol = require('minecraft-protocol/lib/protocol')
```

this no longer works as of 0.13.1:

```
> require('minecraft-protocol/lib/client')
Error: Cannot find module 'minecraft-protocol/lib/client'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:278:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at repl:1:1
    at REPLServer.defaultEval (repl.js:132:27)
    at bound (domain.js:254:14)
    at REPLServer.runBound [as eval] (domain.js:267:12)
    at REPLServer.<anonymous> (repl.js:279:12)
    at REPLServer.emit (events.js:107:17)
```

* http://registry.npmjs.org/minecraft-protocol/-/minecraft-protocol-0.13.0.tgz
* http://registry.npmjs.org/minecraft-protocol/-/minecraft-protocol-0.13.1.tgz

"^0.13.0" semver matches 0.13.1 so this change can be picked up automatically (patch release).. is lib/* not expected to be considered part of the public API? If not I guess I can change to `require('minecraft-protocol').Client`, `require('minecraft-protocol').protocol`, but wanted to be sure. 0.13.1 change log had no mention of this change:

## 0.13.1
* Externalized rsa-wrap library to its own npm module, named ursa-native
* Fixed broken bed-related packets (thanks rom1504)

but it appears to be from https://github.com/PrismarineJS/node-minecraft-protocol/pull/115 Add support for ES Harmony with Babel

wtfaremyinitials:
```js
require('minecraft-protocol/dist/protocol')
```

Should do the trick.

0.13.1 added the new ES6 build system

deathcap:
Concerned about using `require('minecraft-protocol/dist/protocol')`, relying on an implementation detail(?) of the internal module paths (also https://github.com/PrismarineJS/node-minecraft-protocol/issues/125, but npm install fixes that).. `require('minecraft-protocol').protocol` works but for `require('minecraft-protocol')`.Client there's a slight complication, not currently exposed to browserify: https://github.com/PrismarineJS/node-minecraft-protocol/pull/137

[...]

PR #137, opened 2015-03-23
Expose the Client class to browser

deathcap:
https://github.com/PrismarineJS/node-minecraft-protocol/pull/71 added support for browserify by overriding index.js with a browserify-specific browser.js exposing only the `protocol` submodule, a subset of node-minecraft-protocol that works well in web browsers. This PR additionally exposes the `Client` class, which is also browserifiable (I am using it in https://github.com/deathcap/wsmc, previously as require('minecraft-protocol/lib/client') until GH-136; cleaner to expose on the default export anyways). 

For comparison here's the exports from the (non-browser) dist/index.js:

```javascript
module.exports = {
  createClient: createClient,
  createServer: createServer,
  Client: Client,
  Server: Server,
  ping: require("./ping"),
  protocol: protocol,
  yggdrasil: Yggdrasil };
```

and the updated browser.js:

```javascript
 module.exports = {
  Client: require('./dist/client'),
   protocol: require('./dist/protocol')
 };
```

createServer and createClient use ursa so cannot currently be exposed in the browser (despite ursa-purejs — but out of scope of this issue), yggdrasil uses a third-party authentication server so it cannot be used either (without, e.g., a CORS proxy, or replacing with OAuth, etc.), Server and ping might work but haven't tested (if they do, can be exposed later if needed), but both `protocol` and `Client` definitely work so can be safely exported here.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,4 +1,5 @@
 
 module.exports = {
+  Client: require('./dist/client'),
   protocol: require('./dist/protocol')
 };
--- a/src/client.js
+++ b/src/client.js
@@ -1,8 +1,6 @@
-var net = require('net')
-  , EventEmitter = require('events').EventEmitter
+var EventEmitter = require('events').EventEmitter
   , util = require('util')
   , protocol = require('./protocol')
-  , dns = require('dns')
   , createPacketBuffer = protocol.createPacketBuffer
   , compressPacketBuffer = protocol.compressPacketBuffer
   , oldStylePacket = protocol.oldStylePacket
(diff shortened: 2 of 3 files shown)


roblabla:
I'm a bit iffy on merging this. By default, browserify isn't capable of browserifying the "net" module, since the browser can't do raw TCP. Things inside the "index.js" is what I consider being part of the public API for desktop, and similarely, things inside "browser.js" is what I consider being part of the browser public API (which can only create and read raw packet buffers).

We _could_ refactor Client such that connect lives inside index.js, and client only exposes setSocket ? That would remove the client.js dependency on net.

deathcap:
:+1 to moving `net` (and `dns`) out of client.js. Updated this PR with this change, moving the network-specific code.

Note that browserify stubs `net` and other non-browserifiable nodejs core modules with an empty module: https://github.com/substack/node-browserify/blob/master/lib/builtins.js#L16 - so the code still works on the browser, but I agree it is cleaner to make this refactor.

`client.js` now only exposes "setSocket", which actually accepts any `Stream`: this not only includes TCP sockets, but I've also used it with [websocket-stream](https://www.npmjs.com/package/websocket-stream) and [workerstream](https://www.npmjs.com/package/workerstream). Anything implementing an appropriate Stream interface should work.

roblabla:
Oh, I didn't know that. Anyway, thanks for the update ! Merging.

PR #138, opened 2015-03-23
Add circle ci badge

### LINKED ISSUE #133: Circle ci badge
It would be nice to have a circle ci badge on the readme (https://circleci.com/docs/status-badges)
Also, is there a page where everybody can see how the build did  ? (like in travis ci) Or is it restricted to the owner or something ?

rom1504:
fix #133 

It would be better to add a circle-token=:circle-token in the image url so that everybody can see the image (see https://circleci.com/docs/status-badges#access-control)


roblabla:
Now we just have to make the tests pass... :P

rom1504:
@roblabla Yeah that would be better :P

PR #139, opened 2015-03-25
fix readPosition

rom1504:
see http://wiki.vg/index.php?title=Data_Types&diff=6529&oldid=6491
thanks to Fenhl


roblabla:
Wow, very nice catch. Merging.

PR #141, opened 2015-03-26
added hook for before client receives ping

bschlenk:
createServer takes additional option 'beforePing' which is a function
receiving the response to send back to the client and the client object,
and returning a modified response to send back.
The client now has the serverHost, serverPort, and protocolVersion
pulled from the handshake packet. This can be used to proxy
the client to different backend servers depending on the domain name.


wtfaremyinitials:
:+1:

@roblabla?

roblabla:
Merging. Thanks for the PR !

Issue #61, opened 2013-08-05
Transform asserts into proper error managements


roblabla:
A lot of times, I find myself unable to debug an issue because my program will prematurely quit due to an assert failing at the protocol level. This is very problematic, and it would be nice if, instead of using asserts everywhere, the "error" event was triggered. It would help both for debug and recovering the application when it is still possible.

roblabla:
Just to clarify : The problem here isn't assert in and of itself. The problem is any throw in protocol.js does not result in an 'error' event in client, but instead results in an uncatchable error.

roblabla:
#174 mostly fixes this. We just need to check the error handling actually works through some tests now :D

PR #150, opened 2015-03-30
Remove asserts from non-test code

wtfaremyinitials:
#61

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,5 @@
 var EventEmitter = require('events').EventEmitter
         , util = require('util')
-        , assert = require('assert')
         , crypto = require('crypto')
         , bufferEqual = require('buffer-equal')
         , superagent = require('superagent')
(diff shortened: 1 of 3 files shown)


wtfaremyinitials:
Odd...

The tests pass on my machine.

roblabla:
1. This should be using `new Error()` instead of throwing strings. Otherwise, we lose the stacktrace information, and that sucks.
2. The real problem behind #61 is that throwing in protocol.js does not trigger an "error" event in mineflayer, but instead throws an uncatchable error outside of user code. This is obviously unacceptable for a library, especially for creating servers.

rom1504:
There are still some assert in the current code, this should be fixed.

roblabla:
Asserts are normal since we're now using throws instead of `return { error }`. What needs to be fixed is, subclassing errors in a way that they make sense.

rom1504:
@roblabla asserts emit exception ?

roblabla:
@rom1504 asserts throw an Error, which we catch higher up the stack and transform into an emit error. That's what we want. We just want to modify those asserts so they use custom error classes instead.

PR #155, opened 2015-04-16
reorganize badges in readme

rom1504:
I think it's better on one line that way.


roblabla:
Doing this would mean the gitter badge wouldn't show up on npm :(

rom1504:
Good point, I didn't think about that, putting the gitter and circle ci badge on the same line is better though, isn't it ?

rom1504:
Or maybe all badges should be on the second line like this https://github.com/feross/webtorrent ? I'm not sure.
One badge by line is kinda ugly anyway...

PR #159, opened 2015-05-08
Move protocol def to json

rom1504:
Conversion of the packets object to a protocol.json file.
It surely doesn't achieve everything you planned in Protocols, but imho it's a good step.


roblabla:
the problem i have with rushing this is that I don't want to be stuck with previous art. If we add this in minecraft-data, we need to try our best not to break the world and keep the format the same. This format is not ideal for different reasons, the main problem I see being the condition field. 

It's true that it's a good first step though. I think what I'll do is fork your branch and play with it

rom1504:
I don't think it has to go in minecraft-data right now. That format doesn't have to stay the same, it can be changed sure.

For the condition field I chose a simple format that fitted the possible functions : 
```
{
 "field":"name of the field checked",
"values":[possibles,values],
"this":true, // if the check is on ['this'] : not exactly sure what it means, I just did that to make it fit
"different":true // if the check is whether the values are different instead of equal
}
```
If an other format that made more sense is better, sure we should change it !

And yeah I did add a similar format for count for the world_particle packet.

My idea is having a json format in the first place, so we can compare how to adapt it to make it fits Protocols, and make sure Protocols handles everything that's needed for node-minecraft-protocol.

Btw all commits on this branch/PR work (not just the last one), so you can look at every commit to understand/change the modifications I made.
For the "convert protocol_def.js to protocol.js" commit I just did a var a=require("protocol_def.js"); console.log(JSON.stringify(a,null,2); (well actually I used fs.writeFile but you get the idea : I didn't do that manually)

rom1504:
With a condition type I guess we'd transform this : 
```json
{
  "name": "entityId",
  "type": "int",
  "condition": {
    "field": "inventoryType",
    "values": [
      11
    ]
  }
}
```

to 

```json
{
  "name": "entityId",
  "type": "condition",
  "typeArgs": {
    "field": "inventoryType",
    "values": [
        11
      ],
   "type": "int"
  }
}
```

?

roblabla:
something along those lines, yeah.

rom1504:
Here is what the current json schema looks like : https://github.com/rom1504/node-minecraft-protocol/blob/move_protocol_def_to_json/protocol/protocol_schema.json

rom1504:
New condition type done.
I'm not sure what to do about the count field, http://wiki.vg/Protocol#Particle is kind of a pain.

src/protocol.js
```diff
@@ -905,6 +120,38 @@ for (var n in entityMetadataTypes) {
   entityMetadataTypeBytes[entityMetadataTypes[n].type] = n;
 }
 
+function readCondition(buffer,offset,typeArgs, rootNode)
+{
+    if(!evalCondition(typeArgs,rootNode))
+        return null;
```
roblabla: Don't return null here. Instead, return { value: null, size: 0 }.

src/protocol.js
```diff
@@ -1429,7 +680,7 @@ function read(buffer, cursor, fieldInfo, rootNodes) {
     };
   }
   var readResults = type[0](buffer, cursor, fieldInfo.typeArgs, rootNodes);
-  if (readResults == null) {
+  if (readResults == null && fieldInfo.type!=="condition") {
```
roblabla: Then, you can remove this check here. A reader is never supposed to return null.

rom1504:
@roblabla that'll mean the received packet will have a field with value null instead of undefined, is that really the intended behaviour ? (or should I change the `if (readResults === null) continue;` checks to check for readResults.value too ?)

roblabla:
yeah, checking for readResults.value to not set it works.

rom1504:
Okay, I'm not sure how the count field can be any better (it fits http://wiki.vg/Protocol#Particle description fairly close imo).

The format can probably still be improved, but I think it's okay. What do you think @roblabla ?

rom1504:
http://wiki.vg/Protocol#Data_types calls the "condition" type "optional", would that be a better name ?

roblabla:
Meh, both are OK. Condition makes it clear it's based on another field being there. Anyway, review time.

PR #161, opened 2015-05-13
Divide protocol js into several datatypes files

rom1504:
Also reorganize some stuff in protocol.js.
Not quite ready to merge yet.

Left to be done : 
* [ ] see if parse* functions and stuff in client.js can be put in a stream.js
* [ ] better encapsulation of packets indexes
* [ ] more comparing on differences with Protocols


rom1504:
Actually I think the rest of the refactoring can be done in an other PR.
Ready to merge.

PR #166, opened 2015-05-14
fix the indent_size in .editorconfig

rom1504:
I could automatically (using webstorms) fix the indenting of node-minecraft-protocol files, but that means changing a lot of files. Should I do it ?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ charset = utf-8
 
 [*.js]
 indent_style = space
-indent_size = 4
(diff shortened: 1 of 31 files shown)


wtfaremyinitials:
I set it to 4 because it's generally my favorite. I never got around to re indenting everything though :p




If you can do it quickly I say do it.



—

rom1504:
Done, that's a lot of (semantically preserving) changes though ;)

PR #172, opened 2015-05-19
emit error instead of throwing when write callback gives an error

rom1504:
That's probably better.
That means no crash and we let the user of nmp know about errors by listening to "error" if needed. And the user of client.write can act on errors based on the result of the callback.

After this commit only 4 throws are left : 
* one in writeRaw that would disappear if we do the same as for write : add a callback
* 3 in protocol.js that could be removed if we make the errors go into callbacks or returns

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -156,12 +156,9 @@ Client.prototype.end = function(reason) {
   this.socket.end();
 };
 
-function noop(err) {
-  if(err) throw err;
-}
-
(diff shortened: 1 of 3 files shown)


rom1504:
Ok so error crash the program if not listened to, isn't that what we want though ? (by default the program crash when node-minecraft-protocol fail parsing stuff)

rom1504:
closing because of #175

Issue #167, opened 2015-05-16
Unable to locate sign at -58, 4, -325


rom1504:
Connecting with the vanilla client through the proxy gives these error messages in the chat.

tested on a spigot server and on us.mineplex.com

rom1504:
> +SpaceManiac the client prints that if it receives a sign text update before it knows there's a sign there

roblabla:
actually, now that you mention it.... it _is_ possible to send packets out-of-order.

roblabla:
the huge packets (e.g. the map) are decompressed then re-compressed : https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/protocol.js#L230, which are asynchronous operations. Assuming those happen outside the event loop, it is entirely possible for other packets to be processed before this one.

rom1504:
```js
client<-server: play.0x33 : {"id":51,"state":"play","location":{"x":23,"y":54,"z":-18},"text1":"\"\"","text2":"\"\"","text3":"\"\"","text4":"\"\""}
client<-server: play.0x33 : {"id":51,"state":"play","location":{"x":23,"y":54,"z":-19},"text1":"\"\"","text2":"\"\"","text3":"\"\"","text4":"\"\""}
// ...
client<-server: play.0x26 : Packet too big
```
> roblabla> the world packets are sent after the sign packets
> roblabla> uncompressing is probably the culprit

rom1504:
Somehow this doesn't happen with a vanilla server.

rom1504:
Oh fixing the bug said in #168 fix that issue with a spigot server. With us.mineplex.com I got a new error when connecting : "Internal Exception: io.netty.handler.codec.DecoderException: java.io.IOException: Bad packet id 120".

rom1504:
So I think we need : 
* a emit queue (for this https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client.js#L98)
* a write queue (for this https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/proxy/proxy.js#L107) in the proxy

Not sure exactly what kind of queue that would be

roblabla:
Fixed with #174

Issue #157, opened 2015-04-26
Fix recursive stack overflow


roblabla:
We need to add a test to make sure [this kind of bug](http://pastebin.com/uWZKPHkp) doesn't happen anymore.

EDIT : Had to revert the fix for this due to all kind of problems wrecking havoc.

roblabla:
The bug is fixed with 29302ad6836738d04b9fac51c5c257c323848979 still needs a regression test though.

rom1504:
setImmediate is usually needed in js for recursion yeah.
So I guess it's needed to check there aren't other recursive stuff without setImmediate.

roblabla:
Fixed with #174

PR #174, opened 2015-05-21
Rewrite internal pipeline to use transform stremas

roblabla:
This is a huge rewrite of the internal serializer and pipeline. Makes the whole system more resilient, and fixes the following bugs : #167 #157

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,8 +1,10 @@
 var gulp = require('gulp');
 
+var plumber = require('gulp-plumber');
 var babel = require('gulp-babel');
 var options = {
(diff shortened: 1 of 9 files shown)


rom1504:
I think it would really make much more sense to put the functions of protocols from https://github.com/roblabla/node-minecraft-protocol/blob/rewrite-internals/src/protocol.js#L100 to the end of that file in their respective streams.
We can always use a stream directly if we need to use these functions without the rest of the streams.

rom1504:
For reference : that's how to fix the broken test : https://gist.github.com/mjackson/1201196
Does that exist in a package ? If not should we make a package out of it ?

PR #180, opened 2015-05-23
[WIP] Add tests for dataTypes

roblabla:
The tests are directly imported from Protocols.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "prepublish": "gulp",
-    "test": "mocha --require source-map-support/register --reporter spec"
+    "test": "mocha --recursive --require source-map-support/register --reporter spec"
   },
   "keywords": [
     "minecraft",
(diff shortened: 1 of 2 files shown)


rom1504:
Apparently there are some conflicts to solve before merging.
But I think that should be merged : we are continually improving things in node-minecraft-protocol and Protocols will need to be updated with these improvements when we decide to use it.
So we might as well make node-minecraft-protocol as good as possible, and we'll transfer that to Protocols at some point.

package.json
```diff
@@ -28,15 +28,16 @@
...
   "browser": "browser.js",
   "devDependencies": {
+    "batch": "~0.3.1",
+    "chai": "^2.3.0",
     "gulp": "^3.8.11",
-    "gulp-plumber": "^1.0.1",
     "gulp-babel": "^5.1.0",
+    "gulp-plumber": "^1.0.1",
     "gulp-sourcemaps": "^1.3.0",
+    "jsonschema": "~1.0.1",
```
rom1504: this is not used in nmp anymore

PR #193, opened 2015-07-18
Bump MINOR, UUIDs are now represented as dashed string

roblabla:


roblabla:
The test failing is just circleci failing on us.

PR #196, opened 2015-07-27
split index.js into mainly createClient.js and createServer.js

### LINKED ISSUE #179: Move stuff in index.js to a createClient.js and a createServer.js
index.js is doing too much

rom1504:
also create a few other files containing common information and functions :
* mcHexDigest.js which contain the function of the same name used in both createClient.js and createServer.js
* ursa.js which handle the abstraction on top of ursa and ursa-purejs : used in both createClient.js and createServer.js
* version.js used in index.js, createClient.js and createServer.js

I think dividing that logic into 2 files and letting index.js handle only exposing make the code much more readable : createClient.js and createServer.js are now 200 lines files when index.js was a 400 lines file.

I guess it makes sense to keep client separated from createClient and server separated from createServer to keep the possibility of running the code in the browser (browser.js)

Fix #179


rom1504:
This doesn't change anything about the behavior of the package and doesn't change what's exposed either.

PR #198, opened 2015-07-27
Improve badges

wtfaremyinitials:
The inconsistency was bothering me.


roblabla:
Haha, thanks :D

rom1504:
Well trouble is, that seems broken (both the npm and circle ci ones)

wtfaremyinitials:
@rom1504 The npm one looks right to me. I did just notice however that the CI one is set to display the last build, not the last one on master. My bad PR is making the badge display a failure.

rom1504:
yeah the npm one is ok, it's just weird that it is orange when usually it is green (see https://github.com/andrewrk/mineflayer for example)

PR #205, opened 2015-07-30
Snapshot 1.9 updates

rom1504:


rom1504:
When #207 and #208 are merged, the 1.9 PR will be just 2 commits : 1 to change the quit message format (change of 1.9) in test.js and server.js, and one  to use snapshot-1.9 branch of minecraft-data instead of 0.4.0 in package.json.
That mean we can keep working on master for 1.8 as long as needed while making just the needed changes in the 1.9 branch (probably only changes in protocol.json in minecraft-data actually)

(so no point in merging this PR for now)

rom1504:
Cleaned up and put in master and in #212

PR #207, opened 2015-08-01
use protocol.json from minecraft-data

### LINKED ISSUE #163: Putting protocol.json into minecraft-data
This is something we talked about : once protocol.json is stable enough, it could be put in minecraft-data, making that protocol definition language independent.

I guess we need to wait a bit until some modification on that file format have been done (maybe some changes about condition and about count for example)

rom1504:
fix #163


rom1504:
I think this can be merged ?

PR #208, opened 2015-08-02
Better error and fix from 1.9

rom1504:
I took the commit independant from 1.9 and rebased them into a new branch.
That's useful since we don't know when 1.9 is going to be released (might be in 6 months).

(I'll fix the actual 1.9 branch to work, it should now be only a few commits)

### DIFF
--- a/examples/proxy/proxy.js
+++ b/examples/proxy/proxy.js
@@ -159,6 +159,8 @@ srv.on('login', function(client) {
       console.log("client->server: Error in packetId " + state + ".0x" + packetId.value.toString(16));
       console.log(buffer.toString('hex'));
       console.log(packetBuff.toString('hex'));
+      console.log(buffer.length);
+      console.log(packetBuff.length);
     }
   });
   targetClient.on('end', function() {
@@ -167,9 +169,10 @@ srv.on('login', function(client) {
     if(!endedClient)
       client.end("End");
   });
-  targetClient.on('error', function() {
+  targetClient.on('error', function(err) {
     endedTargetClient = true;
-    console.log('Connection error by server', '(' + addr + ')');
+    console.log('Connection error by server', '(' + addr + ') ',err);
+    console.log(err.stack);
     if(!endedClient)
       client.end("Error");
   });
--- a/examples/proxy/proxy.js
+++ b/examples/proxy/proxy.js
@@ -140,6 +140,8 @@ srv.on('login', function(client) {
       console.log("client<-server: Error in packetId " + state + ".0x" + packetId.value.toString(16));
       console.log(buffer.toString('hex'));
       console.log(packetBuff.toString('hex'));
+      console.log(buffer.length);
+      console.log(packetBuff.length);
     }
     /*if (client.state == states.PLAY && brokenPackets.indexOf(packetId.value) !== -1)
      {
--- a/examples/proxy/proxy.js
+++ b/examples/proxy/proxy.js
@@ -140,8 +140,6 @@ srv.on('login', function(client) {
       console.log("client<-server: Error in packetId " + state + ".0x" + packetId.value.toString(16));
       console.log(buffer.toString('hex'));
       console.log(packetBuff.toString('hex'));
-      console.log(buffer.length);
-      console.log(packetBuff.length);
     }
     /*if (client.state == states.PLAY && brokenPackets.indexOf(packetId.value) !== -1)
      {
--- a/src/datatypes/numeric.js
+++ b/src/datatypes/numeric.js
@@ -24,7 +24,14 @@ function generateFunctions(bufferReader,bufferWriter,size)
     };
   };
   var writer=function(value, buffer, offset) {
-    buffer[bufferWriter](value, offset);
+    try {
+      buffer[bufferWriter](value, offset);
+    }
+    catch(err)
+    {
+      console.log("value:"+ value);
+      throw err;
+    }
     return offset + size;
   };
   return [reader, writer, size];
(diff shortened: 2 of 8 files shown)


examples/proxy/proxy.js
```diff
@@ -167,9 +171,10 @@ srv.on('login', function(client) {
     if(!endedClient)
       client.end("End");
   });
-  targetClient.on('error', function() {
+  targetClient.on('error', function(err) {
     endedTargetClient = true;
-    console.log('Connection error by server', '(' + addr + ')');
+    console.log('Connection error by server', '(' + addr + ') ',err);
+    console.log(err.stack);
```
roblabla: ```if (err && err.stack) console.log(err.stack);```

src/datatypes/numeric.js
```diff
@@ -24,7 +24,14 @@ function generateFunctions(bufferReader,bufferWriter,size)
     };
   };
   var writer=function(value, buffer, offset) {
-    buffer[bufferWriter](value, offset);
+    try {
```
roblabla: This looks like debug that should go away.

src/datatypes/structures.js
```diff
@@ -84,7 +84,16 @@ function writeContainer(value, buffer, offset, typeArgs, rootNode) {
...
+    try {
+      offset = this.write(context[typeArgs.fields[index].name], buffer, offset, typeArgs.fields[index], rootNode);
+    }
+    catch(err)
+    {
+      console.log("typeArgs : " + JSON.stringify(typeArgs.fields[index]));
+      console.log("context : "+JSON.stringify(context));
+      console.log("value : " + JSON.stringify(context[typeArgs.fields[index].name]));
+      throw err;
+    }
```
roblabla: same here.

examples/proxy/proxy.js
rom1504: It's an error, there's always an err.

src/datatypes/numeric.js
rom1504: hmm that's pretty useful for debugging precisely, gives more context than when the error is eventually caught, way later.

src/datatypes/structures.js
rom1504: same answer, that gives context to the error
it basically isn't triggered unless there's something going wrong with nmp

src/datatypes/numeric.js
rom1504: Maybe it should be something else than a console.log though but I'm not sure what

roblabla: If you want to add debugging information, use the `debug` function. I still think this one is out-of-place. Might as well add debug info in every single reader.

examples/proxy/proxy.js
roblabla: Nevermind, didn't see it was in the example.

src/datatypes/numeric.js
roblabla: For debugging info, the `debug` function should be used.

I still don't see what gain you have from logging every single number. Structures and Arrays could use some logging to get more context, but numbers ? A try/catch gets a small performance penalty because the function containing it cannot be optimized by v8, so it's a good idea to avoid putting it everywhere.

What would be a good idea is adding context to the Errors inside of Array and Structure readers/writers, so we know exactly which fieldInfo caused the trouble

src/datatypes/structures.js
roblabla: Having something like https://github.com/roblabla/Protocols/blob/master/lib/protocols.js#L98-L105 would help to get more debug information here. The structural fields (Structure and Array) could use it to add context as to where the error occured in the fieldInfo chain.

src/datatypes/numeric.js
rom1504: Hmm there was something wrong with numbers due to something else I think. Yeah you're right that one should be removed.

I turned the return {error:} to a throw Error in write because of https://github.com/PrismarineJS/node-minecraft-protocol/blob/4752a0d759d3828bd358bad389c1f587feffbfa7/src/transforms/serializer.js#L92 , so I guess that should be fixed one way or the other.

Adding context through that error object you do it indeed.

The problem with the debugging function is it displays every packet, kind of hard to find the interesting part in the ouput when enabled, but yeah I see your point.

src/datatypes/structures.js
rom1504: Yeah okay, I'll fix it using something similar to what protocols does.

src/datatypes/numeric.js
roblabla: We could use [debug](https://github.com/visionmedia/debug) module to fix this. By making multiple debug modules for each part of the application, we could remove the spammy logging done.

It might also be a good idea to just remove the spam per-packet in general. I don't think it's actually useful, and can be turned back on when we need it ?

rom1504:
Hmm actually that doesn't make much more sense to return these {error : } things, because you still need to check whether there's an error for example in https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/datatypes/structures.js#L87 and what else can we do from there except of returning the error and giving up ?

If I understand it correctly, it means the offset won't be updated (same thing here https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/transforms/serializer.js#L92) and the whole thing will fail.

So I don't know how to fix that really and I'm just going to revert d2435c8dcdf6de9441a38cf7f605fbda691359be (we can add that debug the right way later) because it's not very important and I do need this PR merged to fix my snapshot-1.9 branch.

rom1504:
ready to merge ?

PR #212, opened 2015-08-08
[WIP] Clean snapshot 1.9

rom1504:
Clean snapshot 1.9 branch.
I've put the PR target as master, it could be put into (upstream) snapshot-1.9, but some things from that branch are now in master, so it should probably be deleted and created again.

To be merged only when 1.9 is out of course.

### DIFF
--- a/src/version.js
+++ b/src/version.js
@@ -1,3 +1,3 @@
-var majorVersion='1.8';
+var majorVersion='1.9';
 var mcData=require("minecraft-data")(majorVersion);
 module.exports=mcData.version;


rom1504:
2 things to fix here : 
1. in src/server.js add a `if(version==1.8) ...`
2. in test.js check an containment instead of an equality to be version independent or something like that

These 2 changes should be done in master, then this branch need to be only one thing : change in version.js

We might even consider adding an option in createServer and createClient to specify the version and the protocol version, so this PR would be 0 change.

rom1504:
Should the version be something like `"version":{"version":47,"minecraftVersion":"1.8.1"}` next to types and states in protocol.json ?
Or is the version unrelated to the protocol and should be in a version.json ?

rom1504:
depends on #232

rom1504:
this pr now assumes merging of #233

rom1504:
based on master again

src/server.js
```diff
@@ -25,6 +25,7 @@ Server.prototype.listen = function(port, host) {
     var client = new Client(true);
     client._end = client.end;
     client.end = function end(endReason) {
+      endReason='{"text":"'+endReason+'"}';
```
roblabla: I believe this change actually needs to be in 1.8 too, according to wiki.vg

rom1504: wiki.vg says "chat" for both 1.8 and 1.9, but 1.8 server sends '"reason"' while 1.9 sends '{"text":"reason"}' which should both be valid in the "chat" format (a quoted string is a json value).

So there are 2 sides to this : 
* the tests should just check that the quoted reason is contained in the message to be cross-version
* we should check sending '{"text":"reason"}' make the vanilla client display the reason properly

roblabla: What ? No ? Both 1.8 and 1.9 send JSON (not quoted json) ?

rom1504: I meant 1.8 sends "reason", 1.9 sends {"text":"reason"}

roblabla: But are you sure 1.8 just sends "reason" ? Because not only does wiki.vg clearly state it needs to be a "chat", but last time I tried to use .end in 1.8, nmp failed and only printed "Disconnected" on the client side ^^'.

rom1504: Ah well no I'm not entirely sure. The main thing I tried is to make the tests pass, so maybe the only bug is in our tests.

rom1504: Alright I just checked again, I get an error there https://github.com/PrismarineJS/node-minecraft-protocol/pull/212/files#diff-c1129c8b045390789fa8ff62f2c6b4a9R442 if I run this code in 1.8 : the 1.8 server really sends "Failed to verify username!"

rom1504:
this PR is now based on the branch of #238 
the only change here is in version.js

rom1504:
Next steps here:
* putting the content of version.js in version.json in minecraft-data (except majorVersion/mcdataversion)
* doing #234

rom1504:
first step is done (using version.json of mcdata), but the version should probably be dynamic or at least an option of createClient and createServer to start with, more details about that in #234

rom1504:
The test fails, but I need #237 to repair it properly (it's just about doing npm install before downloading which #237 does)

rom1504:
now based on #242 
only remaining change here is 9e0373881f2555831fdd24d5ab5ef34af44c48af

rom1504:
Once #257 is merged, this can be closed.

rom1504:
And finally, this PR went from big to now 0 line :)

PR #213, opened 2015-08-08
[WIP] Kick client when he is using the wrong version of the client, fix #209

### LINKED ISSUE #209: Prevent client with the wrong version to connect to the server
Currently makes the script crash
see http://wiki.vg/Protocol#Clientbound

rom1504:


rom1504:
checking for the socket existence is useful because of https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/server.js#L41

rom1504:
Fix crashes when connecting with the wrong version first (for example https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/datatypes/numeric.js#L27 can crash, see https://nodejs.org/api/buffer.html#buffer_buf_writeuintle_value_offset_bytelength_noassert)

rom1504:
the underlying issue is now fixed, so I think this can now be merged.

rom1504:
(this will be done better in #234 and shouldn't be merged for now)

PR #221, opened 2015-08-24
Add the new type syntax support. Old type syntax doesn't work

roblabla:
PrismarineJS/minecraft-data#38

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -14,6 +14,8 @@ var mc = require('../')
   , MC_SERVER_JAR = process.env.MC_SERVER_JAR
   , SURVIVE_TIME = 10000
   , MC_SERVER_PATH = path.join(__dirname, 'server')
+  , getFieldInfo = require('../dist/utils').getFieldInfo
+  , evalCondition = require('../dist/utils').evalCondition
   ;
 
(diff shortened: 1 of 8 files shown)


test/test.js
```diff
@@ -14,6 +14,8 @@ var mc = require('../')
   , MC_SERVER_JAR = process.env.MC_SERVER_JAR
   , SURVIVE_TIME = 10000
   , MC_SERVER_PATH = path.join(__dirname, 'server')
+  , getFieldInfo = require('../dist/utils').getFieldInfo
```
rom1504: this is wrong : tests should test only the exposed interface I think (as is done for the rest of the stuff used in test.js), so getFieldInfo and evalCondition should probably be exposed in index.js

rom1504:
nice !

test/test.js
roblabla: They aren't being tested, they're being used to set up the test. I really don't think exposing them is a good idea.

rom1504: ok

rom1504:
bump node-minecraft-data to 0.6.0 and the tests should pass

Issue #232, opened 2015-09-12
Stop using numerical ids everywhere in nmp


rom1504:
All PLAY ids changed in 1.9, so we should just use the names.
Example of use in PLAY state : 
* https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/test/benchmark.js#L8
* https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/test/test.js#L411
* https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/proxy/proxy.js#L130
* https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/createClient.js#L48
* https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/server.js#L29

Example of use in other states : 
* https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/createClient.js#L49

(numerical ids currently break tests in 1.9, see https://circleci.com/gh/rom1504/node-minecraft-protocol/115 )

rom1504:
Play state packet ids are now not used in nmp ( #233 )
The same can be done for the other states.

roblabla:
Done everywhere except in tests by #243

rom1504:
It's done in tests actually.
Only thing remaining here is the example proxy.

rom1504:
done

PR #233, opened 2015-09-12
transform play numerical ids to names

rom1504:
* start working on #232 
* remove entity metadata test values (that should probably be in a datatypes/ test or something) to support cross version
* improve error handling a bit in serializer.js

### DIFF
--- a/src/createClient.js
+++ b/src/createClient.js
@@ -92,7 +92,7 @@ function createClient(options) {
   }
 
   function onKeepAlive(packet) {
-    client.write(0x00, {
+    client.write('keep_alive', {
       keepAliveId: packet.keepAliveId
     });
   }
--- a/src/createClient.js
+++ b/src/createClient.js
@@ -45,7 +45,7 @@ function createClient(options) {
 
   var client = new Client(false);
   client.on('connect', onConnect);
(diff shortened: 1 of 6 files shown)


roblabla:
Is this ready to merge ?

rom1504:
yeah

roblabla:
With this merged, we'll want to stop emitting packets by ID and [state;id] I think.

rom1504:
@roblabla this PR changes the numerical ids into names for the PLAY state, but I didn't change anything for the other states, would using the name for the other states "just work" ? (I think there was some name conflict)

(example there : https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/createClient.js#L49 )

roblabla:
There shouldn't be any name conflict (we make sure of that in packets.js). It should "just work". The only trouble is, I can't find where we're emitting the packets by [state;id] ._.

roblabla:
Right, we don't emit packetIds or [state; packetIds]. Instead, we overload on to transform it into packetName : https://github.com/PrismarineJS/node-minecraft-protocol/blob/e3b435af1b4a3fad192d9534fca2cac466c520eb/src/client.js#L70

PR #238, opened 2015-09-14
Fix quit reason

rom1504:


src/server.js
```diff
@@ -25,6 +25,7 @@ Server.prototype.listen = function(port, host) {
     var client = new Client(true);
     client._end = client.end;
     client.end = function end(endReason) {
+      endReason='{"text":"'+endReason+'"}';
```
roblabla: Wait so does this work in 1.8 ? Can we send an 1.8 client { "text": "Disconnected" } ?

rom1504: Yep, I just tried adding a 
```js

  setTimeout(function(){
    client.end("go away please");
  },15000);
```
at https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/proxy/proxy.js#L84
and it just works :)

roblabla:
Ready to merge ?

rom1504:
yeah it's ready

Issue #245, opened 2015-09-20
using more of ES6


rom1504:
Since nmp is now es6, I think we should use it a bit more.
In particular we could change classes to the es6 syntax in:
* [client.js](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client.js)
* [server.js](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/server.js)
* [protocol.js](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/protocol.js)

PR #248, opened 2015-09-20
more es6 classes

rom1504:
See #245 
Transform client.js, protocol.js and server.js in es6 classes.
I think it improves the code.


roblabla:
For properties, you can use https://github.com/jeffmo/es-class-properties ^^

rom1504:
Ah indeed, will do.

rom1504:
done

roblabla:
awesome man :D

PR #256, opened 2015-09-24
Colons are supposed to be next to the closest char

mhsjlw:
At least, in English


roblabla:
Heh, this is what happens when the two main maintainers are french :P

Issue #234, opened 2015-09-12
Cross version


rom1504:
There are 2 kinds of cross version support:
- static : for one instance of a program using nmp, only one version is supported
- dynamic : for one instance of a program using nmp, several versions are supported

Currently we almost got static support (see #212 about that).

You'd think dynamic support would be much harder but actually all it would be about is getting the protocol information from the server/client you are connecting to, and using the correct protocol.json to parse and serialize the packets for that server/client.

Of course, that wouldn't solve the multi-version support for users of nmp which would still have to deal with the packets difference between version, one way to deal with that might be #231 

Dynamic support  would allow createClient and createServer to work for both 1.8 and 1.9 without   change to the api, nmp would just need to expose the version of the server/client  it connected to.

This idea might be extendable to multi-platform support (with mcpe) but that would be a bit harder (the protocol architecture of mcpe is different from mcpc)

Related with #65

rom1504:
About dynamic support:
Adding a versions:["1.8", "1.9"] option to createClient so that 
 the bot will only connect to versions supported by the nmp user.

And the default versions might be just ["1.8"] in nmp 0.15.0 . In 0.16.0 it might become ["1.9"] and the nmp user would anyway would have to manually update to 0.16.0 so we wouldn't "pull the rug under the user's feets".

Or we could just say that "versions" is not optional.

Dynamic just mean multiple versions support at the same time, we don't force the user to support everything if he doesn't care to.

(the same thing goes for createServer)

rom1504:
We need to be able to convert the protocolVersion (https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/createServer.js#L143) (=version.version) to the corresponding majorVersion/mcDataVersion. node-minecraft-data should be able to do that in https://github.com/PrismarineJS/node-minecraft-data/blob/master/index.js#L18 (just check if one of the mcData submodule is that version)

Then we basically just need to do `require('minecraft-data')` when we know the version. To have cross version support dynamically I think we'd need to assume the other states than play are the same cross version (is this true ?) in order to go far enough for the client/server to give us the version he wants.

roblabla:
@rom1504 it would work for anything >= 1.7. For < 1.6, there would only be one state (LEGACY, or whatever we end up calling it). I think we should keep the require('minecraft-data')('version'). It's easier and more flexible.

rom1504:
[...]

PR #257, opened 2015-09-29
enable cross version with an option in createClient and createServer

### LINKED ISSUE #65: Protocol only works on major verions
I notice that the protocol only works on major version of Minecraft. This isnt a major bug or issue, but as a suggestion was to make it where the protocols are stored in different libs and defined in the code with a version so it would look like

var mc = require('minecraft-protocol');
var client = mc.createClient({
  host: "localhost",   // optional
  port: 25565,         // optional
  username: "email@example.com",
  password: "12345678",
  version: "1.6.2",
});

Which could help alot.

I may submit a PR to implement this

### LINKED ISSUE #240: Run the tests for each supported version
part of #234

rom1504:
* put parsePacketData in deserializer and createPacketBuffer in serializer
* remove packets from the index and expose readPacket instead
* load packets when needed in various files
* make tests test every supported version

static cross version of #234, fix #65, fix #240

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Parse and serialize minecraft packets, plus authentication and encryption.
 
 ## Features
 
- * Supports Minecraft version 1.8.8
+ * Supports Minecraft version 1.8.8 and 1.9
  * Parses all packets and emits events with packet fields as JavaScript
    objects.
  * Send a packet by supplying fields as a JavaScript object.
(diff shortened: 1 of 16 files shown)


rom1504:
Ah I forgot to fix https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/proxy/proxy.js#L131

rom1504:
3f50a6f19a53df902c96f1be01a37c07e3da5c06 makes it possible to test that cross version functionality easily : just give the version to the proxy

`node proxy.js localhost user password 1.9`

PR #259, opened 2015-09-30
Allow beforePing to receive a callback.

basicer:
Sometimes a server needs to preform some asnyc operation before it can respond to a ping request.  To handle this case, we allow the return value of before ping to be a thenable, and then wait on that promise.

By returning a thenable instead of sending callback, we easily maintain backwards compatibility.

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -100,15 +100,25 @@ function createServer(options) {
         "favicon": server.favicon
       };
 
+      function answerToPing(err, response) {
+        if ( err ) return;
+        client.write('server_info', {response: JSON.stringify(response)});
+      }
+
       if(beforePing) {
-        response = beforePing(response, client) || response;
+        if ( beforePing.length > 2 ) {
+          beforePing(response, client, answerToPing);
+        } else {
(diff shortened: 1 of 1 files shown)


roblabla:
The idea is good, but let's not use promises. NMP is entirely callback-based, it'd be weird to have promises in some places. Instead of doing this, we could detect if beforePing.length === 3, in which case we handle the result asynchronously.

```
function cb(err, res) {
  if (err) // ?
  client.write(
}
if (beforePing && beforePing.length === 2)

**Comment by roblabla** [comment]
You added an empty roar file. If you could remove it, this is ready to merge.

File: src/createServer.js [hunk 1]
```diff
@@ -100,15 +100,25 @@ function createServer(options) {
         "favicon": server.favicon
       };
 
+      function callback(err, response) {
```
rom1504: I think it would be better to give that function an other name, for example answerToPing

basicer:
Ooops, where did that come from!

roblabla:
Thank you for your contribution :D

PR #210, opened 2015-08-06
[WIP] Modularize NMP

roblabla:
This is a first, crude attempt at modularizing NMP, mostly meant to ignite some discussion. I chose to modularize the keepalive module as it is fairly simple and straightforward. I plan on modularizing two other components before calling this feature-complete : 
- [ ] Handshaking
- [x] Ping

Open questions : 
- Should we provide a way to detach a module from NMP ?
- Should we actually standardize modules, or should they be kept internal ?

PR #260, opened 2015-09-30
Allow clients to specify session information and skip talking to Yggdrasil

basicer:
If we already have the selectedProfile in hand, we don't need to talk to Yggdrasil.  When creating a client now, we always refresh the auth token to get the selected profile, causing the access token to be rotated. Passing the session information allows the user to stop the token from being rotated, so it can continue to be used.

### DIFF
--- a/src/createClient.js
+++ b/src/createClient.js
@@ -64,7 +64,14 @@ function createClient(options) {
       }
     };
 
-    if(accessToken != null) getSession(options.username, accessToken, clientToken, true, cb);
+    if(options.selectedProfile) {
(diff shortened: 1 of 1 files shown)


roblabla:
So #262 fixed the problem that we always rotate the accessTokens. We now validate the existing tokens first, and only rotate if necessary. Removing yggdrasil entirely is outside the scope of createClient. If you need to do that, you can either use the raw Client class, or, when the modularization work ( #210 ) is done, it'll be possible to simply remove a module to bypass that.

PR #264, opened 2015-09-30
use mocha for benchmark (allow better mocha grepping + it's cleaner)

rom1504:
These tests always succeed. It would be possible to check they don't take too much time, but what value should we compare against ?
* a fixed value : would fail for slower computer
* something specific to circle ci : producing circle ci artifact containing the last score value + comparing against the score value of the parent commit : this seems a bit overkill though

Anyway, I think this can be merged as-is, we can figure out how to make these tests fail later.

Edit: Well mocha has a build-in timeout so I set it to 20s for now, to be safe.

### DIFF
--- a/test/benchmark.js
+++ b/test/benchmark.js
@@ -1,8 +1,8 @@
 var ITERATIONS = 100000;
 
 var mc = require("../");
-  util = require('util'),
-  states = mc.states;
+var util = require('util');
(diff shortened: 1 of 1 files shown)


rom1504:
any thought on that @roblabla ?

roblabla:
Woops, forgot about that.

PR #274, opened 2015-10-03
use protodef

rom1504:
* the hack in package.json is of course to be removed once Protocols is published
* it might be better to directly expose the datatype instead of creating an instance whenever we need to use a type


rom1504:
1000 lines deleted from nmp :D

rom1504:
Alright it works using the generic ProtoDef parser. Problems:
1. the benchmarks perf are 10x worse, I think we might want to optimize the protodef class (and directly bench protodef)
2. is this https://github.com/PrismarineJS/node-minecraft-protocol/pull/274/files#diff-47e149110fa68fd8748327a5451c3e3dR41 ok ? (redefining the .write method of the serializer to write by name instead of id). It could totally be done in client.js instead

rom1504:
Now the benchmarks are only 3x worse (12s/6s vs 3.5s/2s), I'm not sure what this really means though, it's still unlikely going to be blocking in any usage.

rom1504:
Error handling is back.

Only thing missing: packetToParse. I'm not exactly sure how to do that though.

package.json
```diff
@@ -9,7 +9,8 @@
   },
   "scripts": {
     "prepublish": "gulp",
-    "test": "mocha --recursive --require intelli-espower-loader --require source-map-support/register --reporter spec"
+    "test": "mocha --recursive --require intelli-espower-loader --require source-map-support/register --reporter spec",
+    "postinstall": "node_modules/.bin/babel --recursive  --stage 0 -d node_modules/protodef/dist/ node_modules/protodef/src/"
```
roblabla: Why ? This should be handled by protodef in the prepublish.

rom1504: because of the npm bug about gulp + git dependencies. I'll remove it once protodef new version is published.

src/transforms/compression.js
```diff
@@ -1,4 +1,7 @@
-var [readVarInt, writeVarInt, sizeOfVarInt] = require("../datatypes/utils").varint;
+var ProtoDef = require("protodef").ProtoDef;
+var proto = new ProtoDef();
+
+var [readVarInt, writeVarInt, sizeOfVarInt] = proto.types.varint;
```
roblabla: I feel like this should be easier...

rom1504: ah yeah I forgot about this. datatypes should be exposed by protodef directly hmm...

src/transforms/serializer.js
```diff
@@ -1,185 +1,60 @@
...
-  proto.addTypes(conditional);
   proto.addTypes(types);
+
+  Object.keys(packets).forEach(function(name) {
+    proto.addType("packet_"+name,["container",packets[name].fields]);
+  });
+
+  proto.addType("packet",["container", [
+    { "name": "id", "type": "varint" },
+    { "name": "params", "type": ["switch", {
```
roblabla: That would be a really good use-case for anon type ! It would avoid changing everything to packet.params in our inner structure ;)

roblabla: Or maybe not : we might conflict with ID. Urgh. I think it might be a good idea to rethink switch a tiny bit so it can "contain" what it switches on (id here) and avoid conflict, without making the API suck.

rom1504: well, we do need the id too ;)

roblabla: yes, but it sucks that we break API compat over something so trivial.

rom1504: well the proxy kind of got to change because the serializer doesn't know about packet names anymore (https://github.com/PrismarineJS/node-minecraft-protocol/pull/274/files#diff-662847b072812401e7452d353a20aeb4R136) since these are types now, not the "switch compareTo" or something.

rom1504:
Okay, now using ProtoDef.
The serializer/parser now know only ids, not names. Only Client knows about names. That makes some kind of sense I guess.

I think it's probably ready to merge.

rom1504:
Updated using https://github.com/roblabla/ProtoDef/pull/23
It works fine but 2 potential problems:
* my mapper type hides the packet ids, is that a problem ? do we need to provide the ids ?
* using pstring made the benchmark go up again, so we are currently at 10x the time of master benchmark. The reason why that happens is ProtoDef::read/write/sizeOf involves some overhead, to improve these benchs, we'd have to remove or at least decrease that overhead. Anyway, I'm still not sure if what this benchmark measure is important at all.

rom1504:
So about hiding the packet id : that's okay in the general case but we better provide the id when it fails (something like https://github.com/rom1504/ProtoDef/commit/ce7ed4ea2083e8f2035a8bf97e001b1f591d1b70 )

The slower perf are okay for now, it's not really a problem, we can improve it when it becomes one. (some ideas about improving the perfs there though https://github.com/roblabla/ProtoDef/issues/28 )

rom1504:
Ready for review.

rom1504:
I put the gm3 fix in this PR, since it already contains fixes/changes for the proxy.

roblabla:
Could you squash the commits concerning protodef ?

rom1504:
@roblabla done

roblabla:
Mega-merge of doom done.

rom1504:
so after 5 years, how did performance evolve : 
1. before all this protodef story, 10k iteration of the benchmark took 1s in reading (https://github.com/PrismarineJS/node-minecraft-protocol/pull/274#issuecomment-149351295)
2. protodef interpreter made it be 3s (while simplifying the code a lot)
3. in recent times, with faster computer, it went to 0.4s
4. with the new protodef compiler, we're at 31ms
so assuming everything is independant (hardware and protodef code change), we are now 3x faster than the previous mostly manual solution.

Also according to some benchmarks made by @Karang we're now at protobuf speed levels.

So this seems to be proving for sure that it is not necessary to write manual parser and serializer to get good perfs, even with a bad protocol.

A next step to this story might be to expand the new shiny (and very simple !) protodef compiler to other languages by reusing the IIFE concept. It might happen someday ;)

PR #278, opened 2015-10-08
Fix createServer

basicer:


roblabla:
...  I feel a bit stupid right now xD.

rom1504:
How come the tests didn't fail on this oO

rom1504:
Ah right, there are 0 online mode test for the nmp server, we should fix that.

PR #283, opened 2015-10-12
Enable nmp server compression

rom1504:


rom1504:
@roblabla do you know why these lines were commented ?

Pretty sure it means the nmp server never compress anything, which is a big problem for flying-squid

rom1504:
Hmm hmm it might have been disabled to make this possible https://github.com/PrismarineJS/node-minecraft-protocol/blob/8d9f37c2ba99c7efc1e1823ee511733ec41beca3/examples/proxy/proxy.js#L128 

But this thing of the proxy should be in createServer, not in the proxy.

rom1504:
Ok I just tried, this doesn't break the proxy.

@roblabla could you merge this and release a new version ? I need this fix for flying-squid

rom1504:
Ah it's always possible to force it in flying-squid so I did that for now (https://github.com/mhsjlw/flying-squid/blob/master/src/lib/serverPlugins/login.js#L17), not so urgent.
That's probably what was making minespy so slow though. (and Egoscio proxy)

PR #285, opened 2015-10-13
use bufferEqual instead of buffertools

rom1504:
native modules are annoying


rom1504:
Ok, I tested the proxy, it works with that change.

PR #286, opened 2015-10-13
remove unnecessary proxy dependency (it's already in nmp package.json)

rom1504:
We might want to consider not having the dir for every example but client_chat


rom1504:
Okay let's not do that, that way if someone moves the proxy dir somewhere out of nmp it will still work.

PR #288, opened 2015-10-17
[WIP] Use protodef serializer

rom1504:
based on #274 

I'm most likely doing it wrong, but it doesn't work...


rom1504:
Idea: change the serializer when changing of state.

Not sure about these unpipe, and I don't really understand the errors.

rom1504:
back in #274

PR #305, opened 2015-11-16
update to babel6

rom1504:
idk if we still need babel-runtime as a normal dep, needs testing.


roblabla:
Let's wait for babel6 to become more stable. Right now some basic features are bug-riddled, and there are regressions here and there.

rom1504:
ok

rom1504:
@roblabla I think we could move to babel6 now, what do you think ? Some of the regressions I know of have been fixed.

rom1504:
Well I can't see any remaining problem with babel6, let's move to it.

PR #308, opened 2015-11-30
use require-self to be able to use require('minecraft-protocol') in examples

rom1504:
Seems nicer to me !

Less confusing for users. (and yes, it works)


roblabla:
Awesome :D Merging this.

PR #309, opened 2015-12-20
[WIP] Update to new protodef

rom1504:
The packets test all pass but there are still problems in the other two test series.

works with https://github.com/roblabla/ProtoDef/pull/43

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -62,6 +62,7 @@ class Client extends EventEmitter
 
 
     this.deserializer.on('error', (e) => {
+      e.field=e.field==undefined ? "" : e.field;
       var parts=e.field.split(".");
       parts.shift();
       var deserializerDirection = this.isServer ? 'toServer' : 'toClient';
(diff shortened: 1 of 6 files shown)


rom1504:
So the current problem here is that the new parser is async unlike the old one. What that means is you can't change the parser sync and except it to be changed right now like we expect it currently (and the vanilla server expect it too) : getting a handshake packet and then a play packet for example.
(I think similarly the decompressor is now async, not sure if that's a problem)

A way to solve this might be to have a parser handling all states in nmp and just routing the chunks to the different state parser.

I think there is some ready to use router stream, see if that's interesting.

rom1504:
https://github.com/rom1504/node-minecraft-protocol/commit/0a6707c357c4bcdf54301caab056e610ff7edc62 makes mc-server tests pass but it doesn't make the proxy work for example. Probably because the vanilla server is not as nice as to wait a bit before sending packets of the next state.

rom1504:
I think the serializer need to know about the states, and what should change them.
I can't see how else it can parse packets properly.
I don't even understand why master currently works.

rom1504:
Not doing this anymore, see https://github.com/roblabla/ProtoDef/pull/43#issuecomment-185855622

PR #314, opened 2016-01-13
Update browser.js, sync with index.js excluding createClient/createServer

deathcap:
All of node-minecraft-protocol can be browserified, except for createClient and createServer (since it depends on request, and browsers cannot send arbitrary HTTP requests to third-party servers (yggdrasil, authentication)).

Primarily I'd like access to `states` (instead of using `require('minecraft-data')(mcversion).protocol.states`; see https://github.com/deathcap/wsmc/commit/2fbfd59e4c951580ec98ae4b6dce62b8769e0a24#comments), but I added all the other fields too since they browserify without error.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -1,4 +1,12 @@
+var Client = require('./client');
+var Server = require('./server');
+var serializer = require("./transforms/serializer");
+
 module.exports = {
-  Client: require('./client'),
+  Client: Client,
+  Server: Server,
+  states: require("./states"),
+  createSerializer:serializer.createSerializer,
+  createDeserializer:serializer.createDeserializer,
   supportedVersions:require("./version").supportedVersions
 };


src/browser.js
```diff
@@ -1,4 +1,13 @@
...
+var serializer = require("./transforms/serializer");
+
 module.exports = {
-  Client: require('./client'),
+  Client: Client,
+  Server: Server,
+  states: require("./states"),
+  createSerializer:serializer.createSerializer,
+  createDeserializer:serializer.createDeserializer,
+  ping: require('./ping'),
```
rom1504: ping uses the net package. Can it really be browserified ?

deathcap: Hmm.. ping does require('net') but doesn't seem to use it? Browserifies ok

rom1504: Hmm indeed. I wonder how (and if) it works at all since it uses Client.connect https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/ping.js#L45 which isn't defined in client.js
It should use that net package, so (when it's fixed) it won't work with browserify.

rom1504: .connect is defined there https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/createClient.js#L15

deathcap: I'll remove ping from browser.js

PR #315, opened 2016-01-13
Test on newer version of Node

deathcap:
circle.yml specified a very old version of node (0.10.28), current versions on https://nodejs.org/en/ are v4 (LTS) and v5 (stable).

CircleCI documentation on their installed node versions: https://circleci.com/docs/environment#node-js

How to test multiple node versions on CircleCI: https://discuss.circleci.com/t/testing-multiple-versions-of-node/542 (this is simpler on Travis CI you just list all the versions in an array; requires test overrides on Circle – punting on making this change for this PR for now)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   environment:
     MC_SERVER_JAR_DIR: /home/ubuntu/node-minecraft-protocol/minecraft-server/
   node:
-    version: 0.10.28
+    version: 4
   java:
     version: openjdk7
 dependencies:


deathcap:
Experimented with various circle.yml configurations, notes:

* The CircleCI default node version is 0.10.33 (used if omit `node: version:`), previously this repo was using 0.10.28.
* https://circleci.com/docs/environment#node-js lists 4.2.2, 4.1.0, and 4.0.0 but that page is outdated: they also have 4.2.4, the current 4.x node
* Specifying version of "4" resolves to the latest v4 branch, currently 4.2.4

For Travis CI I've been using "4, 5, stable", but for CircleCI, if we can only specify one at a time I think 4 is reasonable, since it is the current long-term support release branch.

rom1504:
We'd need to change https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/package.json#L27 too if we stop supporting node <4

roblabla:
TBH, I'm fairly sure using NMP on node8 is going to break badly. That line should at least be bumped to 0.12, if not 4.0 ^^.

Issue #312, opened 2016-01-06
WebSocket support


rom1504:
How hard would it be to use a websocket connection instead of tcp in the server ? (and both ?)

That would allow anybody to connect through a web browser to connect to flying-squid using https://github.com/voxel/voxel-clientmc

Gjum:
relevant https://github.com/deathcap/wsmc

rom1504:
Yeah sure I know about wsmc, but it would be cool to have native support for that in nmp.

deathcap:
:+1: for native ws in nmp. Currently wsmc is [broken](https://github.com/deathcap/wsmc/issues/19), I think the relevant bits could be easier to maintain in sync with this repository if they were a part of it.

deathcap:
Or maybe, refactor node-minecraft-protocol to operate on node.js [streams](https://nodejs.org/api/stream.html#stream_stream) instead of TCP sockets? References:

TCP: https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/createClient.js
Stream: https://github.com/deathcap/wsmc/blob/master/minecraft-protocol-stream.js

Since streams are an abstract interface, using them with websockets is as simple as piping to [websocket-stream](https://www.npmjs.com/package/websocket-stream). Although some options may still be needed to control various other tweaks (e.g. for websockets you likely want to disable MC protocol encryption, since it is slow in JavaScript and wss:// URLs (analogous to https://) can be used instead for encryption).

https://github.com/substack/stream-handbook

roblabla:
NPm already operates on streams. All you have to do is call setConnection
with your stream?

deathcap:
But the `createClient()` API [accepts](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/createClient.js#L32) `host` and `port` options, which it passes to `net.connect()` to get a stream for `setSocket()` — I'd like to [pass](https://github.com/deathcap/wsmc/blob/2a4416adcd39a2403a3a73cf1adc947ec1aaa5e8/minecraft-protocol-stream.js#L33) createClient a stream directly.

rom1504:
Being able to pass a stream would be useful for https://github.com/PrismarineJS/node-minecraft-protocol/issues/185 too.

roblabla:
[...]

PR #316, opened 2016-01-17
Send framed packets in one push()

deathcap:
Before the streams refactor, MC packets were sent with the packet length prefix and payload combined in one call. This PR restores this behavior, instead of sending the packet length and payload in two separate `this.push()` calls.

ref: https://github.com/deathcap/wsmc/pull/21#issuecomment-172284334

(With this change, browserified mineflayer now works with the latest node-minecraft-protocol)

### DIFF
--- a/src/transforms/framing.js
+++ b/src/transforms/framing.js
@@ -15,10 +15,10 @@ class Framer extends Transform {
   }
 
   _transform(chunk, enc, cb) {
-    var buffer = new Buffer(sizeOfVarInt(chunk.length));
+    var buffer = new Buffer(sizeOfVarInt(chunk.length) + chunk.length);
     writeVarInt(chunk.length, buffer, 0);
+    chunk.copy(buffer, sizeOfVarInt(chunk.length));
     this.push(buffer);
-    this.push(chunk);
     return cb();
   }
 }


deathcap:
Marking this as WIP, investigating potentially better solutions.

deathcap:
I went with disabling MC packet framing instead (not needed when tunneling over websockets since they have their own framing, https://github.com/deathcap/wsmc/pull/21#issuecomment-172285314), works well and requires no changes to this project. 

There's still the question of whether two push()'s vs one push() has any effect on normal node-minecraft-protocol usage, but I think the TCP stack will coalesce multiple small writes into one larger packet by default, unless `this.socket.setNoDelay` (`TCP_NODELAY`) is set, so packets should not normally be fragmented. This could be tested further, but if they were fragmented (into 1-byte length packet + n-byte payload packets), the other side should still behave correctly, but there could be a performance/latency impact. Relevant:

https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/HISTORY.md#073

> 0.7.3
> * revert removing socket delays to reduce latency as it was causing errors and test failures.
> * server: Client now emits more predictable 'end' events.

deathcap:
Reopening for reconsideration in the normal TCP socket environment (not using WS), per https://github.com/PrismarineJS/node-minecraft-protocol/issues/312#issuecomment-172660685, re: @roblabla 

> we actually send two TCP frames, one with the ID alone, one with the content. I consciously chose this on-the-wire overhead over recreating a buffer for the sole purpose of doing only a single, clean write.

Surprised to hear this, it is actually faster in practice? I would think the network overhead of two packets would be greater than the memory/cpu overhead of coalescing them into one, but haven't benchmarked it.

If the overhead of copying is great, another option might be to cork and uncork the stream — for some reason this wasn't having any effect in my testing with websockets, but it might work with TCP sockets (would need further testing).

(If there was only one write per MC packet, this also could be another option besides having an option to disable MC packet framing for WS (attempting in https://github.com/PrismarineJS/node-minecraft-protocol/pull/320 but running into some abstraction/layering difficulties - single writes may be more straightforward).)

roblabla:
Fun Facts : 

packet type | overhead server-client | overhead client-server
------------------|-------------------------------|---------------------------------
Empty TCP Packet + IP header | 64 | 64  
WS Packet <126 bytes | 6 | 2
WS Packet <64 kb |8 |4
WS Packet <2**63 kb |12 |8

Anyway, the CPU time overhead is probably negligeable. It'd be nice to benchmark it though.

deathcap:
```javascript
const Benchmark = require('benchmark');
const suite = new Benchmark.Suite;
const Transform = require('stream').Transform;

// https://github.com/roblabla/ProtoDef/blob/master/src/datatypes/utils.js
function sizeOfVarInt(value) {
  var cursor = 0;
  while(value & ~0x7F) {
    value >>>= 7;
    cursor++;
  }
  return cursor + 1;
}
function writeVarInt(value, buffer, offset) {
  var cursor = 0;
  while(value & ~0x7F) {
    buffer.writeUInt8((value & 0xFF) | 0x80, offset + cursor);
    cursor++;
    value >>>= 7;
  }
  buffer.writeUInt8(value, offset + cursor);
  return offset + cursor + 1;
}

const chunk = new Buffer('aabbccdd', 'hex');

suite.add('copy buffer', () => {
  const s = new Transform();
  var buffer = new Buffer(sizeOfVarInt(chunk.length) + chunk.length);
  writeVarInt(chunk.length, buffer, 0);
  chunk.copy(buffer, sizeOfVarInt(chunk.length));
  s.push(buffer);
})
.add('push twice', () => {
  const s = new Transform();
  var buffer = new Buffer(sizeOfVarInt(chunk.length));
  writeVarInt(chunk.length, buffer, 0);
  s.push(buffer);
  s.push(chunk);
})
.add('buf concat', () => {
  const s = new Transform();
  var buffer = new Buffer(sizeOfVarInt(chunk.length));
  writeVarInt(chunk.length, buffer, 0);
  s.push(Buffer.concat([buffer, chunk]));
})
.on('cycle', (event) => {
  console.log(String(event.target));
})
.on('complete', function() {
  console.log('Fastest is ' + this.filter('fastest').map('name'));
})
.run({ 'async': true });
```

An attempt at a microbenchmark. Results on my machine:

```shell
$ npm install benchmark
$ node buf.js
copy buffer x 268,775 ops/sec ±7.29% (68 runs sampled)
push twice x 271,827 ops/sec ±5.20% (72 runs sampled)
buf concat x 239,097 ops/sec ±3.90% (75 runs sampled)
Fastest is push twice,copy buffer
```

so `Buffer.concat([buffer, chunk])` is much slower, but `  chunk.copy(buffer, sizeOfVarInt(chunk.length)); s.push(buffer);` versus `s.push(buffer); s.push(chunk);` is about the same within a margin of error

deathcap:
I'd be more concerned about the network overhead though. Wireshark shows they are sent in different packets:

screen shot 2016-01-31 at 6 27 22 pm

Each packet adds overhead, additional headers and it also has to be acknowledged (or that may not necessarily be true, there are various optimizations in TCP, delayed acks, selective acks, etc. but I don't know much about them, either way it is overhead). Overhead of smaller packets may be a worthwhile tradeoff in some situations, e.g. if the packet is for interactive user input or streaming media where low latency is required, but in this specific case, the packet is a length field for the next packet. The server can't do anything without the whole packet, so it just has to wait for two packets now.

Example packet trace before (testing connecting to an outdated server, for a shorter packet capture, using examples/client_echo/client_echo.js):

screen shot 2016-01-31 at 6 30 38 pm

After, with this change:

screen shot 2016-01-31 at 6 31 04 pm

I would also like to disabling framing for websockets, as an optimization, but nonetheless I think this PR itself would be very worthwhile for both TCP (to reduce overhead) and WS. Rebased on master.

PR #319, opened 2016-01-18
Use native node.js ES6 support

deathcap:
Now that ES6 is gaining more widespread adoption, including in Node and modern browsers, this PR attempts to use in node-minecraft-protocol natively, instead of through gulp/babel.

Requires enabling experimental in-progress V8 flags, so this is probably premature, but might be worth doing at some point (unless we want to always use newer ES features through babel, e.g., ES7 when ES6 is well-supported, then ES8 when ES7, etc.).

Node.js ES6 documentation: https://nodejs.org/en/docs/es6/


rom1504:
https://github.com/alekseykulikov/babel-preset-es2015-node5 might be a way to reduce compilation until everything needed is supported in node.

Now idk if we want to use only es-current or also es-next as we currently do.

deathcap:
Note: not sure why the tests fail.

https://github.com/alekseykulikov/babel-preset-es2015-node5 looks like a pretty good option. Only a few ES6 features are missing from Node 5, and all that NMP would currently need, as far as I can tell, is transform-es2015-destructuring (comparable to --harmony_destructuring) and transform-es2015-parameters (comparable to --harmony_default_parameters).

Babel's transform-es2015-destructuring is also more complete than Node v5.4.1's --harmony_destructuring, which annoyingly does not yet support destructuring assignments outside of variable declarations (hence https://github.com/deathcap/node-minecraft-protocol/commit/17f4c65cdfd8c3c43a5181a2dd0593bc99d30744 - Node errors out on `({ a, b } = c)`, even though it is legal ES6 syntax, destructuring support is incomplete/in-progress. On the other hand, `let { a, b } = c` works).

> Now idk if we want to use only es-current or also es-next as we currently do.

I'd be inclined to favor only es-current (or less), since using es-next introduces additional fragility and risk by depending on experimental features that could change at any time. Keeping up with the moving target of the MC protocol is already difficult enough =) the increased burden of tracking ES7 changes also adds to the maintenance cost.

I think the only ES7 feature NMP relies on is class field declarations: https://github.com/deathcap/node-minecraft-protocol/commit/aa722ab3af14ee20d84612718c88760f9a6599af - or is there anything else? Looks like the currently-ES7-based NMP could be ported to ES6-only without too much trouble, if this change is desired. @roblabla @wtfaremyinitials what do you think? My 2¢, while ES7 has some neat experimental features, relying on them has a cost which ought to be carefully considered, especially for a library depended on by other projects. I could PR the ES7-removal separately if there is interest.

---

There's also the potential benefit to consider of removing the added complexity of using babel, gulp, and a build process in NMP. I haven't tracked it down fully, but I've been seeing some strange issues with source maps in projects using NMP (ref https://github.com/deathcap/wsmc/issues/16) - which I haven't seen when NMP is removed, or with projects using babel through the [babelify](https://github.com/babel/babelify) browserify transform instead of a gulp build task. Not to say these issues can't or shouldn't be investigated fixed, but removing additional build tasks reduces complexity and leaves fewer opportunity for issues to arise (kiss principle), less to go wrong and bugs to track down.

roblabla:
Well, if we can remove the whole build process and use NMP **without any flags** (e.g. no --harmony), then that seems like a very reasonable change. This does mean we're breaking compat with nodejs 0.12 and potentially 4.0 though.

I don't like the situation surrounding build process and npm at all. NPM cares very little about the problems surrounding this (https://github.com/npm/npm/issues/3055 is still open, years after the fact), sourcemaps are not supported natively (have to use source-map-support, which is up to the API consumer, not library author, to avoid conflicts), etc...

Besides, es7 is still too far off and is only marginally useful to us (I mostly use decorators, initialisers, and async/await). Async/await has a few very real problem (async generators are still not a thing) making building abstractions on top of it not viable.

So yeah. I'm very much OK with that.

roblabla:
Concerning es-current vs es-next. If babel is here to stay, I am in favor of using es-next. If features disappear, that will be a MAJOR version bump in the babel plugin transform, so a simple npm install will always work as is. We'd just have to eventually update the code to bump the babel version.

Using babel just for es-current would be adding a build step "for the sake of it".

rom1504:
It doesn't seem like `--harmony_destructuring --harmony_default_parameters` are mocha parameters, I can't find them in the doc (which probably explain why the tests are failing)
And yeah requiring node flags is not ok (those exist just for testing).
But I'm also for getting rid of the compilation step if we can as we don't really need any es7 in nmp.
That doesn't seem possible right now because we need these flags but maybe using https://github.com/alekseykulikov/babel-preset-es2015-node5 (along with babel6), and requiring node5,  and then removing it when enough things are in node5 would be a good path.

deathcap:
@roblabla 
> if we can remove the whole build process and use NMP **without any flags** (e.g. no --harmony), then that seems like a very reasonable change

I'd be all for this, but I think the only way it is currently possible is by removing some of the ES6 usage in NMP. 

At one extreme is the nuclear option of reverting everything to ES5, but it seems some ES6 features are safe to use on most modern platforms. There is not terribly much usage of the unsupported features:

Default parameters:

```javascript
src/transforms/compression.js:14:  constructor(compressionThreshold = -1) {
src/transforms/compression.js:44:  constructor(compressionThreshold = -1) {
src/transforms/serializer.js:39:function createSerializer({ state = states.HANDSHAKING, isServer = false , version} = {})
src/transforms/serializer.js:48:function createDeserializer({ state = states.HANDSHAKING, isServer = false,
```

Destructuring:
```
src/transforms/compression.js
50:    var { size, value, error } = readVarInt(chunk, 0);

src/transforms/framing.js
35:    var { value, size, error } = readVarInt(this.buffer, offset) || { error: "Not enough data" };
40:      ({ value, size, error } = readVarInt(this.buffer, offset) || { error: "Not enough data" });

src/transforms/compression.js
1:var [readVarInt, writeVarInt, sizeOfVarInt] = require("protodef").types.varint;

src/transforms/framing.js
1:var [readVarInt, writeVarInt, sizeOfVarInt] = require("protodef").types.varint;
```

@rom1504 
> It doesn't seem like `--harmony_destructuring --harmony_default_parameters` are mocha parameters

They are `node` parameters, shown with `node --v8-options`. There's a bunch more, many default on. I agree with not using these flags, though.

@rom1504 
> But I'm also for getting rid of the compilation step if we can as we don't really need any es7 in nmp.

@roblabla 
> Besides, es7 is still too far off and is only marginally useful to us

Sounds good to me, as a first step I propose removing ES7 / stage 0 proposals, I'll open a separate PR: https://github.com/PrismarineJS/node-minecraft-protocol/pull/321

package.json
```diff
@@ -2,14 +2,14 @@
...
+  "main": "src/index.js",
   "repository": {
     "type": "git",
     "url": "git://github.com/PrismarineJS/node-minecraft-protocol.git"
   },
   "scripts": {
-    "prepublish": "gulp && require-self",
-    "test": "mocha --recursive --require intelli-espower-loader --require source-map-support/register --reporter spec"
+    "prepublish": "require-self",
+    "test": "mocha --harmony_destructuring --harmony_default_parameters --recursive --require intelli-espower-loader --require source-map-support/register --reporter spec"
```
rom1504: This is where I meant these aren't mocha options AFAIK. Except if mocha automatically passes all options to node ?

deathcap: mocha passes all --harmony flags to node:

```
node-minecraft-protocol $ ./node_modules/.bin/mocha --help
[…]
    --harmony<_classes,_generators,...>     all node --harmony* flags are available
```

deathcap:
Updated to remove the experimental v8 flags, tests now pass. This does unfortunately remove the ability to use default parameters and destructuring.

rom1504:
full es6 support is coming soon to node (see chrome score there https://kangax.github.io/compat-table/es6/ : 92%), so I think we should wait for that before stopping to use babel.
Using half of es6 doesn't quite seem right.

deathcap:
Closing for now, I still see the babel/gulp build process as a pain point, but I guess I'll see about improving the specific issues, and/or wait until full ES6 in v8/node/chrome. For my new projects at least, I think I'll stick to the currently well-supported ES6 subset (without babel/gulp), but I agree for node-minecraft-protocol, reverting from full ES6 to a smaller subset seems like a step backwards. Could be revisited once ES6 is mode widespread.

deathcap:
v8 release 4.9 supports destructuring and default parameters: http://v8project.blogspot.com.au/2016/01/v8-release-49.html - coming to nodejs soon,,

deathcap:
Reopened and rebased, I think this is worth reconsidering. Wrote up this list of why, in my humble opinion (feedback welcome):

https://gist.github.com/deathcap/89975d8ede56f27d83e7 My Big List of Annoyances with Babel/ES6 Transpilation

roblabla:
I find most of the problems are unfounded.

Babel is far from diverging from vanilla JS. The only feature that has been reopened for discussion thus far is decorators. The  `const` thing will error out at runtime, babel didn't pick up the error because it didn't transpile it at all (you probably have the es2015 plugin).

We can use the native ES6 support with the node5 babel plugin, which disables the unnecessary transforms.

The dev-code distribution issue, well... Modifying code directly in node_modules is simply not supported. If you do this, we assume you know what you're doing (and as such, are aware of having to run the build scripts). **A user should never have to modify what's under node_module-s, ever**.

There is the small problem of prepublish scripts not being run when installing from git. That's an unfortunate bug with npm. We could fix it by having the transpiled code in the repo, but that solution is ugly. Another potential solution is a postinstall script running the builders.

The watch tool should always be next to you on your screen when you're coding, so when you save, you see the errors. Again, this is a "know-your-tools" problem.

I agree the rebuild is too slow. There are probably things we could do to improve this. We should take a look at how other big projects handle babel.

And yeah, sourcemaps sucks. I pray they will suck less one day. For now, they're decent enough anyway. We should have the tests import source-map-support so at least the tests contain the right lines. We could also make the tests transpiled by mocha.

Removing babel would be a step in the wrong direction. Until ES6 is fully implemented (with destructuring assignments, default function arguments, rest params, let/const, generators) and doesn't have shitty performance due to deoptimization passes, it's a bad idea to remove it.

deathcap:
Haven't considered ES6 vs ES5 performance but that's a good point. Found this article with some more details: http://incaseofstairs.com/2015/06/es6-feature-performance/ (from 2015, benchmarking up to node 2 and chrome 45, but transpiled ES6 vs native ES5 analogues), also some newer microbenchmarks on node 4 and 5: https://github.com/bevry/esnext-benchmarks. It does look like ES5 is generally faster, could be interesting to see how the performance difference applies to node-minecraft-protocol, may be a good reason to stick with it (transpiled ES5 vs native ES6) if it is significantly faster.

rom1504:
One argument :  getting rid of the (at least) 5k files gulp/babel install. (https://github.com/PrismarineJS/flying-squid/issues/212)

rom1504:
one month to go (https://github.com/nodejs/LTS#lts_schedule)  and we'll have node6 with everything we need here (http://node.green/)

rom1504:
People at #node.js say we should either stay at es5 forever either keep a build system forever.
I disagree but it seems like that's the opinion in some circles.

rom1504:
node6 is out. Would be interesting to see if it can run nmp without the babel transformation

deathcap:
Node 6 has ~9x% ES6 support, but IIRC no ES6 modules yet (shouldn't be a problem as long as https://github.com/PrismarineJS/node-minecraft-protocol/pull/399 import/export replacing require isn't merged). Also curious how Chrome stable tracks V8 updates (would it get V8 updates approximately in sync with Node, or does Node current update faster? As I'd also want to run NMP in the browser, etc.)

roblabla:
chrome://version/ to figure out v8 version. Chromium 50 is V8 5.0.71.32 (which is 3 minor behind v8, probably bugfixes).

deathcap:
Cool thanks. Looks like Chrome 50 was released last week https://chrome.googleblog.com/2016/04/chrome-50-releases-and-counting.html so ~90% of ES6 should now be safe to use, at least on modern platforms. I'll update this PR, the changes keep shrinking, since V8 5.0 now supports default function parameters and destructuring, at last.

deathcap:
As part of this PR I originally added `use strict` to all source files, since some browsers (at least Safari) did not allow var/let except in strict mode, but strictly speaking this change is not needed. The unmodified source of node-minecraft-protocol now runs on node without babel transpilation, as of v6.

deathcap:
Now that the same code works under both babel and without it, an interesting possibility is supporting both. Native ES6 is what I'd use for better browserify support avoiding the source maps issues etc., but those running servers may want to use ES5 with babel for better performance, until ES6 catches up. Not sure how good of an idea this is supporting two "versions" of NMP, but it appears to already be possible:

```
> require('minecraft-protocol').createClient.toString()
'function createClient(options) {\n  assert.ok(options, "options is required");\n  assert.ok(options.username, "username is required");\n\n  // TODO: avoid setting default version if autoVersion is enabled\n  var optVersion = options.version || require("./version").defaultVersion;\n  var mcData = require("minecraft-data")(optVersion);\n  if (!mcData) throw new Error(\'unsupported protocol version: \' + optVersion);\n  var version = mcData.version;\n  options.majorVersion = version.majorVersion;\n  options.protocolVersion = version.version;\n\n  var client = new Client(false, version.minecraftVersion, options.customPackets);\n\n  tcp_dns(client, options);\n  caseCorrect(client, options);\n  if (options.version === false) autoVersion(client, options);\n  setProtocol(client, options);\n  keepalive(client, options);\n  encrypt(client, options);\n  play(client, options);\n  compress(client, options);\n\n  return client;\n}'
> require('minecraft-protocol/src/index').createClient.toString()
'function createClient(options) {\n  assert.ok(options, "options is required");\n  assert.ok(options.username, "username is required");\n\n  // TODO: avoid setting default version if autoVersion is enabled\n  const optVersion = options.version || require("./version").defaultVersion;\n  const mcData=require("minecraft-data")(optVersion);\n  if (!mcData) throw new Error(`unsupported protocol version: ${optVersion}`);\n  const version = mcData.version;\n  options.majorVersion = version.majorVersion;\n  options.protocolVersion = version.version;\n\n  const client = new Client(false, version.minecraftVersion,options.customPackets);\n\n  tcp_dns(client, options);\n  caseCorrect(client, options);\n  if (options.version === false) autoVersion(client, options);\n  setProtocol(client, options);\n  keepalive(client, options);\n  encrypt(client, options);\n  play(client, options);\n  compress(client, options);\n\n  return client;\n}'
```

roblabla:
Well, if we go this route, we should keep babel around.

Most projects that support both follow the convention that `nmp` points to ES5, `nmp/es6` (so an es6.js file at root) points to the original source.

The cool thing about using unmodified source is that it makes debugging much easier since we don't have sourcemaps problems 'n stuff.

mhsjlw:
It is worth noting that transpiring does have a *very slight* performance increase, because babel transpiles to more compatible JavaScript the v8 is able to optimize it to make it run more efficiently. Although it probably isn't that big of a deal. After all, we're all waiting for partial packet editing in ProtoDef ;)

deathcap:
@mhsjlw yes see https://github.com/PrismarineJS/node-minecraft-protocol/pull/319#issuecomment-186002342, at least last time I've benchmarked it there was a 1-2% slowdown in NMP's unit tests from removing transpiling. Not too much, but it is measurable.

For what it's worth, I've removed transpiling from all my projects, for various reasons (posted a bit about it here https://medium.com/@deathcap1/voxel-js-modules-now-100-transpile-free-f8df34ea74b5)

Another option may be to switch to a lightweight transpiler, such as [Bublé](https://buble.surge.sh/guide/). It has some differences from Babel, some interesting advantages for example you can specify what platforms to target:

```shell
buble input.js --target chrome:48,node:5,safari:9,edge:12 -o output.js
```

and it will only transpile the ES6 features not supported by the platforms specified (although each transform can also be toggled on/off independently), all the while preserving the source code as much as possible.

NMP could use Bublé to explicitly define what targets it supports (currently node:4, but I at least would be fine with: chrome:50,node:6), and as they change the now-natively-supported ES6 features would drop out of transpilation. And if there are certain ES6 features that are significantly faster when transpiled to ES5, as proven by benchmarks, those transforms can be toggled on an as-needed basis.

rom1504:
node6 has now (almost) full es6 support.
merged at https://github.com/PrismarineJS/node-minecraft-protocol/commit/3ff7ac64c12f4bc99d2605c7527e7518ac2b3502

Issue #176, opened 2015-05-22
Remaining improvements for the new transform streams


rom1504:
* [ ] using a pipeline module : 
 * easily add/remove stream
 * each step is 2 transform stream
 * redirect the error event of streams to the pipeline
 * java inspiration  : http://netty.io/5.0/api/index.html + ctrl+f ChannelPipeline
 * node module that does a part of that https://github.com/juliangruber/multipipe
* [ ] move createPacketBuffer and parsePacketData from protocol.js to serializer.js
* [ ] client.js should somehow be a class used by both client.js and server.js

rom1504:
After rereading what client.js and server.js do, I think it does make sense to use client.js in server.js : server.js is a Server class that handle Client instances. So I'm checking that third point as done.

rom1504:
What's remaining is now only the pipeline module. That has been started here https://github.com/wtfaremyinitials/stream-pipeline but isn't quite done.
Renaming the issue as "Using a stream pipeline module"

deathcap:
https://github.com/wtfaremyinitials/stream-pipeline has been inactive 8 months and is not published to NPM, but there is another module on NPM of the same name: https://www.npmjs.com/package/stream-pipeline - "When you want to connect several disparate pipes in a configuration driven way use this tool"

but wtfaremyinitials/stream-pipeline under the hood uses the `pipe` and `unpipe` [node.js Stream API calls](https://nodejs.org/api/stream.html#stream_readable_unpipe_destination) why not just use those?

rom1504:
To avoid doing that much piping and unpiping (and error relaying and all) in  https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client.js and to make it easy to add/remove a stream in the pipeline.

wtfaremyinitials:
@deathcap Hiya! I'm still here and alive, just super busy with school.

Issue #185, opened 2015-05-29
Document SOCKS and HTTP proxying mechanism


roblabla:
The current way to add proxying into node-mc-proto : 

``` js
var mc = require("minecraft-protocol");
var Http = require("http");
var proxy = { address: '192.168.1.1', port: 80 };
mc.Client.connect = function(port, host) {
  var self = this;
  var req = Http.request({
    host: proxy.address,
    port: proxy.port,
    method: 'CONNECT',
    path: host + ":" + port
  });
  req.on('connect', function(res, socket, head) {
    this.setSocket(socket);
  });
  req.end();
};
```

A better way to do this would be nice

rom1504:
A better way might be adding a Client.proxyConnect that would do that. And add an option in createClient to pass a proxy.

roblabla:
i thought about having createClient take a socket. So we just bypass the connect and just call setSocket with the given socket. Then it's the user's responsibility to create the socket connection.

rom1504:
As discussed in gitter, that'd work provided that the socket is optional (let's not break createClient and break everything built on top of nmp) and that we would create the socket as currently done if the socket isn't passed.

roblabla:
Thought about this a bit more. What we need to do is separate createClient into two function : 

- handshakeClient(client), which does all the necessary steps to transition from HANDSHAKE to PLAY states
- createClient(), which creates a TCP connection and passes it to handshakeClient.

This way, people who want to use any other kind of stream (a File stream for protocol analysis, or a proxy stream) are able to do so without difficulty.

rom1504:
https://github.com/mattcg/socks5-client would probably do for a socks5 proxy

roblabla:
This is now easily possible thanks to #320. We just need to document the mechanism to do so. 

The following code is untested, use at your own risk.

## Sample code for HTTP proxy 

```javascript
var Http = require("http");
var minecraft = { host: "the.minecraft.server.host.com", port: 25565 };
var proxy = { host: '192.168.1.1', port: 80 };
var req = Http.request({
  host: proxy.host,
  port: proxy.port,
  method: 'CONNECT',
  path: minecraft.host + ":" + minecraft.port
});
req.on("connect", function(res, stream, head) {
  require("minecraft-protocol").createClient({
    stream: stream,
    username: "bleh",
    password: "bleh"
  });
});
```

## Sample code for Socks Proxy

```javascript
var Socks  = require("socks5-client");
[...]

PR #320, opened 2016-01-18
Add 'stream' option to client, generic stream alternative to TCP/IP host/port options

deathcap:
To support https://github.com/PrismarineJS/node-minecraft-protocol/issues/312 WebSockets, a variation of `createClient()` recognizing `options.stream`, set to an arbitrary Node.js Stream. To do:

* [x] createClientStream() moved from wsmc/minecraft-protocol-stream.js, functional parity (working in concert with with https://github.com/deathcap/wsmc/pull/26)
* [x] Refactor createClient() to use createClientStream()
* [x] Test - CircleCI unit tests pass, also works in my own testing with wsmc
* ~~Fix packet framing disabling hack (`noPacketFramer` option, client.framer = EmptyTransformStream, has no `.unpipe()` method - surely there must be a cleaner means to achieve disabling framing)~~ removed in favor of https://github.com/PrismarineJS/node-minecraft-protocol/pull/316 and/or https://github.com/PrismarineJS/node-minecraft-protocol/issues/176


deathcap:
This PR needs to be updated for https://github.com/PrismarineJS/node-minecraft-protocol/pull/333 client modularization, but should be much simpler now. The socket is created in the src/client/tcp_dns.js plugin, which calls setSocket(). Proposal: add a new `stream` option recognized by createClient, which would directly setSocket() and skip loading src/client/tcp_dns.js

deathcap:
Updated, this is only a few lines now. Added a 'stream' option. Tested with:

```javascript
require('./').createClient({stream: net.connect(25565,'localhost'),username:'foo'})
```

rom1504:
Pretty simple. Could be used to solve #185 (by adding an example using this stream option to use a proxy)

PR #321, opened 2016-01-19
Remove experimental ES7 features, use only standard ES6

deathcap:
https://github.com/PrismarineJS/node-minecraft-protocol/pull/319


rom1504:
Okay, let's do that. That doesn't change much anyway.

PR #322, opened 2016-01-19
[WIP] Babel6 node5

rom1504:
similar to https://github.com/PrismarineJS/node-minecraft-protocol/pull/305 , but with babel-preset-es2015

nicer dist/

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   environment:
     MC_SERVER_JAR_DIR: /home/ubuntu/node-minecraft-protocol/minecraft-server/
   node:
-    version: 4
+    version: 5
   java:
     version: openjdk7
 dependencies:
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -3,7 +3,7 @@ var gulp = require('gulp');
 var plumber = require('gulp-plumber');
 var babel = require('gulp-babel');
 var options = {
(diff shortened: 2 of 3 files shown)


rom1504:
This adds the requirement to use (the latest) node5, I'm not 100% sure if we want to do that right now really, node5 is changing fairly fast.

roblabla:
Node changing fast isn't really the problem.

The biggest problem is whether node5 is available on debian/ubuntu. I believe the current nodejs available on debian is node4. So you'll have to tell those people how to install a newer version of node or live with them unable to use NMP :|

rom1504:
The newest version available in Debian/Ubuntu is 0.10
But people are already installing node by other ways AFAIK

deathcap:
[nvm](https://www.npmjs.com/package/nvm) (node version manager) is a good way to install node.js, and manage multiple versions. I use it but my concern is support, according to https://nodejs.org/en/ the "Mature and Dependable" long-term support node version is v4, so I think it would be prudent to support the LTS version to maximize long-term compatibility, unless there are compelling reasons to do otherwise.

rom1504:
Let's close this for now. It can be reopened later when es6 situation is clearer (related to https://github.com/PrismarineJS/node-minecraft-protocol/pull/319 )

PR #323, opened 2016-01-22
Show disconnect reason in examples

deathcap:
When the examples clients are disconnected, they currently silently exit. This PR adds the disconnect reason, with informative text explaining why the disconnection occurred, such as:

```
node-minecraft-protocol $ node examples/client_echo/client_echo.js localhost 25565
connected
disconnected: "This server requires FML/Forge to be installed. Contact your server admin for more details."
```

(note: there are quotes around the reason because the reason field in this packet http://wiki.vg/Protocol#Disconnect_2 is not a string but a 'chat' type, but I think it's good enough for an example)

### DIFF
--- a/examples/client_chat/client_chat.js
+++ b/examples/client_chat/client_chat.js
@@ -93,6 +93,10 @@ client.on('connect', function() {
   console.info(color('Successfully connected to ' + host + ':' + port, "blink+green"));
 });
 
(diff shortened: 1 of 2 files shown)


rom1504:
Ah that's the login-state disconnect packet, interesting.

PR #326, opened 2016-01-23
Forge client support (support integrating with node-minecraft-protocol-forge)

deathcap:
Add client-side https://github.com/PrismarineJS/node-minecraft-protocol/issues/114 Forge Support

* [x] Send tagged hostname in set_protocol packet
* [x] Receive server channel registrations (FML|HS, FML, FML|MP, FML, FORGE)
* [x] Handle the FML|HS ServerHello with the player's dimension
* [x] Register client plugin channels (FML|HS, FML, FML|MP, FML, FORGE)
* [x] Send the FML|HS ClientHello
* [x] Send client modList
* [x] Receive server modList
* [x] Check client/server mod compatibility - emits `forgeMods` event for app to check
* [x] Send HandshakeAck
* [x] Receive RegistryData packet(s) until hasMore=false
* [x] Send/receive final HandshakeAcks and CompleteHandshake
* [x] Refactor discriminator byte / 'FMLIndexedMessageToMessageCodec' to use a mapper
* [x] Refactor discriminator byte / 'FMLIndexedMessageToMessageCodec' parsing to use one big switch
* [x] Move FML|HS handshake to src/fml.js, support and document forge/forgeMods opts, event
* [x] Refactor into separate module?
* [x] Update for https://github.com/PrismarineJS/node-minecraft-protocol/pull/333 Modularization attempt of createClient()
* [x] Move tagHost=\0FML\0 to be installed by src/client/forgeHandshake.js, not the client example

Not included, out of scope for this PR:
* Fully parse the RegistryData packets and emit registry events - would be nice but only needed for advanced usage, can come later
* FORGE plugin channel (observed in packet logging, sends lava and water)
* FML and FML|MP plugin channels (registered, but not seen that they are used at least at login)
* Server-side Forge support https://github.com/PrismarineJS/node-minecraft-protocol/issues/114

Issue #152, opened 2015-04-05
Handle official (MC| - prefixed) Custom Payload packets


roblabla:
Minecraft defines a couple of packets through the Custom Payload mechanism. In 1.8, those are : 

`MC|AdvCdm`:

Name |   Type   | Description
---------|-----------|-------------
Mode         | Byte    |  
X               | Int        | If Mode == 0
Y               | Int        | If Mode == 0
Z               | Int        | If Mode == 0
EntityID    | Int        | If Mode == 1
Command | String |

`MC|Beacon`
`MC|BEdit`
`MC|BSign`
`MC|BOpen`
`MC|ItemName`
`MC|RPack`
`MC|TrList`
`MC|TrSel`
`MC|PingHost`
`MC|Brand`

Furthermore, the following two channels are reserved for channel management mechanisms : 

- `Register`
- `Unregister`

It may be a good idea to provide official serializers for the official channels, and perhaps provide a way for users to easily add serializers for their own custom channels.

rom1504:
See https://github.com/SpockBotMC/SpockBot/pull/172/files on that topic

rom1504:
This is needed to support command block updating.
An example of updating is:
```
client->server: play custom_payload : 
{"channel":"MC|AdvCdm","data":{"type":"Buffer","data":[0,0,0,0,0,0,0,0,71,255,255,255,189,17,47,116,112,32,114,111,109,49,53,48,52,32,48,32,48,32,48,1]}}
```

rom1504:
https://gist.github.com/rom1504/742a400592260329bcb1 : started to work on it

rom1504:
http://wiki.vg/Plugin_channel is more up to date

rom1504:
Some of that currently put in mineflayer, https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/command_block.js, it will move here when ready I think.

deathcap:
and/or to https://github.com/PrismarineJS/minecraft-data right? (protocol definition)

Plugin channels would also be useful for https://github.com/PrismarineJS/node-minecraft-protocol/issues/114 Forge Support  - FML/Forge has several of its own used during handshaking

rom1504:
yeah and minecraft-data

rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol/pull/423 brings parsing of channel to the client and server with registration/unregistration. Missing for this issue:
[...]

Issue #114, opened 2015-02-22
Forge Support


roblabla:
It has recently been suggested in the gitter that support for the forge protocol would be interesting. BungeeCord supports Forge servers and clients now, so it could be used as a reference.

mrfrase3:
Before I start, I should mention that I have been actively running large public forge servers since tekkit first came out, and have been a fairly silent member of this community for years now.
My two cents on this topic:
  - I support this, I actually looked at implementing this myself, but ran away when I saw just the forge connection protocol.
  - Before this is implemented, this project would first need to start supporting previous versions of minecraft again, really because most of the community is on 1.5.2, 1.6.4 and only just starting on 1.7.10. Additionally, forge servers aren't going to consider going any higher(1.8) until sponge is finished, due to cauldron being abandoned, which is going to be a while.
  - Last time I checked, the documentation is rather poor, especially for the earlier protocols back when they had the "just look at the source code" mentality.
  - It's worth noting that it took a fair bit of effort to get bungee to support forge, if I remember correctly, there was an unknown packet 4 that was causing a lot of issues. which leads to the next point.
  - Some mods are unpredictable, and some have tried implementing their own protocols, which can lead to unpredictable results. Whilst this isn't really an issue for a chatbot that can ignore these random packets, its worth considering.

Anyway, just thought I might weigh in, I'll keep tabs and lemme know if you need anything ;)

Corgano:
Just wanted to note, from what I have seen, most the community is on 1.7 or later versions (http://www.gametracker.com/games/minecraft/).

roblabla:
node-forge-protocol will be implemented in its own packet, in the following form : 

```js
module.exports = function injectForge(mc) {
  if (mc == null)
    mc = require('node-minecraft-protocol');
  if (mc.protocol.version == '1.7')
    return injectForge1_7(mc);
  else if (mc.protocol.version == '1.6')
    return injectForge1_6(mc);
  else
    return null;
}
```

This should avoid the problem of protocol support somewhat.

As for how horrible forge's protocol hack is, that's not a problem. I'm quite used to reading decompiled minecraft source by now :smile:.

[...]

PR #328, opened 2016-01-24
[WIP] Add plugin channel API to NMP

deathcap:
Adds the start of an API for "plugin channels" in node-minecraft-protocol

Use cases include:

* https://github.com/PrismarineJS/node-minecraft-protocol/issues/152 Handle official (MC| - prefixed) Custom Payload packets
* https://github.com/PrismarineJS/node-minecraft-protocol/issues/114 Forge Support
 * https://github.com/PrismarineJS/node-minecraft-protocol/pull/326 Forge client support - experimenting building on this API for `FML|HS` in internal PR: https://github.com/deathcap/node-minecraft-protocol/pull/2
* https://github.com/PrismarineJS/mineflayer/issues/361 Plugin channel API in mineflayer
 * [MC|AdvCdm](https://github.com/PrismarineJS/mineflayer/commit/24f24349261c5b5031ff4ec624fef670af3028a5), [MC|Brand](https://github.com/PrismarineJS/mineflayer/issues/253)

To do:

* [ ] Rework to use nodejs streams

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -95,6 +95,13 @@ False if it is a connection from client to server.
 
 Returns the internal nodejs Socket used to communicate with this client.
 
+### client.outgoingPluginChannels.write(channel, data)
+### client.outgoingPluginChannels.register(channels...)
+### client.outgoingPluginChannels.unregister(channels...)
+
+Writes a raw payload buffer to an outgoing plugin channel. Registering and unregistering
+channels notifies the other side.
+
 ### client.uuid
 
 A string representation of the client's UUID. Note that UUIDs are unique for
@@ -134,6 +141,11 @@ parameters.
 ### per-packet events
 
 Check out the [minecraft-data docs](https://prismarinejs.github.io/minecraft-data/?v=1.8&d=protocol) to know the event names and data field names.
+
+### per-channel events
+
+`client.incomingPluginChannels` emits events for each plugin channel.
+The raw payload buffer is passed as the first argument.
 
 ## Not Immediately Obvious Data Type Formats
 
--- a/doc/README.md
+++ b/doc/README.md
@@ -96,8 +96,6 @@ False if it is a connection from client to server.
 Returns the internal nodejs Socket used to communicate with this client.
 
 ### client.outgoingPluginChannels.write(channel, data)
-### client.outgoingPluginChannels.register(channels...)
-### client.outgoingPluginChannels.unregister(channels...)
 
 Writes a raw payload buffer to an outgoing plugin channel. Registering and unregistering
 channels notifies the other side.
--- a/src/createClient.js
+++ b/src/createClient.js
@@ -25,9 +25,6 @@ function createClient(options) {
 
   var client = new Client(false, options.majorVersion);
 
-  client.incomingPluginChannels = new IncomingPluginChannels(client);
-  client.outgoingPluginChannels = new OutgoingPluginChannels(client);
-
   tcp_dns(client, options);
   setProtocol(client, options);
   keepalive(client, options);
--- a/src/createClient.js
+++ b/src/createClient.js
@@ -9,6 +9,8 @@ var yggserver = require('yggdrasil').server({});
 var states = require("./states");
 var debug = require("./debug");
 var UUID = require('uuid-1345');
+var IncomingPluginChannels = require('./pluginChannels').IncomingPluginChannels;
+var OutgoingPluginChannels = require('./pluginChannels').OutgoingPluginChannels;
 
 module.exports=createClient;
 
@@ -45,6 +47,8 @@ function createClient(options) {
 
 
   var client = new Client(false,version.majorVersion);
+  client.incomingPluginChannels = new IncomingPluginChannels(client);
+  client.outgoingPluginChannels = new OutgoingPluginChannels(client);
   client.on('connect', onConnect);
   if(keepAlive) client.on('keep_alive', onKeepAlive);
   client.once('encryption_begin', onEncryptionKeyRequest);
--- a/src/pluginChannels.js
+++ b/src/pluginChannels.js
@@ -1,0 +1,69 @@
+var EventEmitter = require('events').EventEmitter;
+
+function serializeChannelList(channels) {
+  return new Buffer(channels.join('\0'));
+}
+
+function parseChannelList(buffer) {
+  return buffer.toString().split('\0');
+}
+
+class PluginChannels extends EventEmitter {
+  constructor(client) {
+    super();
+    this.client = client;
+    this.channels = [];
+  }
+}
+
+class IncomingPluginChannels extends PluginChannels {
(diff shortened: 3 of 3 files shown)


deathcap:
This could be enhanced further (let NMP deserialize the channels it knows about, have a mechanism to register custom codecs, etc.), but I think this basic implementation is good enough to integrate for now, allowing other code to build upon it - including some which may be better in separate modules (https://github.com/PrismarineJS/node-minecraft-protocol/pull/326)

deathcap:
Not sure of the best API design here, does it make sense to have separate incomingPluginChannels and outgoingPluginChannels objects, or would it be cleaner to have only one pluginChannels object, with both incoming and outgoing channel sets?

If so same object could support both receiving (emitted events, `on`) and sending (`write()`), but the client/server can have different plugin channels registered, so I think they should be kept separate to some extent. Open to ideas..

roblabla:
I'm fine with both side them being separate or kept together. They both have their own set of advantages.

The way I envision the API is like so : 

```
var forgeChannel = client.incomingChannel('SOME_CHANNEL_I_OWN');
forgeChannel.on('data', function(data) {
});
```

The first call to "incomingChannel" should result in a "REGISTER" event being sent, the others the same object should be retrieved from a cache.

This has the advantage of being fully compatible with the nodejs stream api. So I can pipe stuff from and to the channel without problems. Meaning if I want to deserialize the data, I can do

```
var forgeChannel = client.incomingChannel('SOME_CHANNEL_I_OWN');
forgeChannel.pipe(new Protocol([{ type: "byte", "name": "data" }]).createStream())
.on('data', function(data) {
  console.log(data.data);
});
```

(Probably not the correct way to create a Protocol stream but nvm that :P)

deathcap:
:+1: for nodejs stream api compatibility

deathcap:
Some notes about registration:

The REGISTER/UNREGISTER packets accept multiple null-delimited channel names, and FML/Forge does send them all at once, so I think supporting multiple simultaneous registrations/unregistrations at the API level may be worthwhile.

On the other hand, it doesn't seem registration has much effect in the software I've tested. FML/Forge lets you send packets on a channel without REGISTER, and it happily proceeds just fine. On our side, packets for non-registered channels are still forwarded, it would be an extra check to drop non-registered channel payloads, not really clear why they added registration/unregistration in the first place (maybe to allocate resources? to advertise the fact that the other side supports some feature? but they'll already ignore unknown channels..)

There's also the complication that built-in `MC|` channels do not need registration. For these reasons, maybe (un)registration should be explicit. Have the API allow listening on arbitrary channels (possibly multiple listeners per channel?), and register/unregister a list of channels as specified.

Anyways, I like the idea of creating channels using nodejs streams. This solves most of the API design problem, since nodejs streams are a standard API other modules can easily hook into. Maybe bidirectional streams, to easily allow sending/receiving on an outgoing/incoming channel of the same name. [stream.Duplex](https://nodejs.org/api/stream.html#stream_class_stream_duplex_1). Or keep them separate `Writable`/`Readable` for simplicity, not too strong of an opinion about either way.

todo: change to use streams and update nmp-forge

roblabla:
Hmm, a Duplex stream might make sense. Then the api would be reduced to a simple call like `client.channel('CHANNEL_NAME`);`, which is pwetty neat.

Concerning the whole registration fiasco, CraftBukkit, IIRC, forces registration, otherwise it will just ignore the packets. I think we should do this by the spec.

As far as registering multiple channels at once, I really don't see much gain in that. The overhead of the channel packet is very very small, we're going from 10 bytes ("REGISTER".length + packetId + len) to 1 byte overhead. Not a huge win in the grand scheme of things, especially since channels are REGISTER'd extremely rarely in the 99.999% case.

deathcap:
I suppose if Forge accepts its channel registration in multiple REGISTER packets, it could be acceptable for the API to not support multiple simultaneous channel registrations. Would need to test to be sure, but it is probably fine. Another quirk in Forge's channel registration is that it registers the "FML" channel twice (it sends `FML|HS\x00FML\x00FML|MP\x00FML\x00FORGE`), probably a bug I but I tried to match the behavior byte-for-byte in nmp-forge, granted maybe unnecessarily

roblabla:
Yeah, that is a bug. And i know multiple register packets work. Used to do
that back when i did mod dev, and it shouldn't have changed since then.

deathcap:
Started to convert to use a Duplex stream, but got stuck on implementing the `_read()` stream method from an event emitter. This looks useful:

https://github.com/substack/emit-stream - turn event emitters into streams and streams into event emitters

but not clear how it interacts with Duplex streams (as it returns a Stream). Having second thoughts about Duplex streams versus separate readable and writables. The WHATWG standard web stream project, which has Node.js developers participating to improve on the streams spec, decided against duplex streams in preference to pairs of streams on each end: https://github.com/whatwg/streams/blob/master/FAQ.md#what-are-the-main-differences-between-these-streams-and-nodejs-streams

> For the concept of something with a writable end and a readable end, we have "duplex streams." Duplex streams are really just containers for a pair of streams, one writable and one readable, stored in the properties writable and readable respectively.

roblabla:
For _read(), what you should do is something like

```js
var readable = new Readable();
readable._read = function() {
  // First, read everything currently in the cache.
  var areWeDone = false;
  while (readable.readStuff.length > 0 && areWeDone) {
    areWeDone = this.push(this.readStuff.unshift());
  }
  // If we want more, wait until we get more
  if (!areWeDone)
    client.on('custom_payload', function bleh(data) {
       // When push returns false, we should stop reading until _read is called again.
      if (!this.push(this.readStuff.unshift()))
        client.removeEventListener('custom_payload', bleh);
    });
};

readable.readStuff = [];
client.on('custom_payload', function(data) {
  if (data.channel === "stuff")
    readable.readStuff.push(data);
});
```

deathcap:
Thanks, will add

deathcap:
Hmm, maybe it would be cleaner if the parsed protocol data was available as a Stream as well, (as opposed an event emitter)? 

This would avoid the need for the plugin channel API to translate event emitter to stream (which could also probably be done with https://github.com/substack/emit-stream).

And then the plugin channel code could simply read the parsed packet stream, and [divert](https://github.com/voxel/voxel-clientmc/blob/master/mf-worker.js#L15-L30) custom_payload to its own per-channel streams.

rom1504:
It is available in client.deserializer I think. (But that stream is changed
when the state changed so you need to redo your piping when it changes,
maybe we could wrap this into a single stream so people can just pipe it
without worrying about it changing)

deathcap:
Closing this for now, I think a plugin channel API could be interesting but would need to nail down the use cases and API, maybe have a 'packet stream' for parsed packets available first, wrapping client.deserializer. Sort of like the other end of https://github.com/PrismarineJS/node-minecraft-protocol/pull/320 (streams in, and streams out)

PR #330, opened 2016-01-25
[WIP] Intra-release multi-protocol support

deathcap:
`createClient()` accepts a "version" option to specify a Minecraft release, which is used to lookup the dataset in [minecraft-data](https://github.com/PrismarineJS/minecraft-data) via [node-minecraft-data](https://github.com/PrismarineJS/node-minecraft-data) (translating it to a major version, but see https://github.com/PrismarineJS/node-minecraft-data/issues/13, e.g. 1.8.9 -> 1.8), in turn used to set the protocol version in the `set_protocol` packet handshake, limiting the multi-protocol support of NMP to one protocol per major release.

This PR adds another option `protocolVersion`, clients can use to specify the exact protocol version to use. Data will still be used from minecraft-data based on `version`, but since `version`/minecraft-data specifies a major release, this additional `protocolVersion` option will allow clients to use minor "sub-versions" within the major version.

---

(Note: the client app is on their own when it comes to handling minor version discrepancies. 15w40b/76 for example is not compatible with 16w03a/96 because they renumbered all the packets. minecraft-data will probably just support one snapshot version at a time, the latest since it was updated. But `protocolVersion` would still be useful for smaller protocol updates where packets are added/removed but mostly compatible otherwise.)

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -18,6 +18,7 @@ automatically logged in and validated against mojang's auth.
  * maxPlayers : default to 20
  * keepAlive : send keep alive packets : default to true
  * version : 1.8 or 1.9 : default to 1.8
+ * protocolVersion : optional protocol version number, if omitted inferred from `version`
 
 ## mc.Server(version)
 
--- a/src/createClient.js
+++ b/src/createClient.js
@@ -42,6 +42,7 @@ function createClient(options) {
   var optVersion = options.version || require("./version").defaultVersion;
   var mcData=require("minecraft-data")(optVersion);
   var version = mcData.version;
+  var protocolVersion = options.protocolVersion || version.version;
 
 
   var client = new Client(false,version.majorVersion);
@@ -91,7 +92,7 @@ function createClient(options) {
 
   function onConnect() {
     client.write('set_protocol', {
-      protocolVersion: version.version,
+      protocolVersion: protocolVersion,
       serverHost: host,
       serverPort: port,
       nextState: 2


roblabla:
I'm not sure I agree with this. Anything can occur within a proto version bump. If someone wants to do this, he can fork mc-data and edit the default protocolVersion. I don't think we should make it easier than that.

roblabla:
Note that we could support snapshots in mc-data by naming them by their snapshot them, e.g. 15w40b, and passing that "version" to NMP instead, instead of putting the "fake" 1.9.

rom1504:
Oh I released mcdata 0.18.0 providing `mcdata.versionsByMinecraftVersion` except I forgot I have to use it in the loader too. I guess I'll do that later.

> Note that we could support snapshots in mc-data by naming them by their snapshot them, e.g. 15w40b, and passing that "version" to NMP instead, instead of putting the "fake" 1.9.

Yes that is definitely possible with the new versions data in mcdata. (Both passing 1.8 and 1.8.8 could be supported that way). I'll add support for that in node-minecraft-data.

deathcap:
Yeah this is not really a good solution. Now that minecraft-data has protocolVersions I think using it is the right solution going forward. I guess most of the changes will be in node-minecraft-data.

But some datasets in minecraft-data will support multiple protocol versions, right? If so I'd want to ensure the node-minecraft-protocol handshake uses the exact protocol version I give it (presuming it is supported). The use case here is nmp-auto / GH-331, where it pings the server to get the precise protocol version to use, then uses it in the handshake. If the protocol version does not match exactly the server will refuse the connection so it's important to have this level of control, in some form.

roblabla:
I thing this is (again) another use-case for the modules. I'd imagine substituting the "handshake" module with a slightly modified one that sends the protocol version you want would be more appropriate. 

The thing is, with the monolithic and huge createClient, we already have way too many options. Adding more options only useful in the 0.001% case to an already somewhat bloated API sucks.

Instead, we should let the users redefine small parts of the big "createClient" function through some kind of module system, and have createClient just use some sane defaults.

deathcap:
Indeed, modularizing is becoming more compelling with the various new use cases (cross-version, Forge, auto-negotiation, websockets). I think ideally node-minecraft-protocol would mostly consist of small composable components for each piece of functionality. The minecraft-data/protodef refactor was a good start, also the compression/framing/serializer transforms, but much more would be possible with some judicious modularization. Relating https://github.com/PrismarineJS/node-minecraft-protocol/pull/210 NMP Modularization I'll put some more notes there in a few.

As for this issue, what it comes down to is minecraft-data supporting datasets, currently 1.8 and '1.9', with fixed version.json for each (protocol 47 and 76, respectively). Clients can specify a minor version but 1.8.9, 1.8.8, etc. will all get stripped down to 1.8. For 1.8.x this is not much of a problem since they're largely all 47. But this is not true in general. For example 1.6.1 and 1.6.2 have different protocols (73 and 74) even though they are both 1.6 major. Not that either are currently supported by minecraft-data but it could be a problem as new protocols are added. Translating the release versions to protocol versions through minecraft-data https://github.com/PrismarineJS/node-minecraft-data/issues/13 should help fix this (though the datasets in minecraft-data would also need appropriate support.).

deathcap:
Closing in favor of https://github.com/PrismarineJS/node-minecraft-data/issues/13 Update to use minecraft-data for toMajor()  and https://github.com/PrismarineJS/node-minecraft-data/issues/15 versionsByProtocolVersion index (although some fixes may be needed to NMP, can be handled separately if needed, not sure any are)

Issue #329, opened 2016-01-24
Missing ping timeout callback when server is running but not responding to pings


deathcap:
ping.js does not appear to always call its callback method, it will be called with an error when a) the port is closed (example: options `{host:'github.com', port:25565}`), b) when a protocol error occurs (example: `{host:'github.com', port:80}`), but not c) when the server is running but does not acknowledge the pings.

To test this with a server, from https://github.com/PrismarineJS/node-minecraft-protocol/issues/327#issuecomment-174310946 made this modification 

```javascript
diff --git a/src/createServer.js b/src/createServer.js
index 5b56fdd..ef2b84e 100644
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -39,7 +39,7 @@ function createServer(options) {
   server.on("connection", function(client) {
     client.once('set_protocol', onHandshake);
     client.once('login_start', onLogin);
-    client.once('ping_start', onPing);
+    //client.once('ping_start', onPing);
     client.on('end', onEnd);
 
     var keepAlive = false;
```

```shell
NODE_DEBUG=mc-proto node examples/server/server.js
```

Testing the ping client:

```javascript
> node-minecraft-protocol $ node
// connection timeout calls callback
> require('./').ping({host:'github.com', port:25565}, function(){console.log(arguments)})
undefined
[…wait a few minutes…]
> { '0': 
   { [Error: connect ETIMEDOUT 192.30.252.128:25565]
     code: 'ETIMEDOUT',
     errno: 'ETIMEDOUT',
     syscall: 'connect',
     address: '192.30.252.128',
     port: 25565 } }

// protocol error calls callback
> require('./').ping({host:'github.com', port:80}, function(){console.log(arguments)})
undefined
> { '0': { [Error: Deserialization error for status.toClient : Read error for name : 84 is not in the mappings value] field: 'status.toClient' } }
{ '0': { [Error: read ECONNRESET] code: 'ECONNRESET', errno: 'ECONNRESET', syscall: 'read' } }

// using above example (commenting out ping_start), never responds (waited 10+ minutes)
> require('./').ping({host:'localhost', port:25565}, function(){console.log(arguments)})
undefined
```

rom1504:
I guess the timeout time should be an option. What should be its default value ? (for example what value does the vanilla client uses ?)

ca1ek:
just got burned by this today, I guess I need to write my own library

rom1504:
[...]

PR #331, opened 2016-01-25
Add dynamic cross-protocol support

deathcap:
Implements the dynamic part of https://github.com/PrismarineJS/node-minecraft-protocol/issues/234 Cross version support

Ported from https://github.com/deathcap/node-minecraft-protocol-auto (which I can deprecate if this is merged)

To do:

* [x] Enhance createClient() to allow waiting/pausing so it can be reconfigured before connect -- src/client/setProtocol.js? (installs client.on('connect') listener)
* [x] Change the protocol negotiation to be supported with a createClient()-compatible API (vs createClientAsync)
* [x] Update to use minecraft-data protocolVersions
* [x] Move into src/client/ to act as plugin/middleware

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,3 +8,4 @@ machine:
 dependencies:
   pre:
     - mkdir minecraft-server
+    - git clone https://github.com/PrismarineJS/minecraft-data node_modules/minecraft-data/minecraft-data || true
--- a/circle.yml
+++ b/circle.yml
@@ -8,4 +8,3 @@ machine:
 dependencies:
   pre:
     - mkdir minecraft-server
-    - git clone https://github.com/PrismarineJS/minecraft-data node_modules/minecraft-data/minecraft-data || true
--- a/src/client/autoVersion.js
+++ b/src/client/autoVersion.js
@@ -28,7 +28,6 @@ function createClientAuto(options) {
   options.wait_connect = true; // don't let createClient / src/client/setProtocol proceed on socket 'connect' until 'connect_allowed'
   var client = createClient(options); // vanilla
   debug('pinging',options.host);
-  // TODO: refactor with DNS SRV lookup in NMP
   // TODO: detect ping timeout, https://github.com/PrismarineJS/node-minecraft-protocol/issues/329
   ping(options, function(err, response) {
     if (err) throw err; // hmm
--- a/src/client/autoVersion.js
+++ b/src/client/autoVersion.js
@@ -54,7 +54,7 @@ module.exports = function(client) {
 
     // reinitialize client object with new version TODO: move out of its constructor
     client.version = majorVersion;
-    client.setSerializer(states.HANDSHAKING);
+    client.state = states.HANDSHAKING;
 
     if (response.modinfo && response.modinfo.type === 'FML') {
       // Use the list of Forge mods from the server ping, so client will match server
--- a/src/client/autoVersion.js
+++ b/src/client/autoVersion.js
@@ -7,7 +7,7 @@ var assert = require('assert');
 var minecraft_data = require('minecraft-data');
 
 module.exports = function(client, options) {
-  options.wait_connect = true; // don't let src/client/setProtocol proceed on socket 'connect' until 'connect_allowed'
+  client.wait_connect = true; // don't let src/client/setProtocol proceed on socket 'connect' until 'connect_allowed'
   debug('pinging',options.host);
   var pingOptions = {host: options.host, port: options.port};
   // TODO: use 0xfe ping instead for better compatibility/performance? https://github.com/deathcap/node-minecraft-ping
--- a/src/client/autoVersion.js
+++ b/src/client/autoVersion.js
@@ -24,11 +24,11 @@ function protocolVersion2MinecraftVersion(n) {
 module.exports = function(client) {
   var options = client.options;
 
-  debug('creating client');
   options.wait_connect = true; // don't let src/client/setProtocol proceed on socket 'connect' until 'connect_allowed'
   debug('pinging',options.host);
-  // TODO: detect ping timeout, https://github.com/PrismarineJS/node-minecraft-protocol/issues/329
-  ping(options, function(err, response) {
+  var pingOptions = {host: options.host, port: options.port};
+  // TODO: use 0xfe ping instead for better compatibility/performance? https://github.com/deathcap/node-minecraft-ping
+  ping(pingOptions, function(err, response) {
     if (err) throw err; // hmm
     debug('ping response',response);
     // TODO: could also use ping pre-connect to save description, type, negotiate protocol etc.
--- a/src/client/autoVersion.js
+++ b/src/client/autoVersion.js
@@ -9,11 +9,10 @@ var minecraft_data = require('minecraft-data');
 // Get the minecraft-data version string for a protocol version
 // TODO: add to node-minecraft-data index (protocol to newest release, if multiple)
 function protocolVersion2MinecraftVersion(n) {
-  var usesNetty = n > 0;
+  var usesNetty = true; // for now, only Netty protocols are supported TODO: pre-Netty (beware, colliding protocol version numbers)
   for (var i = 0; i < minecraft_data.versions.length; ++i) {
     var version = minecraft_data.versions[i];
-    if (version.version === Math.abs(n) && version.usesNetty === usesNetty) {
-      console.log(version);
+    if (version.version === n && version.usesNetty === usesNetty) {
       return [version.minecraftVersion, version.majorVersion];
     }
   }
--- a/src/client/autoVersion.js
+++ b/src/client/autoVersion.js
@@ -31,8 +31,7 @@ module.exports = function(client) {
   ping(pingOptions, function(err, response) {
     if (err) throw err; // hmm
     debug('ping response',response);
-    // TODO: could also use ping pre-connect to save description, type, negotiate protocol etc.
-    //  ^ see https://github.com/PrismarineJS/node-minecraft-protocol/issues/327
+    // TODO: could also use ping pre-connect to save description, type, max players, etc.
     var motd = response.description;
     debug('Server description:',motd); // TODO: save
 
@@ -47,12 +46,9 @@ module.exports = function(client) {
     // even though it is in a format accepted by minecraft-data. Instead, translate the protocol.
     var [minecraftVersion, majorVersion] = protocolVersion2MinecraftVersion(protocolVersion);
     client.options.version = minecraftVersion;
-
-    // Use the exact same protocol version
-    // Requires https://github.com/PrismarineJS/node-minecraft-protocol/pull/330
     client.options.protocolVersion = protocolVersion;
 
-    // reinitialize client object with new version TODO: move out of its constructor
+    // Reinitialize client object with new version TODO: move out of its constructor?
     client.version = majorVersion;
     client.state = states.HANDSHAKING;
 
@@ -65,7 +61,7 @@ module.exports = function(client) {
       // TODO: modify client object to set forgeMods and enable forgeHandshake
       throw new Error('FML/Forge not yet supported');
     }
-    // done configuring client object, let connection proceed
+    // Finished configuring client object, let connection proceed
     client.emit('connect_allowed');
   });
   return client;
--- a/src/client/autoVersion.js
+++ b/src/client/autoVersion.js
@@ -6,9 +6,7 @@ var states = require('../states');
 var assert = require('assert');
 var minecraft_data = require('minecraft-data');
 
-module.exports = function(client) {
-  var options = client.options;
-
+module.exports = function(client, options) {
   options.wait_connect = true; // don't let src/client/setProtocol proceed on socket 'connect' until 'connect_allowed'
   debug('pinging',options.host);
   var pingOptions = {host: options.host, port: options.port};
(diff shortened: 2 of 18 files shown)


rom1504:
createClient doesn't create an "all ready" client: people will want to wait for the login packet for example.
So I think it would be okay to return a client synchronously, and do the ping and set protocol thingies inside of createClient "afterwards".
That would mean changes in Client, but it might be okay.
This would have the benefit of keeping just one createClient.

rom1504:
Err probably not login, that's server side, but there are many "ready" events in mineflayer (spawn, received inventory,...), so people wait for one of these before writing anything to the client.
It's also important to be able to client "error" event asap

deathcap:
I'd definitely like to maintain API compatibility with createClient, will have to investigate more to try to do this. The problem I was running into is that the client created by createClient would continue the protocol handshake even before I received the ping. Maybe need some way to "pause" the client object? Also to dynamically change the version and hostname after object construction.

rom1504:
I think to do it cleanly you'd have to change createClient so it would
(optionally) "wait" (listen to the right event,...) for the ping pong to be
over before starting the handshake. It's createClient doing the state
changing, Client only provide the functionality to do so.
Also I think it would be possible that way to use the same socket for
pinging and for the rest. The vanilla client probably doesn't use the same
socket, but I think it makes sense.
Yeah it would be needed to change the version after the pong. It is kind of
already being done anyway since mc.ping default to 1.8.

deathcap:
> Yeah it would be needed to change the version after the pong. It is kind of
already being done anyway since mc.ping default to 1.8.

I think that's problematic for other reasons, imho the 0xfe01 ping https://github.com/PrismarineJS/node-minecraft-protocol/issues/332 is superior, as it does not require a status handshake, nor a protocol version (and works on all 1.4.4+)

> I think to do it cleanly you'd have to change createClient so it would
(optionally) "wait" (listen to the right event,...) for the ping pong to be
over before starting the handshake. It's createClient doing the state
changing, Client only provide the functionality to do so.

Notes for testing this scenario:

edit src/createServer.js to comment out `client.once('ping_start', onPing);`

`node examples/server/server.js`

`NODE_DEBUG=mc-proto node examples/client_auto/client_auto.js localhost 25565`

client should hang (hits https://github.com/PrismarineJS/node-minecraft-protocol/issues/329 Missing ping timeout callback when server is running but not responding to pings) or timeout:

```
node-minecraft-protocol $ NODE_DEBUG=mc-proto node examples/client_auto/client_auto.js localhost 25565
MC-PROTO: 54954 creating client
MC-PROTO: 54954 pinging localhost
MC-PROTO: 54954 writing packet handshaking.set_protocol
MC-PROTO: 54954 { protocolVersion: 47,
  serverHost: 'localhost',
  serverPort: 25565,
  nextState: 1 }
MC-PROTO: 54954 writing packet status.ping_start
MC-PROTO: 54954 {}
```

then edit src/createClientAuto.js (from this PR), move the `createClient()` call outside of the ping callback directly into `createClientAsync()`, and rerun client_auto.js

now, the client will despite not receiving a ping (callback not executed), it will still begin to connect:

```
node-minecraft-protocol $ NODE_DEBUG=mc-proto node examples/client_auto/client_auto.js localhost 25565
MC-PROTO: 54970 creating client
MC-PROTO: 54970 pinging localhost
MC-PROTO: 54970 writing packet handshaking.set_protocol
MC-PROTO: 54970 { protocolVersion: 47,
  serverHost: 'localhost',
  serverPort: 25565,
  nextState: 2 }
MC-PROTO: 54970 writing packet login.login_start
MC-PROTO: 54970 { username: 'echo' }
MC-PROTO: 54970 writing packet handshaking.set_protocol
MC-PROTO: 54970 { protocolVersion: 47,
  serverHost: 'localhost',
  serverPort: 25565,
  nextState: 1 }
MC-PROTO: 54970 writing packet status.ping_start
MC-PROTO: 54970 {}
MC-PROTO: 54970 read packet login.compress
MC-PROTO: 54970 { threshold: 256 }
MC-PROTO: 54970 read packet login.success
MC-PROTO: 54970 { uuid: 'c39fa4c6-85a3-3582-8167-db3805dbe851',
  username: 'echo' }
MC-PROTO: 54970 read packet play.chat
MC-PROTO: 54970 { message: '{"translate":"chat.type.text","with":["Server","echo joined the game."]}',
  position: 0 }
MC-PROTO: 54970 read packet play.login
MC-PROTO: 54970 { entityId: 1,
  gameMode: 1,
  dimension: 0,
  difficulty: 2,
  maxPlayers: 127,
  levelType: 'default',
  reducedDebugInfo: false }
MC-PROTO: 54970 read packet play.position
MC-PROTO: 54970 { x: 0, y: 256, z: 0, yaw: 0, pitch: 0, flags: 0 }
MC-PROTO: 54970 read packet play.keep_alive
MC-PROTO: 54970 { keepAliveId: 2138214823 }
MC-PROTO: 54970 writing packet play.keep_alive
MC-PROTO: 54970 { keepAliveId: 2138214823 }
MC-PROTO: 54970 read packet play.keep_alive
MC-PROTO: 54970 { keepAliveId: 648251752 }
MC-PROTO: 54970 writing packet play.keep_alive
MC-PROTO: 54970 { keepAliveId: 648251752 }
MC-PROTO: 54970 read packet play.keep_alive
MC-PROTO: 54970 { keepAliveId: 1917937053 }
MC-PROTO: 54970 writing packet play.keep_alive
MC-PROTO: 54970 { keepAliveId: 1917937053 }

```

---

Now that the client is more modular, it's easier to see what's going on, need to test to confirm but I think this is where it is unexpectedly (or undesirably) executing from:

```javascript
// src/client/setProtocol.js
ar states = require("../states");

module.exports = function(client, options) {
  client.on('connect', onConnect);

  function onConnect() {
    client.write('set_protocol', {
      protocolVersion: options.protocolVersion,
      serverHost: options.host,
      serverPort: options.port,
      nextState: 2
    });
```

setProtocol sends the set_protocol packet as soon as the socket connects.

deathcap:
caseCorrecting and tcp_dns (SRV lookup and TCP initiation) can occur beforehand, but sending set_protocol should be delayed until the client knows (via ping) the protocolVersion.

deathcap:
Listen for another event in setProtocol before sending the packet? 'connect' is going to be received when the TCP socket it connected, whether we are ready for it or not. Additionally block on another event (optionally), something like 'connect_allowed'?

deathcap:
> Also I think it would be possible that way to use the same socket for
pinging and for the rest. The vanilla client probably doesn't use the same
socket, but I think it makes sense.

The server immediately closes the socket after sending the ping response, so unfortunately I don't think reusing it is possible.

deathcap:
Converted to use "sync" API ok, started to move into createClient() via a src/client/autoVersion.js plugin. Having some difficulty re-versioning the client object after it is instantiated. The constructor calls setHandshake() and sets .version to the major version, so I do this too in autoVersion.js after the version is ready, and then emit connect_allowed to uncork setProtocol, it sends the set_protocol packet but then the program just exits:

```
node-minecraft-protocol $ NODE_DEBUG=mc-proto node examples/client_auto/client_auto.js localhost 9999
MC-PROTO: 7249 creating client
MC-PROTO: 7249 pinging localhost
connected
MC-PROTO: 7249 writing packet handshaking.set_protocol
MC-PROTO: 7249 { protocolVersion: 47,
  serverHost: 'localhost',
  serverPort: 9999,
  nextState: 1 }
MC-PROTO: 7249 writing packet status.ping_start
MC-PROTO: 7249 {}
MC-PROTO: 7249 read packet status.server_info
MC-PROTO: 7249 { response: '{"description":{"text":"A Minecraft Server"},"players":{"max":20,"online":0},"version":{"name":"15w40b","protocol":76}}' }
MC-PROTO: 7249 writing packet status.ping
MC-PROTO: 7249 { time: [ 0, 0 ] }
MC-PROTO: 7249 read packet status.ping
MC-PROTO: 7249 { time: [ 0, 0 ] }
MC-PROTO: 7249 ping response { description: { text: 'A Minecraft Server' },
  players: { max: 20, online: 0 },
  version: { name: '15w40b', protocol: 76 },
  latency: 1 }
MC-PROTO: 7249 Server description: { text: 'A Minecraft Server' }
MC-PROTO: 7249 Server version: 15w40b, protocol: 76
{ minecraftVersion: '15w40b',
  version: 76,
  usesNetty: true,
  majorVersion: '1.9' }
MC-PROTO: 7249 writing packet handshaking.set_protocol
MC-PROTO: 7249 { protocolVersion: 76,
  serverHost: 'localhost',
  serverPort: 9999,
  nextState: 2 }
MC-PROTO: 7249 writing packet login.login_start
MC-PROTO: 7249 { username: 'echo' }
```

deathcap:
Looks like the ping connection is somehow interfering with the subsequent normal connection. `connected` is logged right after "pinging localhost", when receiving the client `connect` event, but not for the re-connect with the new version. The program ends because the client receives the `end` event.

deathcap:
Updating the client state using `client.state = states.HANDSHAKING` instead of `client.setSerializer(states.HANDSHAKING)` resolves the above issue, now able to connect with the auto-negotiated protocol.

deathcap:
Removing [WIP], ready to pull, but see some thoughts for future improvements to versioning in https://github.com/PrismarineJS/node-minecraft-protocol/issues/234

deathcap:
Failing tests in requiring minecraft-data: https://github.com/PrismarineJS/node-minecraft-data/issues/16

src/client/caseCorrect.js
```diff
@@ -1,7 +1,8 @@
 var yggdrasil = require('yggdrasil')({});
 var UUID = require('uuid-1345');
 
-module.exports = function(client, options) {
+module.exports = function(client) {
+  var options = client.options;
```
rom1504: Why client.options ? The `client, options` pattern is consistent with what mineflayer is doing, and it seems to makes sense. (It also allows destructuring : for example https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/plugins/world.js#L10 )

rom1504: Ah I saw your commit message "This allows client plugins to change the options on the client object,
and have it reflected in other plugins. Previously, had to rely on the
options being the same object, and hold a reference to it. Now it is
always accessible as `client.options`."

It sounds like it assume some order of execution between plugins. Which plugins is accessing the modified property ?

circle.yml
```diff
@@ -8,3 +8,4 @@ machine:
 dependencies:
   pre:
     - mkdir minecraft-server
+    - git clone https://github.com/PrismarineJS/minecraft-data node_modules/minecraft-data/minecraft-data || true
```
rom1504: this can be removed now (with minecraft-data 0.19.1)

src/client/caseCorrect.js
rom1504: ah that kind of thing `https://github.com/PrismarineJS/node-minecraft-protocol/pull/331/files#diff-9053b72ce1984293dcb954a0ccc89097R9`.
Couldn't it be put in a client.wait_connect ?
But the problem of relying on the order of execution of plugins remains.

rom1504:
I made some comments
* minecraft-data 0.19.1 is now usable
* I'd like to keep the `client, options` pattern if possible, I think it would be better to put the changing attributes in the Client object https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client.js#L14

(I guess it's probably necessary for some plugins to depend on other, but let's probably keep it implicit for now)

roblabla:
I am very, very, **very** much against the `client.options` pattern. Ideally, I should be able to do

```
var client = new Client();
tcpConnect(client, { opts for tcpConnect });
forgeHandshake(client, { opts for forgeHandshake });
play(client, { opts for play });
etc...
```

What plugin needs modifying the options ? I believe that if any plugin requires this, it should instead wrap the original plugin in some way.

deathcap:
Updated to data 0.19.1, removing need for git clone post-action

Plugins currently modifying options:

* autoVersion: sets options.wait_connect, used by setProtocol plugin
* caseCorrect: sets accessToken and haveCredentials
* tcp_dns: sets port and host to defaults if not set, and sets connect to callback
* forgeHandshake: sets tagHost

I suppose these 'options' could be moved onto the client object, since they're not really options but data passed between plugins (which can initially come from options, but only as a default). tagHost could use the system described in https://github.com/PrismarineJS/node-minecraft-protocol/issues/336#issuecomment-177942918

src/client/autoVersion.js
```diff
@@ -0,0 +1,67 @@
...
+
+var ping = require('../ping');
+var debug = require('../debug');
+var states = require('../states');
+var assert = require('assert');
+var minecraft_data = require('minecraft-data');
+
+// Get the minecraft-data version string for a protocol version
+// TODO: add to node-minecraft-data index (protocol to newest release, if multiple)
+function protocolVersion2MinecraftVersion(n) {
```
rom1504: this can now be implemented using https://github.com/PrismarineJS/node-minecraft-data/blob/master/example.js#L19 I think

deathcap: ah, missed that, will use postNettyVersionsByProtocolVersion

deathcap:
Alright I've restored the `options` parameters to each of the client plugin parameters. This gets us closer to separate independent options per plugin, but a few plugins still modify their options making them interdependent.

deathcap:
Moved `wait_connect` from options to the client object, but moving all the other plugin-mutable options to client is a much bigger change, would prefer it as a separate PR (poses some non-obvious questions, e.g. how to handle the version parameter in the options vs the client object)

rom1504:
Ok good enough. We can track the remaining things to do here in https://github.com/PrismarineJS/node-minecraft-protocol/issues/338 and https://github.com/PrismarineJS/node-minecraft-protocol/issues/234

PR #333, opened 2016-01-26
Modularization attempt of createClient()

deathcap:
An attempt at beginning to modularize createClient() into separate modular components, see also https://github.com/PrismarineJS/node-minecraft-protocol/pull/210 [WIP] Modularize NMP which covered server, but this PR is for client only. I also did not tackle how to enable/disable/load 'modules' in this PR, considering it out of scope — this is only a source-level refactor/cleanup to make the client more modular.

Note: lots of code moved around, but few actual changes; best reviewed by individual commit, I strived to keep each new module to one commit (but did not meet this goal, had to do some cleanup afterwards, did not bother to rebase/squash, though I could if needed). 

The basic design is each submodule is called with `(client, options)`, and it is responsible for modifying the client to add some functionality, given the optional options object. The goal being to reduce tight coupling, as much as possible, but it could still be improved further. Or the modules could be broken up differently. Currently includes:

* src/client/compress.js: sets compression threshold
 * would like the src/transforms/compression.js transformation to be piped in here; has a hard dependency with other internals not touched here, but ought to be more isolated so can simply remove this module to disable compression entirely
* src/client/caseCorrect.js: gets the case-correct username from yggdrasil before connecting
 * is also responsible for performing the next step - calling connect() - possible to factor this dependency out? if no credentials given, it doesn't case-correct, but still has to call connect() to trigger the next step in the connection process
* src/client/encrypt.js: ursa, yggdrasil, crypto, encryption key request. not extensively tested
* src/client/keepalive.js: listen for `keep_alive`, respond
* src/client/play.js: listen for `success`, transition to `PLAY` state
 * maybe could be renamed, or merged into another module. encrypt? handshake/login?
* src/client/setProtocol.js: sends `set_protocol` packet, this is very interesting from a modularization perspective, since FML|HS would want to change it, as would protocol negotiation add-ons
* src/client/tcp_dns.js: includes creating the TCP/IP socket with `net.connect()` and DNS resolution with `dns.resolveSrv()`
 * this is the biggest change, couldn't come up with a good name for the module, but it performs essential TCP/IP socket networking functions — this is very desirable to factor out, since, for example, a web socket server would not use it, and instead call setSocket() itself; the crucial abstraction here is a nodejs [stream](https://github.com/deathcap/wsmc/blob/master/minecraft-protocol-stream.js)

createClient() still includes the minecraft-data version lookup, but with these changes, the file is not much more than a list of modules to inject:

```javascript
var encrypt = require('./client/encrypt');
var keepalive = require('./client/keepalive');
var compress = require('./client/compress');
var caseCorrect = require('./client/caseCorrect');
var setProtocol = require('./client/setProtocol');
var play = require('./client/play');
var tcp_dns = require('./client/tcp_dns');

module.exports=createClient;

function createClient(options) {
…[version lookup code elided…]

  tcp_dns(client, options);
  setProtocol(client, options);
  keepalive(client, options);
  encrypt(client, options);
  play(client, options);
  compress(client, options);
  caseCorrect(client, options);

  return client;
}
```

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -215,6 +215,13 @@ class Client extends EventEmitter
     else
       this.compressor.write(buffer);
   }
+
+  // TCP/IP-specific (not generic Stream) method for backwards-compatibility
+  connect(port, host) {
+    var options = {port, host};
+    require('./client/tcp_dns')(this, options);
+    options.connect(this);
+  }
 }
 
 module.exports = Client;
--- a/src/client/caseCorrect.js
+++ b/src/client/caseCorrect.js
@@ -16,7 +16,7 @@ module.exports = function(client, options) {
         client.username = session.selectedProfile.name;
         accessToken = session.accessToken;
         client.emit('session');
-        client.connect(options.port, options.host);
+        options.connect(client);
       }
     };
 
@@ -38,6 +38,6 @@ module.exports = function(client, options) {
   } else {
     // assume the server is in offline mode and just go for it.
     client.username = options.username;
-    client.connect(options.port, options.host);
+    options.connect(client);
   }
 };
--- a/src/client/caseCorrect.js
+++ b/src/client/caseCorrect.js
@@ -3,10 +3,10 @@ var UUID = require('uuid-1345');
 
 module.exports = function(client, options) {
   var clientToken = options.clientToken || UUID.v4().toString();
-  var accessToken;
-  var haveCredentials = options.password != null || (clientToken != null && options.session != null);
+  options.accessToken = null;
+  options.haveCredentials = options.password != null || (clientToken != null && options.session != null);
 
-  if(haveCredentials) {
+  if(options.haveCredentials) {
     // make a request to get the case-correct username before connecting.
     var cb = function(err, session) {
       if(err) {
@@ -14,7 +14,7 @@ module.exports = function(client, options) {
       } else {
         client.session = session;
         client.username = session.selectedProfile.name;
-        accessToken = session.accessToken;
+        options.accessToken = session.accessToken;
         client.emit('session');
         options.connect(client);
       }
--- a/src/client/caseCorrect.js
+++ b/src/client/caseCorrect.js
@@ -1,0 +1,43 @@
+var yggdrasil = require('yggdrasil')({});
+var UUID = require('uuid-1345');
+
+module.exports = function(client, options) {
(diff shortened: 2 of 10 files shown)


rom1504:
Nice !
Now to figure out how will forge fit with that pattern.

roblabla:
```
var client = createClient();
forgeSupport(client, options);
```

It should be possible that way, no ?

rom1504:
Something like that would be nice yes, but the "get the mods and version
from ping before doing anything" situation need to be solved before that
can work. Maybe it can be a createClient option (to wait the ping pong) and
then there'd be a client.pongAnswer that the forge plugin can use.

roblabla:
Hmm. I'm thinking there should be a 'next' parameter passed to the modules so they can control when the next Middleware is called. Then we can easily implement the forge ping stuff and delay starting the connection until the ping got a response.

rom1504:
Yeah that sounds good

deathcap:
A next parameter could be interesting, it may also solve the issue with src/client/caseCorrect.js, where it is responsible for initiating the client connection. Would need some more design/prototyping how this would work exactly, though. 

I think the dynamic Forge ping to get mods issue would be best solved separately from this PR, e.g. using the same technique in https://github.com/PrismarineJS/node-minecraft-protocol/pull/331 dynamic cross-protocol support. For now, (or until https://github.com/PrismarineJS/node-minecraft-protocol/pull/326 Forge client support, gets updated to use these submodules), Forge clients could specify the exact `forgeMods` array they want to offer to the server. Then adding support would be a simple matter as @roblabla mentions, installing the middleware something like:

    forgeHandshake(client, options);

maybe some of the other submodules should be consolidated into 'handshake' (implied vanilla handshake) for consistency. But the `options` here would include `forgeMods`. Most of https://github.com/PrismarineJS/node-minecraft-protocol/pull/326 can be easily factored out, since the bulk of the code is a custom_payload listener, the only wrinkle is the need to tag the hostname, so a change to src/client/setProtocol.js (to add a custom tag option to serverHost?) may be needed. But lets put this off until https://github.com/PrismarineJS/node-minecraft-protocol/pull/326, for this PR I tried to add no new functionality, but pave the way for future enhancements made possible by the improved modularization.

rom1504:
Okay let's merge :+1:

Issue #332, opened 2016-01-26
FE01 ping support


deathcap:
src/ping.js implements a server list ping using `ping` and `ping_start` in the `STATUS` protocol state, but there is another type of ping initiated by sending the bytes 0xfe 0x01. Sometimes known as the "legacy" ping, but it supported by modern versions of Minecraft, including 1.8.9 and the 1.9 snapshots: http://wiki.vg/Protocol#Legacy_Server_List_Ping

> While not technically part of the current protocol, legacy clients may send this packet to initiate Server List Ping, and modern servers should handle it correctly.

node-minecraft-protocol should support FE01 ping, in both the server and client. Server is especially important since third-party ping/status software may still send FE01 pings, due to simplicity and widespread support (vanilla servers support it). Client support would be useful when pinging servers of an unknown version, including those with the modern Netty protocol, or earlier versions (all the way back to Minecraft 1.4.4 release is supported by the FE01 ping).

Example of fe01 ping in nodejs: https://github.com/deathcap/mcping16/blob/master/mcping16.js#L124 - but needs to be cleaned up to use protodef, etc.

deathcap:
This packet is supported for serialization/deserialization by https://github.com/PrismarineJS/minecraft-data/blob/master/data/1.8/protocol.json#L148:

```javascript
        "legacy_server_list_ping": {
          "id": "0xfe",
          "fields": [
            {
              "name": "payload",
              "type": "ubyte"
            }
          ]
        }
```

but nothing currently reads/writes it

roblabla:
That probably doesn't work. It needs to be sent without the length-prefix and with a byte ID instead of varint ID, no ?

deathcap:
Yeah I think needs to be special-cased somehow

rom1504:
Yeah maybe just with a simple if (packet.name=="legacy_server_list_ping"){
/* write directly to this.socket */ } in Client.write

rom1504:
Ah, not sure how that should be handled by the parsing pipeline though

deathcap:
Researching the various pings in https://github.com/deathcap/node-minecraft-ping, details in the repo but overall summary:

| Minecraft Version | ping_fe01fa   | ping_fe01     | ping_fe       | Netty status ping(*)
| -------------     | ------------- | ------------- | ------------- | -------------
| 1.9               | YES           | YES           | Limited       | YES
| 1.8.9             | YES           | YES           | Limited       | YES
[...]

PR #334, opened 2016-01-31
Server response to legacy ping (fe and fe01) packets

deathcap:
Implements server-side support for https://github.com/PrismarineJS/node-minecraft-protocol/issues/332 FE legacy ping support

Supports type 0 (fe) and type 1 (fe01) pings, testing with https://github.com/deathcap/node-minecraft-ping:

```
node-minecraft-protocol $ NODE_DEBUG=mc-proto node examples/server/server.js
node-minecraft-ping $   node demo.js localhost 25565
received ping_fe01fa { pingVersion: 1,
  protocolVersion: 47,
  gameVersion: '1.8.8',
  motd: 'Vox Industries',
  playersOnline: 0,
  maxPlayers: 12 }
received ping_fe { pingVersion: 0,
  motd: 'Vox Industries',
  playersOnline: 0,
  maxPlayers: 12 }
received ping_fe01 { pingVersion: 1,
  protocolVersion: 47,
  gameVersion: '1.8.8',
  motd: 'Vox Industries',
  playersOnline: 0,
  maxPlayers: 12 }
```

Some notes about issues/limitations with this pull request:

* The deserializer cannot parse legacy ping packets, because of the missing varint packet type header, so I worked around by having the splitter prepend a varint packet id. Ideally minecraft-data could specify how to parse these packets: https://github.com/PrismarineJS/minecraft-data/issues/95 but idk how
* minecraft-data specifies server_legacy_list_ping as requiring a ubyte payload, but for ping type 0 it is not, so I fill in a \0 in the ping payload to avoid deserialization errors. Also would need changes to https://github.com/PrismarineJS/minecraft-data/issues/95
* Serializer is not used to write the ping responses (or the ping requests, which aren't supported in this PR) for the same reason; client.writeRaw() is also unsuitable because it prepends length, so I use client.socket.write(). Maybe this could be refactored further, a new "legacy" packet state, which doesn't have length varints? Could be revisited later if/once pre-Netty protocol support is added.
* ~~Node.js has utf16le support in Buffer but not utf16be (https://github.com/nodejs/node-v0.x-archive/issues/1684); encoded hackishly~~ updated to use native utf16le + swap with https://github.com/substack/endian-toggle, works well
* No client support for sending legacy ping requests or parsing legacy ping responses (can use n-mc-ping, but would be nice to have in nmp)

but it is functional

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -14,6 +14,7 @@ class Client extends EventEmitter
     super();
     this.version=version;
     this.isServer = !!isServer;
+    this.splitter=framing.createSplitter();
     this.setSerializer(states.HANDSHAKING);
     this.packetsToParse={};
     this.serializer;
@@ -21,7 +22,6 @@ class Client extends EventEmitter
     this.framer=framing.createFramer();
     this.cipher=null;
     this.decipher=null;
-    this.splitter=framing.createSplitter();
     this.decompressor=null;
     this.deserializer;
     this.isServer;
@@ -51,6 +51,7 @@ class Client extends EventEmitter
     this.deserializer = createDeserializer({ isServer:this.isServer, version:this.version, state: state, packetsToParse:
       this.packetsToParse});
 
+    this.splitter.recognizeLegacyPing = state === states.HANDSHAKING;
 
     this.serializer.on('error', (e) => {
       var parts=e.field.split(".");
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   "browser": "browser.js",
   "devDependencies": {
     "babel-preset-es2015": "^6.3.13",
+    "endian-toggle": "0.0.0",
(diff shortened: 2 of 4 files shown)


src/client.js
```diff
@@ -51,6 +51,7 @@ class Client extends EventEmitter
     this.deserializer = createDeserializer({ isServer:this.isServer, version:this.version, state: state, packetsToParse:
       this.packetsToParse});
 
+    this.splitter.recognizeLegacyPing = state === states.HANDSHAKING;
```
rom1504: isn't `state === states.HANDSHAKING` always true at that point ?

rom1504:
This is kind of a hack, but well, so is that packet. It would be ideal to handle this kind of thing properly directly in protodef, but let's do that later in #332

PR #337, opened 2016-02-01
much let, very const

### LINKED ISSUE #335: Use ES6 let/const everywhere instead of var
With node-minecraft-protocol embracing ES6 (https://github.com/PrismarineJS/node-minecraft-protocol/pull/322 et. al.), I think it is worth considering updating the variable declarations to use the new block-scoped `let` and `const` keywords, instead of the function-scoped `var`. Current status:

```
node-minecraft-protocol $ grep -r 'var ' src/|wc -l
     175
node-minecraft-protocol $ grep -r 'let ' src/|wc -l
       3
node-minecraft-protocol $ grep -r 'const ' src/|wc -l
       0
```

Many of the `var`s could be updated to `const`s (if assigned only once), and the others to `let`s. `let` avoids some of the pitfalls of `var`'s scoping rules.

New code can use let/const, but it feels odd to have a mix of let/const/var, so a comprehensive conversion could make sense. Should be a fairly easy change to make, but results in a large diff and will cause merge conflicts so I've put it off for now.

rom1504:
fix #335

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -1,6 +1,6 @@
-var Client = require('./client');
-var Server = require('./server');
-var serializer = require("./transforms/serializer");
+const Client = require('./client');
+const Server = require('./server');
+const serializer = require("./transforms/serializer");
 
 module.exports = {
   Client: Client,
--- a/src/client.js
+++ b/src/client.js
@@ -1,12 +1,12 @@
-var EventEmitter = require('events').EventEmitter;
-var debug = require('./debug');
-var compression = require('./transforms/compression');
-var framing = require('./transforms/framing');
-var crypto = require('crypto');
-var states = require("./states");
+const EventEmitter = require('events').EventEmitter;
+const debug = require('./debug');
+const compression = require('./transforms/compression');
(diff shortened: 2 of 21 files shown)


rom1504:
WIP.
Apparently `constint` is not a thing :D

rom1504:
I'll probably wait for the other PRs to be merged before merging this, to avoid creating tons of merge conflicts.

rom1504:
rebased after #331

rom1504:
Ok let's merge this. #326 shouldn't conflict too much with it.

deathcap:
@rom1504 Looks good but what about `var` in the test scripts?

rom1504:
@deathcap test scripts (and examples) are currently es5.
I think examples probably need to stay es5 since we support node4.

Tests could be compiled like I did in https://github.com/PrismarineJS/prismarine-provider-anvil/blob/master/gulpfile.js#L29 (or use mocha's babel register option https://babeljs.io/docs/setup/#mocha). Do you think it's worth it ?

Issue #343, opened 2016-02-10
Exception in client chat example


nsporillo:
Exception: https://gist.github.com/nsporillo/429bddc88896cf147d24

Server version: 1.7.10

Not sure if the client chat example has support for 1.7.10, but I'm getting this exception when I try to connect to a server.

deathcap:
@nsporillo How are you able to get the client to connect to a 1.7.10 server, which version of node-minecraft-protocol are you using?

With the latest node-minecraft-protocol on git (which supports 1.8.9 and a 1.9 snapshot), connecting to a 1.7.10 vanilla I see this error instead:

```
node-minecraft-protocol $ node examples/client_chat/client_chat.js localhost 17100 testuser
connecting to localhost:17100
user: testuser
Successfully connected to localhost:17100
disconnected: "Outdated server! I\u0027m still on 1.7.10"
Connection lost
```

nsporillo:
Oh I didn't realize latest version doesn't support older mc. 

I believe the server I connected to accepts 1.8 clients due to protocol hacks.

I wrote a CLI chat client back in the day (https://gist.github.com/nsporillo/325cf29b3fe9e4b80fd8)
and wanted to use it again but don't have my fork of n-m-p from then.

Any ideas?

deathcap:
Ah, presumably a 1.7.10/1.8 protocol hack from Spigot? There have been other reports of compatibility problems with this server software and PrismarineJS: https://github.com/PrismarineJS/mineflayer/issues/369 (mineflayer uses node-minecraft-protocol).

nsporillo:
Yes. I was just reading that, however it should be irrelevant. I just need a copy of node-minecraft-protocol from the 1.7 era and my mine chat bot will work with some modifications. I'll close this issue, but you guys should 100% work on adding some sort of compatibility especially if it's only a few extra things.

deathcap:
@nsporillo can you try this patch:

```javascript
--- a/examples/client_chat/client_chat.js
+++ b/examples/client_chat/client_chat.js
@@ -171,8 +171,9 @@ function parseChat(chatObj, parentState) {
 
       chat += color(util.format.apply(this, args), getColorize(parentState));
     }
-    chatObj.extra.forEach(function(i) {
-      chat += parseChat(chatObj.extra[i], parentState);
+    if (chatObj.extra)
+    chatObj.extra.forEach(function(item) {
+      chat += parseChat(item, parentState);
     });
[...]

PR #344, opened 2016-02-10
Fix client_chat example extra array iteration. Closes GH-343

deathcap:
examples/client_chat tries to iterate the `extra` array (see reference examples at http://wiki.vg/Chat) using forEach, but the function is called with the array items not the indices (`for ... in` will get the indices, and ES6 `for ... of`will get the items like forEach), so the subscript is undefined causing the error reported in https://github.com/PrismarineJS/node-minecraft-protocol/issues/343 https://gist.github.com/nsporillo/429bddc88896cf147d24. Regressed not long ago as part of some code cleanup https://github.com/PrismarineJS/node-minecraft-protocol/commit/b2b53c93433a599e02b7c23b52f91df988c1a86d

This PR fixes this error by using the array items from forEach, and also adding a check for `chatObj.extra` since it is optional. Resolves the issue according to the discussion in https://github.com/PrismarineJS/node-minecraft-protocol/issues/343

### DIFF
--- a/examples/client_chat/client_chat.js
+++ b/examples/client_chat/client_chat.js
@@ -171,9 +171,11 @@ function parseChat(chatObj, parentState) {
 
       chat += color(util.format.apply(this, args), getColorize(parentState));
     }
-    chatObj.extra.forEach(function(i) {
-      chat += parseChat(chatObj.extra[i], parentState);
-    });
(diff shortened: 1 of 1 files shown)


rom1504:
indeed, I missed that !

PR #346, opened 2016-02-15
Add compressedNbt data type, used in MC 1.7

deathcap:
Testing this change with https://github.com/PrismarineJS/minecraft-data/pull/104


deathcap:
Tested with the related minecraft-data PR on 1.7.10, works well, example debug output:

```
MC-PROTO: 8592 read packet play.entity_equipment
MC-PROTO: 8592 { entityId: 25246,
  slot: 0,
  item: 
   { blockId: 261,
     itemCount: 1,
     itemDamage: 0,
     nbtData: { type: 'compound', name: '', value: [Object] } } }
```

(using client_echo with added option `version: 1.7.10`). Removing [WIP]

rom1504:
nice

rom1504:
(let's not add native datatypes for everything but this is specific enough to be hard to implement within protodef abstraction, so better do it that way (until protodef can handle that kind of thing someday maybe))

PR #351, opened 2016-02-17
[WIP] 1.7 protocol support in node-minecraft-protocol

deathcap:
Changes to node-minecraft-protocol to enable 1.7 protocol support (testing with 1.7.10)

Requires:

* https://github.com/PrismarineJS/minecraft-data/pull/104
* https://github.com/PrismarineJS/node-minecraft-data/pull/19

Currently hardcoding the version where whole-protocol compression was added, but probably a better solution should be devised (maybe adding "uses/doesn't use whole protocol compression" to somewhere in minecraft-data per protocol version, and having node-minecraft-protocol honor it).

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -253,8 +253,10 @@ function createServer(options) {
       if(onlineMode == false || isException) {
         client.uuid = nameToMcOfflineUUID(client.username);
       }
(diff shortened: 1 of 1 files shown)


rom1504:
good enough for now

PR #357, opened 2016-02-18
update mcdata : 16w05b and 15w40b kept

rom1504:


rom1504:
it's getting stuck for some reason. I'll try to figure that out

rom1504:
okay it's failing just for 15w40b now. I bet it's something about majorVersion/minecraftVersion, it must be using the wrong protocol at some point.

rom1504:
hmm or maybe not, 0x79 is not an id anywhere

rom1504:
yeah the problem was there was majorVersion in half the code

PR #369, opened 2016-02-20
add an optional property to add custom packets to createClient and createServer

rom1504:
@deathcap I think this should be enough for your forge use cases.

Should customPackets be indexed by version too, so packets only get added for a given version ?

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -18,8 +18,9 @@ automatically logged in and validated against mojang's auth.
  * maxPlayers : default to 20
  * keepAlive : send keep alive packets : default to true
  * version : 1.8 or 1.9 : default to 1.8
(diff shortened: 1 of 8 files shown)


rom1504:
So it seems greenkeeper got @roblabla test account banned, great.

rom1504:
@deathcap this would also make it possible to override packets for https://github.com/deathcap/minecraft-data/blob/mc17varshort/data/1.7/protocol.json#L1908

rom1504:
That works.

roblabla:
@rom1504 my account wasn't banned. Mojang sent me an email saying they got "suspicious activity" and reset my password :laughing: 

I can put back the same password if you want.

Issue #367, opened 2016-02-19
Handling unknown packets, and/or registering custom packets


deathcap:
When connecting to Forge servers, some mods send their own custom packets (instead of using the custom_payload plugin channel packet meant for this, for some reason), causing node-minecraft-protocol to throw an error since it is unrecognized:

```
Error: Read error for name : -26 is not in the mappings value
    at ProtoDef.readMapper (/Users/admin/games/voxeljs/ProtoDef/dist/datatypes/utils.js:38:41)
    at ProtoDef.read (/Users/admin/games/voxeljs/ProtoDef/dist/protodef.js:103:31)
    at /Users/admin/games/voxeljs/ProtoDef/dist/datatypes/structures.js:114:32
    at tryCatch (/Users/admin/games/voxeljs/ProtoDef/dist/utils.js:31:12)
    at tryDoc (/Users/admin/games/voxeljs/ProtoDef/dist/utils.js:38:10)
    at /Users/admin/games/voxeljs/ProtoDef/dist/datatypes/structures.js:113:5
    at Array.forEach (native)
    at ProtoDef.readContainer (/Users/admin/games/voxeljs/ProtoDef/dist/datatypes/structures.js:108:12)
    at ProtoDef.read (/Users/admin/games/voxeljs/ProtoDef/dist/protodef.js:46:25)
    at ProtoDef.read (/Users/admin/games/voxeljs/ProtoDef/dist/protodef.js:103:31)
```

any ideas how to best handle this. The list of custom packets cannot be known beforehand (at least for a simple bot, more sophisticated clients could in theory know what packets to accept from modded servers), and although there is an option in node-minecraft-protocol to specify what packets to parse, it should still be possible to pass these unknown packets through proxies and to clients/servers as opaque blobs since in 1.7+ all the packets are length-prefixed.

Maybe an unknown_packet event? Or unknown_packet_+(id)?

rom1504:
Hmm, I guess we could customize the error yeah. But it would still be an error I think and then people could listen on the error and parse the packet themselves or something ?

(by customize I mean something like https://github.com/roblabla/ProtoDef/blob/master/src/utils.js#L39)

(the problem of not considering this to be an error is that most often it means we have some error in protocol.json and it is an actual error)

rom1504:
if you listen to the error event, it doesn't manage to ignore the packet ?

deathcap:
[...]

PR #371, opened 2016-02-20
don't stop after an error

rom1504:
https://github.com/nodejs/node/issues/3045#issuecomment-142955116 #367

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "buffer-equal": "1.0.0",
     "minecraft-data": "^1.0.3",
     "prismarine-nbt": "0.2.1",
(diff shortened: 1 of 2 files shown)


rom1504:
That certainly allow the "ignore" the unknown packets.
Not sure if it's a good thing to continue even though there's an error though.

rom1504:
It's still possible to not listen on "error" event to make things crash though.

rom1504:
To test this change 7A to 79 in https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/client_custom_packets/client_custom_packets.js#L49

It works fine. We'll see if that poses any problem or if we can do better later

PR #390, opened 2016-03-21
use debug package, fix #386

### LINKED ISSUE #386: use debug package instead of debug.js
https://www.npmjs.com/package/debug

rom1504:


rom1504:
untested. Make sure it works ok before merging.

PR #394, opened 2016-04-03
Upgrade minecraft-data to v1.4.1

holic:
So we can use Minecraft version 1.9.2


rom1504:
it's ^1.2.6 so it should not be needed as long as you npm update

holic:
Ah, great. Misinterpreted the `^` in semver.

PR #398, opened 2016-04-04
Fixed options.favicon in createServer

alrek-consulting:
- if the server is created with option "favicon" (base64 encoded image string), it now loads it into the server object inside "createServer", and passes it on to the "response" object used in the "onPing" function. If it is empty, it is set to 0. Tested working.


alrek-consulting:
Could anyone help point me in the right direction? Is the tests broken, or am I doing something wrong?

rom1504:
What you are doing is correct but there's a little thing to fix in the test so it passes.

rom1504:
Thanks, I'll fix it after merging.

PR #399, opened 2016-04-10
use esnext to get more es6, mostly require/module.exports -> import/export

rom1504:


rom1504:
For reference:
1. `npm install -g esnext`
2. `for i in `find src -name "*.js"` ; do esnext -I "$i"; done`
3.  fix stuff it broke

roblabla:
I'm not a big fan of the es6 import/export syntax personally. The fact that es6 has different "types" of export (export default and the rest) pisses me off greatly.

IMO, the require function/module.exports variable are much clearer. Besides they're unlikely to be unsupported in the future.

Just my 2c.

rom1504:
Yeah I agree import/export don't improve the code much. This was more an experiment that something I really care to change.

I'll try running telling esnext to not change the require, see if some improvements can be done anyway.

rom1504:
The good point about import/export is it makes static analysis a bit easier. But yeah not sure if it's worth it.

rom1504:
Looking at babel code ATM and they have really clean code. They use import/export which actually look pretty good. They also use flow which is pretty useful. I'm going to try and see if any of this can apply here.

rom1504:
Let's not merge this.

PR #402, opened 2016-04-27
Enable strict mode

deathcap:
Adds "use strict" to all sources. This adds additional strictness which can help find bugs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode?redirectlocale=en-US&redirectslug=JavaScript%2FReference%2FFunctions_and_function_scope%2FStrict_mode

Split out from https://github.com/PrismarineJS/node-minecraft-protocol/pull/319 Use native node.js ES6 support

### DIFF
--- a/examples/client_auto/client_auto.js
+++ b/examples/client_auto/client_auto.js
@@ -1,3 +1,5 @@
+'use strict';
+
 var mc = require('minecraft-protocol');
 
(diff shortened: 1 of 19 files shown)


roblabla:
I'm very much :+1: on this. Strict mode would help catch a lot of subtle bugs.

rom1504:
:+1: 

Let's do this.

PR #423, opened 2016-08-10
plugin channel support for server and client

plexigras:
still needs to be adde to the documantation


rom1504:
How can I try the PR ? do you have an example of vanilla plugin channel that is easy to try ? (might be useful to put that in an example btw)

rom1504:
added things in https://github.com/rom1504/node-minecraft-protocol/commit/a0717a1232f83799bd87fb7e918392cb175183d4

Still missing before merging : the doc.

PR #434, opened 2016-10-17
Fix the REGISTER channel type

### LINKED ISSUE #432: Connection through BungeeCord/Waterfall forcibly closed
Can we update the minecraft-data dependency to 2.7.0 to support the newer minecraft versions?

roblabla:
Fixes #432

### DIFF
--- a/src/client/pluginChannels.js
+++ b/src/client/pluginChannels.js
@@ -7,12 +7,14 @@ module.exports = function(client, options) {
   var proto = new ProtoDef();
   proto.addTypes(mcdata.protocol.types);
   proto.addTypes(minecraft);
+  proto.addTypes({'registerarr': [readDumbArr, writeDumbArr, sizeOfDumbArr]});
+
   client.registerChannel = registerChannel;
   client.unregisterChannel = unregisterChannel;
   client.writeChannel = writeChannel;
 
-  client.registerChannel("REGISTER",["string",[]]);
-  client.registerChannel("UNREGISTER",["string",[]]);
+  client.registerChannel("REGISTER",["registerarr",[]]);
+  client.registerChannel("UNREGISTER",["registerarr",[]]);
 
 
   function registerChannel(name, parser, custom) {
@@ -56,6 +58,36 @@ module.exports = function(client, options) {
         channel:channel,
         data:proto.createPacketBuffer(channel,params)
       });
+  }
+
+  function readDumbArr(buf, offset) {
+    var ret = {
+      value: [],
+      size: 0
+    };
+    let results;
+    while (offset < buf.length) {
+      if (buf.indexOf(0x0, offset) == -1)
+        results = this.read(buf, offset, "restBuffer", {});
+      else
+        results = this.read(buf, offset, "cstring", {});
+      ret.size += results.size;
+      ret.value.push(results.value.toString());
+      offset += results.size;
+    }
+    return ret;
+  }
+
+  function writeDumbArr(value, buf, offset) {
+    // TODO: Remove trailing \0
+    value.forEach(function(v) {
+      offset += this.write(v, buf, offset, "cstring", {});
+    });
+    return offset;
+  }
+
+  function sizeOfDumbArr(value) {
+    return value.reduce((acc, v) => acc + this.sizeOf(v, "cstring", {}), 0);
   }
 };
 


rom1504:
omg really

rockmacaca:
...not quite there!

First of all, line 70 in pluginChannels.js, `buffer` is not defined, should be `buf`

```
      results = this.read(buffer, offset, "cstring", {});
```

I fixed that, but it's still dropping out after after logging `REGISTER`, I'll see if I can pull up the error

rockmacaca:
Here's the error now:

```
{ Error: Read error for undefined : undefined
    at PartialReadError.ExtendableError (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\node_modules\protodef\dist\utils.js:105:11)
    at new PartialReadError (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\node_modules\protodef\dist\utils.js:118:91)
    at ProtoDef.readCString (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\node_modules\protodef\dist\datatypes\utils.js:275:49)
    at ProtoDef.read (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\node_modules\protodef\dist\protodef.js:102:31)
    at ProtoDef.readDumbArr (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\dist\client\pluginChannels.js:73:22)
    at ProtoDef.read (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\node_modules\protodef\dist\protodef.js:46:25)
    at ProtoDef.read (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\node_modules\protodef\dist\protodef.js:102:31)
    at e.message (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\node_modules\protodef\dist\protodef.js:160:23)
    at tryCatch (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\node_modules\protodef\dist\utils.js:83:12)
    at ProtoDef.parsePacketBuffer (C:\Users\kherock\Desktop\mc\node_modules\minecraft-protocol\node_modules\protodef\dist\protodef.js:159:23)
  name: 'Error',
  message: 'Read error for undefined : undefined',
  partialReadError: true }
```

roblabla:
Trying it out myself, I think I figured it out, and I hate mojang.

```
> new Buffer("46 4d 4c 00 46 4d 4c 7c 48 53 00 46 4f 52 47 45 00 42 75 6e 67 65 65 43 6f 72 64".split(" ").join(""), "hex").toString()
'FML\u0000FML|HS\u0000FORGE\u0000BungeeCord'
```

Notice how there's no \0 at the end ? >_>

roblabla:
@rockmacaca and here's a fix (I should note, a very bad and under-performing one). Can you try this version ?

rockmacaca:
That's an affirmative

rockmacaca:
I'm looking at improving the implementation, but first trying to understand it - Why is the length of a CString (# of characters) + 1? This makes the sizeOf for registerarr one too long. Also, any reason we're avoiding string helpers? Are they just inefficient?

roblabla:
@rockmacaca A CString is a string with a \0 at the end. It's called a CString because in C, strings are encoded as a series of characters with a trailing 0 to mark its end.

A simple way to make this much better is to avoid using CString/restBuffers entirely and instead use the indexOf return to slice the buffer directly.

@rom1504 I just had a thought. Why don't we have a generic "length-prefixed" and "framed" type in ProtoDef, taking a type as a parameter, and passing the buffer as a parameter. For instance : 

```
{
    "string": ["prefixed", { "type": "rawstring" , countType: "varint"}],
    "cstring": ["framed", { "type": "rawstring", "endSequence": "\0"}]
}
```

rawstring being a simple buf.toString(). The idea is that the reader should just consume the entire buffer they're provided. I think it'd be interesting to toy with the idea of providing readers with just the data they're supposed to read, not less/more. This would make a bunch of datatypes much easier to write (I'm thinking about the map_chunk_bulk, among others).

rockmacaca:
@roblabla 

Here's my implementation suggestion, using the `pstring` type, if it means anything:

```
function readDumbArr(buf, offset) {
  var results = this.read(buf, offset, ["pstring", { count: buf.length }], {});
  results.value = results.value.split('\0');
  return results;
}

function writeDumbArr(value, buf, offset) {
  var str = value.join('\0');
  offset += this.write(v, buf, offset, ["pstring", { count: str.length }], {});
  return offset;
}

function sizeOfDumbArr(value) {
  return this.value.join('\0').length;
}
```

rom1504:
@roblabla sounds good. I guess prefixed fit for buffer, array and pstring.

In this case, the buffer is already delimited by the size of the packet (http://wiki.vg/Protocol#Plugin_Message_.28clientbound.29) : it's a rest buffer. And indeed some kind of split would probably make sense here, even though it's highly specific.

Issue #373, opened 2016-02-20
Consider allow an option in the test to run them in parallel


rom1504:
https://circleci.com/docs/parallel-manual-setup

I think we can use CIRCLE_NODE_INDEX and CIRCLE_NODE_TOTAL to give `total_version_count/CIRCLE_NODE_TOTAL` versions to each VM

That can give us a 4x test time decrease. (currently takes 7min, should take around 2min if parallel)

rom1504:
To make it possible to run tests in parallel on a single machine 2 things are needed:
* [x] using a random port
* [ ] using a random (sub)dir for the vanilla server data

rom1504:
Would def be useful, the tests are so damn slow.

rom1504:
done

PR #449, opened 2016-11-16
parallel circle ci testing of clientTest.js (#373)

rom1504:


rom1504:
https://circleci.com/gh/PrismarineJS/node-minecraft-protocol/692 something seems wrong in the circle ci file. I don't understand what

rom1504:
works, build twice as fast (4 nodes but 5 versions means at least one node need to run 2 versions)

PR #463, opened 2017-01-28
Add errorHandler option

roblabla:


roblabla:
Derp, why did I never merge this

PR #490, opened 2017-05-10
try adding 17w18b to tests

rom1504:
https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/17w18b/protocol.json#L3659 breaks the packet test
probably a pb with anon + switch not being handled by the value generation

fix the test


rom1504:
done

PR #500, opened 2017-07-09
trying to fix autoversion in online mode

rom1504:


rom1504:
Ah I get it.

We should guess the version before auth, not after. Otherwise changing protocol in the middle causes weird stuff.

rom1504:
I'm getting
> Internal Exception: io.netty.handler.codec.DecoderException: The received encoded string buffer length is longer than maximum allowed (335 > 64)

I don't understand

rom1504:
adding a timeout at https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/ping.js#L29 fixes it.
I don't really understand why

Issue #409, opened 2016-06-10
Use node's `crypto` module for RSA


roblabla:
It seems like nodejs added RSA capabilities to its crypto modules at some point. https://nodejs.org/api/crypto.html#crypto_crypto_privatedecrypt_private_key_buffer .

Using this would avoid people the painful install process. It looks like the browserify folks have a purejs shim for it too. Sadly it doesn't seem to use a vetted library for RSA :(.

rom1504:
Apparently it's there since 0.12 https://nodejs.org/dist/latest-v0.12.x/docs/api/crypto.html
(it's not there in 0.10 https://nodejs.org/dist/latest-v0.10.x/docs/api/crypto.html)
So indeed we could use that.
Could that replace ursa completely ?

roblabla:
Yup. We just need publicEncrypt/privateDecrypt. Only thing is, we'd have to write a bit of code to generate the keys (I don't think node's crypto comes with a key generator) but surely someone already has a module out there to do just that.

PR #506, opened 2017-07-20
remove ursa, uses crypto rsa functionalities instead, and keypair for keys generation

rom1504:
But
I'm getting errors for the public key coming from our server.

I believe the key generated by keypair isn't quite right.


rom1504:
will fix https://github.com/PrismarineJS/node-minecraft-protocol/issues/409

rom1504:
error fixed by using node-rsa instead of keypair for key generation
It's the same module used by ursa purejs.

PR #507, opened 2017-07-20
split createServer in 4 plugins

rom1504:
It seems to work but I probably forgot some things.
Does it look ok ?
will close https://github.com/PrismarineJS/node-minecraft-protocol/pull/210

### DIFF
--- a/src/server/login.js
+++ b/src/server/login.js
@@ -1,0 +1,143 @@
+const yggserver = require('yggdrasil').server({});
+const UUID = require('uuid-1345');
+const bufferEqual = require('buffer-equal');
+const crypto = require('crypto');
+const pluginChannels = require('../client/pluginChannels');
+const states = require("../states");
+
+module.exports=function(client,server,options) {
+  const {
+    'online-mode' : onlineMode = true,
+      kickTimeout = 30 * 1000,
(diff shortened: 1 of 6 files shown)


src/server/login.js
```diff
@@ -0,0 +1,143 @@
...
+      client.uuid = nameToMcOfflineUUID(client.username);
+    }
+    if (server.mcversion.version >= 27) { // 14w28a (27) added whole-protocol compression (http://wiki.vg/Protocol_History#14w28a), earlier versions per-packet compressed TODO: refactor into minecraft-data
+      client.write('compress', { threshold: 256 }); // Default threshold is 256
+      client.compressionThreshold = 256;
+    }
+    client.write('success', {uuid: client.uuid, username: client.username});
+    client.state = states.PLAY;
+    loggedIn = true;
+    if(enableKeepAlive) client._startKeepAlive();
```
roblabla: I very much dislike that _startKeepAlive is not in the keepalive plugin. Instead we should listen for the `state` event (https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client.js#L138) and if the new state is `states.PLAY`, start the keepalive. This way, keepalive stays self-contained.

PR #508, opened 2017-07-21
dynamic autoversion server

rom1504:
will (finally) close https://github.com/PrismarineJS/node-minecraft-protocol/issues/234

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -18,17 +18,19 @@ function createServer(options={}) {
     port = serverPort || 25565,
     motd = "A Minecraft server",
     'max-players' : maxPlayers = 20,
-    version : optVersion = require("./version").defaultVersion,
+    version,
     favicon,
     customPackets
   } = options;
 
+  const optVersion = version===undefined || version===false ? require("./version").defaultVersion : version;
+
   const mcData=require("minecraft-data")(optVersion);
-  const version = mcData.version;
+  const mcversion = mcData.version;
 
 
-  const server = new Server(version.minecraftVersion,customPackets);
-  server.mcversion=version;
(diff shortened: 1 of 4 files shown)


rom1504:
version: false

If you could make a PR to explain this feature in the part of the doc you'd
expect it to be, it would be appreciated

jojomatik:
That was fast, thanks :)

```js
const wakeUpListener = createServer({
      port: this.port,
      "online-mode": this.properties.onlineMode,
      motd: "",
      maxPlayers: this.properties.maxPlayers,
      favicon: this.favicon,
      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
      // @ts-ignore
      version: false,
    });
```
Minecraft still shows an incompatible client, do I misunderstand the feature? I want to disable the warning for every version:
image

I also don't see how this line would differentiate between `undefined` and `false`?
https://github.com/PrismarineJS/node-minecraft-protocol/blob/9cff34efc0ac35fb3693a766133ca988ce40b93c/src/createServer.js#L30

If I find out what's wrong here I can create a PR, sure. *Also the types would need to be adjusted. I'm only allowed to use strings.*

rom1504:
I don't think we send all versions as green in the ping indeed, but you can
still connect. Indeed undefined might work as well
The way this works is we first open a connection with a default version and
use that to communicate with the client, then the client give us the
version and we switch to that other version.

We could probably also answer all good in the ping if the version is false,
feel free to look into it, it should be an easy change

jojomatik:
I've created a PR that should tackle my problems. Feel free to take a look and review :) I hope the versions are assigned correctly

PR #514, opened 2017-08-10
Changed let scope, added undefined check.

lkolbly:
I was using the example proxy because I wanted to figure out how the scoreboard protocol worked, and this one has the nifty dump feature, and got some errors running the proxy because the let is referenced in the if statement after the for loop, and also because sometimes result can be undefined in matches().

These changes got it working for me.

Thanks,
Lane


rom1504:
Nice catch.
Thanks for the PR !

PR #518, opened 2017-08-28
store minecraft jars in /tmp unless overriden by ENV variable

allain:
Without this change, tests were not succeeding.

### DIFF
--- a/test/clientTest.js
+++ b/test/clientTest.js
@@ -19,7 +19,7 @@ mc.supportedVersions.forEach(function(supportedVersion,i) {
   const PORT=Math.round(30000+Math.random()*20000);
   const mcData = require("minecraft-data")(supportedVersion);
   const version = mcData.version;
-  const MC_SERVER_JAR_DIR = process.env.MC_SERVER_JAR_DIR;
+  const MC_SERVER_JAR_DIR = process.env.MC_SERVER_JAR_DIR || '/tmp';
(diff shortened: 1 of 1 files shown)


rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol#testing

but I guess why not.

Is everyone ok with "/tmp" ?

I guess that will only work on linux, but does it matter for tests ? @roblabla @mhsjlw

test/clientTest.js
```diff
@@ -19,7 +19,7 @@ mc.supportedVersions.forEach(function(supportedVersion,i) {
   const PORT=Math.round(30000+Math.random()*20000);
   const mcData = require("minecraft-data")(supportedVersion);
   const version = mcData.version;
-  const MC_SERVER_JAR_DIR = process.env.MC_SERVER_JAR_DIR;
+  const MC_SERVER_JAR_DIR = process.env.MC_SERVER_JAR_DIR || '/tmp';
```
roblabla: You can use [`os.tmpdir()`](https://nodejs.org/api/os.html#os_os_tmpdir) to get the system's temp directory. I'd like it if this used it instead. On OSX, the tmp dir is something like `/var/folders/_1/<some random string>/T`, and on windows it's `\\temp`. Letting the stdlib take care of these kind of discrepency seems like a good idea :)

allain: Good catch. Made it use os.tmpdir().

PR #520, opened 2017-10-01
added fallback for session to normal auth

plexigras:


src/client/auth.js
```diff
@@ -2,7 +2,7 @@ const yggdrasil = require('yggdrasil')({});
 const UUID = require('uuid-1345');
 
 module.exports = function(client, options) {
-  const clientToken = options.clientToken || UUID.v4().toString();
+  const clientToken = options.clientToken || (options.session && options.session.clientToken) || UUID.v4().toString();
```
roblabla: I suppose this is a different fix ? Looks like the old `option.session` never used the clientToken 😱

plexigras: No this is so we dont use a new `clientToken` when we run the `yggdrasil.auth` fallback.

PR #523, opened 2017-11-01
added raw buffer to packet events

plexigras:


plexigras:
nope dont think so

this is something i got by doing `client.on('packet', console.log)`

```
{ location: { x: 2971, y: 40, z: -1067 },
  action: 4,
  nbtData: 
   { type: 'compound',
     name: '',
     value: 
      { Owner: [Object],
        Rot: [Object],
        x: [Object],
        y: [Object],
        z: [Object],
        id: [Object],
        SkullType: [Object] } } } { size: 339, name: 'tile_entity_data', state: 'play' }
```

rom1504:
the test is failing for some reason.

plexigras:
idk, maybe rerun the test

rom1504:
I can't, it says I need write access on your branch to rerun them. I think you can though.

rom1504:
hmm nevermind, i wasn't connected. rerunning

rom1504:
Test still failing. I believe these changes actually make it fail.

plexigras:
why what is the problem i don't get where the errors are coming from

rom1504:
I believe something in the tests assume packets events don't have that additional argument ?
I don't really know, just run the tests and figure it out.

plexigras:
this is the error

```
     Uncaught TypeError: Cannot read property 'emit' of null
      at Socket.onData (node_modules/minecraft-wrap/lib/wrap_server.js:124:26)
```

https://github.com/PrismarineJS/node-minecraft-wrap/blob/master/lib/wrap_server.js#L124

i don't know who it has anything to do with nmp it seams like a completely different issue.

maybe the mojang servers where down or something idk.

rom1504:
Oh ok yeah it passes.

Can you add that to the doc ? 

And are you sure it doesn't fit into the metadata object ?

Any opinion @roblabla ?

What's the use case of this by the way ?

plexigras:
Yes maybe we should add the buffer to the metadata object instead.


Also here is an examle, its a proxy that turns all the blocks into coal.

```js
const version = '1.9.4'

const nmp   = require('minecraft-protocol')
const Chunk = require('prismarine-chunk')(version)
const Block = require("prismarine-block")(version)
const Vec3  = require('vec3').Vec3

const coalBlock = new Block(173, 1, 0)

const server = nmp.createServer({ port:25566, version: version })

server.on('login', connect)
function connect(client) {
  const remote = nmp.createClient({
    username: client.username,
    version:version,
    keepAlive: false
  })

  remote.on('packet', (data, metadata, buffer) => {
    if (metadata.state !== 'play') return
    if (metadata.name === 'map_chunk') {
      return client.write('map_chunk', loadChunk(data))
    }
    client.writeRaw(buffer)
  })
  client.on('packet', (data, metadata, buffer) => {
    remote.writeRaw(buffer)
  })
}

function loadChunk(data) {
  const chunk = new Chunk()
  chunk.load(data.chunkData, data.bitMap, true)

  const next = new Chunk()
  next.initialize((x,y,z) => {
    const block = chunk.getBlockType(new Vec3(x,y,z))
    if (block !== 0) return coalBlock
  })

  return {
    ...data,
    bitMap: 0xffff,
    chunkData: next.dump()
  }
}
```

rom1504:
I think it could have bad performance consequences.

It can be achieved by using "raw" directly

PR #525, opened 2017-11-08
Added New Example For Generating A World

oscartbeaumont:
This is a new example that shows how to generate a chunk (Using prismarine-chunk) and send it to the user. I though this might be helpful for other people. If you don't want to include it for any reason it is fine but I had just written the code and wanted to share it.

### DIFF
--- a/examples/server_world/mc.js
+++ b/examples/server_world/mc.js
@@ -1,0 +1,47 @@
+const mc = require('minecraft-protocol');
+const Chunk = require('prismarine-chunk')("1.12.1");
+const Vec3 = require('vec3');
+var server = mc.createServer({
+  'online-mode': true,
+  encryption: true,
+  host: '0.0.0.0',
+  port: 25565,
+});
+var chunk = new Chunk();
+
+for (var x = 0; x < 16;x++) {
+  for (var z = 0; z < 16; z++) {
+    chunk.setBlockType(new Vec3(x, 100, z), 2);
+    for (var y = 0; y < 256; y++) {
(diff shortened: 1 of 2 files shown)


Gjum:
I'd recommend sending the position after the chunk packet so the chunk is loaded when the player is spawned.

> ### Player Position And Look (clientbound)
> Updates the player's position on the server. **This packet will also close the “Downloading Terrain” screen when joining/respawning.**

roblabla:
TYVM, we definitely want an example like that ! If you could include the changes @Gjum asked for (to make sure we don't trigger some edge case), that'd be awesome

plexigras:
also add a `package.json` file and add `prismarine-chunk` and `vec3` as dependecys

roblabla:
the package.json should look like https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/client_chat/package.json

oscartbeaumont:
I have updated the code to add a package.json and sending the player position after sending the chunk.

examples/server_world/mc.js
```diff
@@ -0,0 +1,47 @@
+const mc = require('minecraft-protocol');
+const Chunk = require('prismarine-chunk')("1.12.1");
+const Vec3 = require('vec3');
+var server = mc.createServer({
+  'online-mode': true,
+  encryption: true,
+  host: '0.0.0.0',
+  port: 25565,
```
rom1504: Can you add version:"1.12.1" here ?

oscartbeaumont:
Done. The Minecraft version is now in the createServer function.

oscartbeaumont:
Thanks.

Issue #538, opened 2018-01-15
Unhandled errors on Client connection failure?


ethanent:
My code looks a bit like this:

```js
let client = mc.createClient({
	host, // host (defined elsewhere)
	port, // port (defined elsewhere)
	'username': 'testacc',
	'keepAlive': false,
	'version': false
})

client.on('error', (err) => {
	console.error('Connection error: ' + err)
	client.end('Onticord > An error occurred while routing to a server.')
})

client.socket.on('error', (err) => {
	console.error('Connection error: ' + err)
	client.end('Onticord > An error occurred while routing to a server.')
})
```

So I'm trying to handle connection issues, but when the connection can't be established, the whole Node process crashes.

```
Connection error: Error: connect ECONNREFUSED 127.0.0.1:25566
Connection error: Error: connect ECONNREFUSED 127.0.0.1:25566
/Users/ethan/Stuff/Projects/onticord/onticord/node_modules/minecraft-protocol/src/client/autoVersion.js:14
    if (err) throw err; // hmm
             ^

Error: connect ECONNREFUSED 127.0.0.1:25566
    at Object._errnoException (util.js:1031:13)
    at _exceptionWithHostPort (util.js:1052:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1195:14)
```

oscartbeaumont:
It is running the console.error statement and closing because there is no more code to run. You need to put some code in where you have the console.error statements so that it does not close.

rom1504:
The bug is that throw.
We should be doing client.error instead

rom1504:
I mean client.emit('error',...)

ethanent:
👍  That's what I figured. Thought it was something I was doing and that I was going crazy. 😆

ethanent:
Patched with merged PR #539!

PR #539, opened 2018-01-16
Emit errors instead of throwing.

ethanent:
It's probably good to prevent connected clients from crashing servers at all costs, for obvious reasons. :)

#538

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -194,7 +194,7 @@ class Client extends EventEmitter
 
   setEncryption(sharedSecret) {
     if (this.cipher != null)
-      throw new Error("Set encryption twice !");
+      this.emit('error','Set encryption twice!');
     this.cipher = crypto.createCipheriv('aes-128-cfb8', sharedSecret, sharedSecret);
     this.cipher.on('error', (err) => this.emit('error', err));
     this.framer.unpipe(this.socket);
--- a/src/client.js
+++ b/src/client.js
@@ -194,7 +194,7 @@ class Client extends EventEmitter
 
   setEncryption(sharedSecret) {
     if (this.cipher != null)
-      this.emit('error','Set encryption twice!');
+      this.emit('error', new Error('Set encryption twice!'));
     this.cipher = crypto.createCipheriv('aes-128-cfb8', sharedSecret, sharedSecret);
     this.cipher.on('error', (err) => this.emit('error', err));
     this.framer.unpipe(this.socket);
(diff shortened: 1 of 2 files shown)


rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol/search?utf8=%E2%9C%93&q=throw&type=

@roblabla do you remember if there was any reason we wanted to throw in these places ?

For example I'm not sure if these errors are recoverable

roblabla:
The only throws that need removal are : 

- https://github.com/PrismarineJS/node-minecraft-protocol/blob/6ce09ba9b0369d1f787267fbad0842c6bbae78dd/src/client.js#L197
- https://github.com/PrismarineJS/node-minecraft-protocol/blob/6246e0e6442a4133c37ff5306604a48ade3798cd/src/client/autoVersion.js#L14
- https://github.com/PrismarineJS/node-minecraft-protocol/blob/6246e0e6442a4133c37ff5306604a48ade3798cd/src/client/autoVersion.js#L33

So they don't crash the whole server if someone sends something weird. And I guess the try/catch in the framer could turn the non-PartialRead errors into emits instead, but it shouldn't be necessary. minecraft.js should still be throwing PartialReadErrors though, that's how protodef signals that it needs more data.

ethanent:
@roblabla Ok, I went ahead and added the fix for the `"Set encryption twice!"` error to the PR's proposed `src/client.js`!

src/client/autoVersion.js
```diff
@@ -30,7 +30,7 @@ module.exports = function(client, options) {
     .sort(function (a, b) { return b.version - a.version })
     .concat(minecraft_data.postNettyVersionsByProtocolVersion["pc"][protocolVersion]||[])
     if (versions.length === 0) {
-      throw new Error(`unsupported/unknown protocol version: ${protocolVersion}, update minecraft-data`);
+      client.emit('error',`unsupported/unknown protocol version: ${protocolVersion}, update minecraft-data`);
```
rom1504: This should be client.emit('error', new Error(...))

src/client.js
```diff
@@ -194,7 +194,7 @@ class Client extends EventEmitter
 
   setEncryption(sharedSecret) {
     if (this.cipher != null)
-      throw new Error("Set encryption twice !");
+      this.emit('error','Set encryption twice!');
     this.cipher = crypto.createCipheriv('aes-128-cfb8', sharedSecret, sharedSecret);
```
rom1504: Same here, you need to create the error object

ethanent:
Thanks, @rom1504! Both non-Error `error` emits revised.

PR #551, opened 2018-04-30
Resolve issue with refreshing sessions

CS-Birb:
Resolves issue whereby access token was passed to callback instead of the session object, causing an unhandled error.


rom1504:
good catch, thanks !

PR #555, opened 2018-05-07
Fix offline UUID generation

omegablitz:
Replaced the (not working) `new UUID` approach with a simpler, working, substring solution.


rom1504:
Is it working or not ? Last time I tried it was working.

omegablitz:
This was resolved, got confused between the `id` and `uuid` fields... my bad!

PR #556, opened 2018-05-10
[MAJOR] Fixed some consistency & Renamed options

ShayBox:
Moved to eslint from editorconfig
(More consistent coding style)
Added rules based on existing code, and fixed inconsistency
Renamed max-players to maxPlayers
Renamed online-mode to onlineMode

Note: If there's a coding style you don't like let me know, i'll change it and update the code to reflect it.


rom1504:
Hi,
If we're going to standardize the code style of node-minecraft-protocol, I'd prefer we use standard like we already do in mineflayer.
See https://github.com/PrismarineJS/mineflayer/pull/547
And https://github.com/PrismarineJS/mineflayer/blob/master/.eslintrc and https://github.com/PrismarineJS/mineflayer/blob/master/package.json#L8 and https://standardjs.com/

ShayBox:
So close it.

rom1504:
Why ? Other commits can be pushed to a pull request.
If you want to open a new one you can though.

PR #566, opened 2018-06-03
added session data to session even

plexigras:


rom1504:
:+1: 
please add to doc https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/doc/README.md

PR #572, opened 2018-07-10
Add MCLeaks support

funniray:
This is all that's needed for MCLeaks to work. Here's how you would authenticate (Side note, this requires my fork of node-mcleaks)
```JS
const options = {
    username: "MCLeaks", //You have to put something here
    mcLeaks: true,
    token: "huidfshdfushsdf" //Your MCLeaks token
};
```
```JS
    mcleaks.redeem(options, function (err, data) { //options.token should be defined
        if (err) {
            log("Something went wrong with mcLeaks authentication.");
            throw err;
        }
        options.session = {
            accessToken: data.result.session,
            selectedProfile: {
                name: data.result.mcname
            }
        };
        options.token = undefined;
        cb(mineflayer.createBot(options));
    });
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "lodash.get": "^4.1.2",
     "lodash.merge": "^4.3.0",
     "minecraft-data": "^2.26.0",
+    "node-mcleaks": "^1.0.3",
     "node-rsa": "^0.4.2",
     "prismarine-nbt": "^1.2.0",
     "protodef": "^1.6.7",
--- a/src/client/auth.js
+++ b/src/client/auth.js
@@ -20,7 +20,9 @@ module.exports = function (client, options) {
       }
     }
 
-    if (options.session) {
+    if (options.mcLeaks && options.session) {
+      cb(null, options.session);
+    } else if (options.session) {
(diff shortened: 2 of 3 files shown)


rom1504:
Hi,
No, let's not add mcleaks as a dependency.
You can probably change nmp to add an option to do what you want without actually putting the mcleaks code in nmp

SuperOP535:
yeah, i dont know if nmp supports plugins but that'd be a good idea

funniray:
Yeah, if adding plugins was possible (for logging in and such) were possible, then I would do that. I'm just not sure of the best way to do this currently.

roblabla:
https://github.com/PrismarineJS/node-minecraft-protocol/blob/203c98eef75365d42b2de0f9a04dfd69e8f8be2b/src/client/encrypt.js#L22 Could take joinServerRequets in the options ?

rom1504:
please re-open / recreate one with something generic

leocornelius:
Hey, sorry to necro. WIll this still work?

funniray:
Probably, but you would have to patch it in yourself for newer versions of PrismarineJS

PR #580, opened 2018-08-23
Fix autoVersion

lluiscab:
Simple fix for autoVersion, ends the execution of the `ping` callback if an error is found.
Prevents more errors from getting logged to the console if the server is not reachable.


rom1504:
thank you, good catch

PR #585, opened 2018-08-26
Update default version to 1.13.1

DeliciousJaffa:


rom1504:
why not, but it doesn't change much.
It's basically used only when you create a server.
for clients, it uses autoversion

PR #587, opened 2018-08-28
Test cleanup

lluiscab:
I've cleaned up the test files in preparation for the non-vanilla server tests. Test output is now cleaner and it's easier to see what's going on.

### DIFF
--- a/test/benchmark.js
+++ b/test/benchmark.js
@@ -13,39 +13,32 @@ const testDataWrite = [
 ]
 
 const {firstVersion, lastVersion} = require('./common/parallel')
-console.log({firstVersion, lastVersion})
+console.log(`Testing versions from ${firstVersion} to ${lastVersion}`)
 
 mc.supportedVersions.forEach(function (supportedVersion, i) {
   if (!(i >= firstVersion && i <= lastVersion)) { return }
 
   const mcData = require('minecraft-data')(supportedVersion)
   const version = mcData.version
-  describe('benchmark ' + version.minecraftVersion, function () {
+  describe(`Benchmark for ${version.minecraftVersion}`, function () {
     this.timeout(60 * 1000)
     const inputData = []
-    it('bench serializing', function (done) {
+    it('Bench serializing', function (done) {
       const serializer = mc.createSerializer({state: states.PLAY, isServer: false, version: version.minecraftVersion})
-      let start, i, j
-      console.log('Beginning write test')
-      start = Date.now()
+      let i, j
       for (i = 0; i < ITERATIONS; i++) {
         for (j = 0; j < testDataWrite.length; j++) {
           inputData.push(serializer.createPacketBuffer(testDataWrite[j]))
         }
       }
-      const result = (Date.now() - start) / 1000
-      console.log('Finished write test in ' + result + ' seconds')
       done()
     })
 
-    it('bench parsing', function (done) {
+    it('Bench parsing', function (done) {
       const deserializer = mc.createDeserializer({state: states.PLAY, isServer: true, version: version.minecraftVersion})
-      console.log('Beginning read test')
-      const start = Date.now()
       for (let j = 0; j < inputData.length; j++) {
         deserializer.parsePacketBuffer(inputData[j])
       }
-      console.log('Finished read test in ' + (Date.now() - start) / 1000 + ' seconds')
       done()
     })
   })
--- a/test/clientTest.js
+++ b/test/clientTest.js
@@ -11,8 +11,6 @@ const { Wrap, download } = require('minecraft-wrap')
 
 const {firstVersion, lastVersion} = require('./common/parallel')
 
-
-
 mc.supportedVersions.forEach(function (supportedVersion, i) {
   if (!(i >= firstVersion && i <= lastVersion)) return
 
@@ -75,12 +73,12 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
           delete results.latency
           delete results.favicon
           delete results.players
-          delete results.version.protocol
           assert.deepEqual(results, {
             version: {
-              name: version.minecraftVersion
+              name: version.minecraftVersion,
+              protocol: version.version
             },
-            description: MOTD
+            description: results.version.protocol > 5 ? { text: MOTD } : MOTD
           })
           done()
         })
--- a/test/clientTest.js
+++ b/test/clientTest.js
@@ -78,7 +78,7 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
               name: version.minecraftVersion,
               protocol: version.version
             },
-            description: results.version.protocol > 5 ? { text: MOTD } : MOTD
+            description: results.version.protocol > 47 ? { text: MOTD } : MOTD // Post 1.8 MOTD is sent as object
           })
           done()
         })
@@ -241,15 +241,10 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
           port: PORT
         })
         client.on('error', err => done(err))
-        let gotKicked = false
         client.on('disconnect', function (packet) {
           assert.ok(packet.reason.indexOf('"Failed to verify username!"') !== -1 || packet.reason.indexOf('multiplayer.disconnect.unverified_username') !== -1)
-          gotKicked = true
-        })
-        client.on('end', function () {
-          assert.ok(gotKicked)
-          client.end()
-          done()
+          client.end();
+          done();
         })
       })
     })
--- a/test/clientTest.js
+++ b/test/clientTest.js
@@ -7,16 +7,17 @@ const assert = require('power-assert')
 const SURVIVE_TIME = 10000
 const MC_SERVER_PATH = path.join(__dirname, 'server')
 
-const Wrap = require('minecraft-wrap').Wrap
+const { Wrap, download } = require('minecraft-wrap')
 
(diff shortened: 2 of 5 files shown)


lluiscab:
Seems like the parallel detection is broken or something like that as it's only testing the first two versions. I don't really have an idea of how that file should work, any help @rom1504 ?

rom1504:
No the parallel detection is working as expected. It runs 2 versions by container.
Each container is represented by a tab in circle ci (there are 4 containers)

test/benchmark.js
```diff
@@ -13,39 +13,32 @@ const testDataWrite = [
...
   const version = mcData.version
-  describe('benchmark ' + version.minecraftVersion, function () {
+  describe(`Benchmark for ${version.minecraftVersion}`, function () {
     this.timeout(60 * 1000)
     const inputData = []
-    it('bench serializing', function (done) {
+    it('Bench serializing', function (done) {
       const serializer = mc.createSerializer({state: states.PLAY, isServer: false, version: version.minecraftVersion})
-      let start, i, j
-      console.log('Beginning write test')
```
rom1504: Why did you remove these console.log ?

test/clientTest.js
```diff
@@ -25,11 +24,9 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
     minMem: 1024,
     maxMem: 1024
   })
-  wrap.on('line', function (line) {
-    console.log(line)
-  })
+  // wrap.on('line', console.log);
```
rom1504: Why did you remove this ? This is very useful to know what didn't work in a test.

test/clientTest.js
```diff
@@ -25,11 +24,9 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
...
 
-  describe('client ' + version.minecraftVersion, function () {
+  describe(`Testing minecraft client ${version.minecraftVersion}`, function () {
```
rom1504: Should be "Testing minecraft-protocol client" (this is using the minecraft server)

test/clientTest.js
```diff
@@ -41,31 +38,31 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
       })
     })
 
-    describe('offline', function () {
+    describe('Offline mode tests', function () {
       before(function (done) {
-        console.log(new Date() + 'starting server ' + version.minecraftVersion)
+        console.log(`      Starting minecraft server for ${version.minecraftVersion}`)
```
rom1504: Please leave the date, it is useful to see what took time

test/clientTest.js
```diff
@@ -41,31 +38,31 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
...
         wrap.startServer({
           'online-mode': 'false',
           'server-port': PORT,
-          'motd': 'test1234',
+          'motd': MOTD,
           'max-players': 120
         }, function (err) {
           if (err) { console.log(err) }
-          console.log(new Date() + 'started server ' + version.minecraftVersion)
+          console.log(`      Started minecraft server for ${version.minecraftVersion}`)
```
rom1504: Leave the date

test/clientTest.js
```diff
@@ -153,15 +151,14 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
...
-      it('produce a decent error when connecting with the wrong version', function (done) {
+      it('Produces a decent error when connecting with the wrong version', function (done) {
         const client = mc.createClient({
           username: 'Player',
           version: version.minecraftVersion === '1.8.8' ? '1.11.2' : '1.8.8',
           port: PORT
         })
         client.once('error', function (err) {
           if (err.message.startsWith('This server is version')) {
-            console.log(new Date() + 'Correctly got an error for wrong version : ' + err.message)
```
rom1504: Don't remove this

test/clientTest.js
```diff
@@ -237,20 +234,15 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
...
         const client = mc.createClient({
           username: 'Player',
           version: version.minecraftVersion,
           port: PORT
         })
         client.on('error', err => done(err))
-        let gotKicked = false
         client.on('disconnect', function (packet) {
           assert.ok(packet.reason.indexOf('"Failed to verify username!"') !== -1 || packet.reason.indexOf('multiplayer.disconnect.unverified_username') !== -1)
-          gotKicked = true
```
rom1504: Why ?

test/common/parallel.js
```diff
@@ -14,7 +14,7 @@ function testedRange (nodeIndex, nodeTotal, numberOfVersions) {
 
   return {firstVersion: nodeIndex * nbFirsts, lastVersion: (nodeIndex + 1) * nbFirsts - 1}
 }
-console.log({nodeIndex, nodeTotal, versions: mc.supportedVersions.length})
+// console.log({nodeIndex, nodeTotal, versions: mc.supportedVersions.length})
 const {firstVersion, lastVersion} = parallel ? testedRange(nodeIndex, nodeTotal, mc.supportedVersions.length) : {firstVersion: 0, lastVersion: mc.supportedVersions.length - 1}
```
rom1504: Use for debugging

test/clientTest.js
lluiscab: Tests shouldn't log useless data, they should provide a clean output so the user can see what's going on. It's not really useful to have the minecraft server log. Don't really know why you would want that

test/benchmark.js
lluiscab: What's the reason to have them?

test/clientTest.js
lluiscab: Will update with more precise texts

test/clientTest.js
lluiscab: If you want to know the startup time, then extract it from the server log and print it with the server started text.

test/clientTest.js
lluiscab: If you want to know the startup time, then extract it from the server log and print it with the server started text.

test/clientTest.js
lluiscab: Why not? In any case we should log something when we get an unexpected error, we shouldn't log "everything is going ok" messages

test/clientTest.js
lluiscab: There's no reason to wait for the bot to end, therefore the `gotKicked` variable is not needed.

test/common/parallel.js
lluiscab: Could be changed into a formatted console.log message

test/benchmark.js
rom1504: To know how much time the benchmark take ? It's a benchmark !

test/clientTest.js
rom1504: This is the most useful data in the client tests. The client tests are testing against the vanilla server so we want to know everything we can.

In practice it was possible to catch several problems using the vanilla server logs (recently that a test was stuck even though the vanilla server correctly saw the player log in)

test/clientTest.js
rom1504: How is that better ?
This is 100% reliable even if the minecraft wrap or the vanilla server have bugs (which is not that uncommon)

test/common/parallel.js
rom1504: Feel free to improve how it looks but the information should be printed

test/clientTest.js
rom1504: ???
The purpose of the test is to make sure we get kicked and exit properly, so yes this is required.

test/clientTest.js
rom1504: Not convinced. I only recently fixed this test and this message was very useful to know there was no problem with the test and actually there was a problem with some other unrelated error happening here.

rom1504:
As said we want to have as much info as possible because it's very common these tests fail.

If anything we are not display enough logs. I think we should also display the debug log but circle ci doesn't support that many logs. Could be saved to a file for circle ci to display some better way (would need to look at circle ci doc)

test/benchmark.js
lluiscab: Mocha already tells you

test/clientTest.js
lluiscab: Then use start and end times and get the difference. Printing random dates to console makes things hard to read.

test/clientTest.js
lluiscab: Not if you only finish the test when you get disconnected and the reason is that you got kicked. Otherwise send an error. Make things simple

lluiscab:
Let's say a test fails, it fails becauae you've called the `done` function with an error, you get that error logged onto the console when mocha ends, don't really know why we'd log the minecraft server or some random dates

test/clientTest.js
rom1504: Well it's not random it's the continuity of all the server log printing the time. But yes we could also print the difference

test/clientTest.js
rom1504: "disconnect" event isn't "you got disconnected", it's a packet.
We received that packet, mark that we got kicked, and then when the client end make sure we indeed for that packet before getting disconnected

rom1504:
Because the error in the callback is often not enough information to know what is wrong.

rom1504:
Tests aren't only there to say "something doesn't work".
They are also there to say "it fails for that and that reasons, here are some clues how to fix it"

rom1504:
Let's redo or close this.

PR #589, opened 2018-08-30
add dts for better autocomplete suggestion

bangbang93:
add dts for better autocomplete suggestion
image

No js code change, so should not have any side effects

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ test/npm-debug.log
 test/server
 package-lock.json
 versions/
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ test/npm-debug.log
 test/server
 package-lock.json
 versions/
-.idea/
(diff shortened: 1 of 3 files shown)


rom1504:
Looks good. Do you know if most ide handle this ?

rom1504:
And could we have some tool checking that the type definition and the code match ? (Otherwise I'm afraid it'll be out of sync soon)

bangbang93:
Both VS Code and Webstorm supports typescript definition by default. The others I have not use them so I'm not sure.
The definition cannot be auto generated when js code was changed.
Definition test can use the typescript compiler. But it only checks type definition, not the underlying js object.
This can only be manually updated unless the project is written in typescript.

rom1504:
thanks, merged !

ShayBox:
Finally typings 😄 Now you can use them to generate docs, and as long as they're kept up to date when changes are made you can have good updated docs

ShayBox:
Thought it would be great if you could not use the any type, and make the typings using strict mode, so they're actually usable for anything but auto completion.

jwulf:
The best way to ensure typings do not drift from the code is to convert the code to TypeScript.

PR #595, opened 2018-10-03
Updated typings

ShayBox:
Removed I from interfaces.
Renamed interfaces to be more clear.
Added TODO: Create typings on protodef to define here.
Added supportedVersions static strings.
Rearranged order of exports and properties.
Changed unused returns of functions from any to void (More strict).
Put entire typings inside module declaration.

Now works in strict mode.
Suggestion: Add typings to protodef so createSerializer and createDeserializer have valid types.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,6 +1,7 @@
 import { EventEmitter } from 'events';
 import { Socket } from 'net'
 import * as Stream from 'stream'
+import { on } from 'cluster';
 
 declare module 'minecraft-protocol' {
 	export class Client extends EventEmitter {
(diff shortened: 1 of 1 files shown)


rom1504:
Tests are failing, please fix

ShayBox:
Yeah that's not my problem, fix your tests 
image

rom1504:
ok seems there's a pb with circle ci.
Do you have passing tests when running `npm test` locally ?

rom1504:
ah yeah we don't have tests for the typing here.
having something like https://github.com/PrismarineJS/node-minecraft-data/blob/master/package.json#L8 might make sense here.
It's kind of sad that there's no verification the type match anything

ShayBox:
Well normally to typematch you write your tests in typescript instead of js, that way you run `tsc` which compiles your code to js, tests the typings, then run the compiled js to test the actual lib code.

ShayBox:
Also a pro tip, you can make a test script, and a pretest script, have the pretest run "tsc" and test run "node file.js" and then you just run "npm test" in cmd, and it runs pretest then test, you can also do the same with start and prestart, those are the only two that I know of

ShayBox:
And as far as I can tell the typings are working, I've used them in my typescript project and compile successfully now, and still provide intellisense in editor 👍

rom1504:
Alright

PR #602, opened 2018-11-09
Add skipValidation flag for authentication

vlakreeh:
Added simple flag to skip validation of provided session as suggested [here](https://github.com/PrismarineJS/node-minecraft-protocol/issues/481#issuecomment-294284297).


rom1504:
Thanks for the PR, sorry for taking so long to look at it !
Merged

PR #609, opened 2018-12-18
Add related helper mapItem library

Sceat:
https://github.com/AresRPG/aresrpg-map-colors


Sceat:
for people who don't know why they can't send 128 columns in the packet it's due to a bug in prismarine protocol with signed bytes
see https://github.com/AresRPG/aresrpg-map-colors/issues/1#issuecomment-448035003

PR #610, opened 2018-12-22
Distinguish socket ending from other end reasons

Gjum:
Ideally all different end reasons would be expressed, but at least this is a start.


rom1504:
Yeah good idea

PR #621, opened 2019-03-11
Fix crash when pinging with an unsupported version

Moudoux:
This library will crash if you ping a server created with it using an unsupported Minecraft version (Tested with the latest 1.14 snapshot 19w09a). This will fix it and stop it from crashing.

### DIFF
--- a/src/transforms/serializer.js
+++ b/src/transforms/serializer.js
@@ -17,6 +17,10 @@ function createProtocol (state, direction, version, customPackets) {
   const proto = new ProtoDef(false)
   proto.addTypes(minecraft)
   const mcData = require('minecraft-data')(version)
+  if (mcData === null) {
+    // Unsupported protocol
+    return createProtocol(state, direction, "1.13.2", customPackets)
+  }
   proto.addProtocol(merge(mcData.protocol, get(customPackets, [mcData.version.majorVersion])), [state, direction])
   protocols[key] = proto
   return proto


rom1504:
What error do you get exactly and in what context ?
The way you fixed it is wrong, we definitely don't want to force using some other version.

Moudoux:
Pinging a server using 19w05a results in this: 

```
C:\Users\que\Desktop\server\node_modules\minecraft-protocol\src\transforms\framing.js:66
          } else { throw e }
                   ^

TypeError: Cannot read property 'protocol' of null
    at createProtocol (C:\Users\que\Desktop\server\node_modules\minecraft-protocol\src\transforms\serializer.js:20:34)
    at createSerializer (C:\Users\que\Desktop\server\node_modules\minecraft-protocol\src\transforms\serializer.js:26:25)
    at Client.setSerializer (C:\Users\que\Desktop\server\node_modules\minecraft-protocol\src\client.js:38:23)
    at Client.set state [as state] (C:\Users\que\Desktop\server\node_modules\minecraft-protocol\src\client.js:104:10)
    at Client.onHandshake (C:\Users\que\Desktop\server\node_modules\minecraft-protocol\src\server\handshake.js:18:20)
    at Object.onceWrapper (events.js:276:13)
    at Client.emit (events.js:188:13)
    at FullPacketParser.deserializer.on (C:\Users\que\Desktop\server\node_modules\minecraft-protocol\src\client.js:82:12)
    at FullPacketParser.emit (events.js:188:13)
    at addChunk (C:\Users\que\Desktop\server\node_modules\readable-stream\lib\_stream_readable.js:297:12)```

rom1504:
Ok so an error event should be emitted in that case and we should drop the
connection to that client.
That's the correct fix

Moudoux:
Should it not send an out of date packet to the client, this is just when pinging, the server should not drop the connection and say its offline, it should say out of date for the version

matthi4s:
I would suggest an option in the `createServer` function, e.g. a `fallbackVersion` or some kind of callback like `beforePing`. If such an option is not set/`false` (default) there would be just an error event and a connection drop as suggested by @rom1504.

PR #646, opened 2019-08-12
Get the actual socket object from Socks#CreateConnection

MythTheWolf:
In newer versions of socks, we need to obtain the socks object from setSocket(), as CreateConnection does not automatically supply it. (This will make the socks example work in examples/)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -142,7 +142,7 @@ class Client extends EventEmitter {
 
     const onError = (err) => this.emit('error', err)
 
-    this.socket = socket
+    this.socket = socket.socket
 
     if (this.socket.setNoDelay) { this.socket.setNoDelay(true) }
 


rom1504:
Which node version would that be ?
As you can see, the tests say it doesn't work with currently tested node version

MythTheWolf:
Socks v2.3.2
v10.16.2

roblabla:
This change is wrong. You're supposed to pass a `net.Socket` to setSocket (`setSocket` is not socks-specific). Old version of `SocksClient.createConnection` returned such an object directly, but newer version (since v2) returns some kind of info object wrapper (see the [Quick Start Example](https://www.npmjs.com/package/socks#quick-start-example) of socks). Just pass `info.socket` to setSocket and it should work.

MythTheWolf:
Ah, I see.

PR #662, opened 2019-11-15
socket end timeout

matthi4s:
Currently, calling `client.end()` might result in a hanging socket if the other side doesn't send back the FIN packet. On a server, this is a potential memory leak and can even result in crashes due to open file limits. I don't know what's the best solution for this problem, I've adapted the [solution](https://github.com/websockets/ws/blob/master/lib/websocket.js) from the `ws` library, which works fine.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -10,6 +10,8 @@ const states = require('./states')
 const createSerializer = require('./transforms/serializer').createSerializer
 const createDeserializer = require('./transforms/serializer').createDeserializer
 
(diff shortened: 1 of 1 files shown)


rom1504:
indeed that fixes that issue, thanks for the PR

Sceat:
Any reason for waiting that long ? I'm running tests with mineflayer and once the bot is kicked I want the server to close, but the node process will hang for 30s

PR #665, opened 2019-12-22
Add connect to types

ShayBox:
Add missing connect function to typings for Client
<https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client.js#L222>

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -17,8 +17,8 @@ declare module 'minecraft-protocol' {
 		username: string
 		uuid: string
 		protocolVersion: number
+		connect(port: string, host: string): void
 		end(reason: string): void
-		connect(address: string, port: string): void
 		registerChannel(name: string, typeDefinition: any, custom?: boolean): void
 		unregisterChannel(name: string): void
 		write(name: string, params: any): void
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -17,7 +17,7 @@ declare module 'minecraft-protocol' {
 		username: string
 		uuid: string
 		protocolVersion: number
-		connect(port: string, host: string): void
+		connect(port: number, host: string): void
 		end(reason: string): void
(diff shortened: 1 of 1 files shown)


ShayBox:
I would also like to point out the typings export "States" but the actual code exports "states", whichever one should be changed I don't know.

rom1504:
the typing should be changed probably

ShayBox:
Documentation is the clear downfall of this library,
- Typings aren't complete or updated consistently
- The manually written documentation markdown isn't updated consistently
- And there's no other form of documentation like (ex. jsdoc) that is

I think there needs to be a big focus on getting some form of documentation on this project complete, and kept up to date with the actual source, whether that's typescript definitions or jsdoc comments.

There's also a few inconsistent styling choices, like in ServerOptions, all of the options are camelCase except for "online-mode", or having both callbacks (es5) and events (es6).

rom1504:
I don't think jsdoc would help at all here.  Especially since there have
been basically no change in the API for years.
There are very few things that should be used from the library, basically
only createClient and createServer.
There are lot of examples for all the possible use cases.

If you're interested in working on docs, I think something that would be
useful is a hand written tutorial on the basic js concept here (callback
async), some basic network explanation (tcp, packets, serialization), and
what this lib can be used for.
The majority of people that get confused here are people who miss big area
of knowledge in some of these things.

ShayBox:
I mentioned jsdoc as an example, I thought it would likely be updated more being comments visible from the code.

There's a lot about this project I can't fix in a pr, the mix of ES5 and ES6 mainly, but I fixed some small things here <https://github.com/ShayBox/node-minecraft-protocol/commit/b327f93bcf38df77b5e436a6059c78ab47612c41>

rom1504:
Yeah what I think wouldn't help is comment in code and doc generated from
that. You lose the coherence when doing that which is probably the most
important thing to understand for this lib

PR #676, opened 2020-02-07
Pass through http agent option to yggdrasil for proxy support

IdanHo:
To be merged only if and when PrismarineJS/node-yggdrasil#36 is merged.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -21,10 +21,8 @@ automatically logged in and validated against mojang's auth.
  * customPackets (optional) : an object index by version/state/direction/name, see client_custom_packet for an example
  * errorHandler : A way to override the default error handler for client errors. A function that takes a Client and an error.
  The default kicks the client.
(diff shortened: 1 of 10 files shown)


wvffle:
Bunp the version when PrismarineJS/node-yggdrasil#36 gets merged and published

IdanHo:
> can you update the proxy example and the doc so people know how to use this ?

i can update the doc, but updating the proxy example would require a package like ProxyAgent (https://www.npmjs.com/package/proxy-agent) in order to actually pass through proxy settings via the http agent field, is that fine?

rom1504:
yes you can add the dependency in the package.json of the example

IdanHo:
done.

rom1504:
@IdanHo I released 1.3.0 of yggdrasil with your changes, can you update dependency in package.json here ?

IdanHo:
done.

PR #680, opened 2020-02-29
Fix clientToken to be updated

Toshimichi0915:
The clientToken should be updated otherwise it cannot be used from mineflayer

### DIFF
--- a/src/client/auth.js
+++ b/src/client/auth.js
@@ -16,6 +16,7 @@ module.exports = function (client, options) {
       } else {
         client.session = session
         client.username = session.selectedProfile.name
+        options.clientToken = session.clientToken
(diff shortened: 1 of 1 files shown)


src/client/auth.js
```diff
@@ -16,6 +16,7 @@ module.exports = function (client, options) {
       } else {
         client.session = session
         client.username = session.selectedProfile.name
+        options.clientToken = session.clientToken
         options.accessToken = session.accessToken
```
IdanHo: adding that there adds it to the options object, which the user doesnt have access to anyways, whats the point?
(If you do need access to the clientToken for some reason, its just a refrence to session.clientToken which is accesible under bot.session.clientToken)
This should be closed.

PR #684, opened 2020-03-09
Follow the authentication spec more closely.

Kneesnap:
Theoretically this shouldn't make much of a difference, but I did notice it when I was reading through the npm code. We should trust the API, not our own understanding of the API, just for safety reasons.

Information Source: https://wiki.vg/Protocol_Encryption

### DIFF
--- a/src/server/login.js
+++ b/src/server/login.js
@@ -82,8 +82,8 @@ module.exports = function (client, server, options) {
         }
         // Convert to a valid UUID until the session server updates and does
         // it automatically
-        client.username = profile.name
         client.uuid = profile.id.replace(/(\w{8})(\w{4})(\w{4})(\w{4})(\w{12})/, '$1-$2-$3-$4-$5')
+        client.username = profile.name
         client.profile = profile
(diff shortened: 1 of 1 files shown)


Kneesnap:
After further research, this is actually more important than it would initially seem.

The Mojang API is case-insensitive, theoretically someone could login with a username of the wrong case, and NMP would accept it. Not the end of the world, but should be fixed.

src/server/login.js
```diff
@@ -82,6 +82,7 @@ module.exports = function (client, server, options) {
         }
         // Convert to a valid UUID until the session server updates and does
         // it automatically
+        client.username = profile.name
         client.uuid = profile.id.replace(/(\w{8})(\w{4})(\w{4})(\w{4})(\w{12})/, '$1-$2-$3-$4-$5')
```
wvffle: You've placed this just between a line and a comment reffering to that line

Kneesnap: Good point.

wvffle:
LGTM

PR #691, opened 2020-03-30
update mcdata, add explicit test for 1.16

rom1504:
This is not finished, we need to implement new way to encode/decode uuid (see https://wiki.vg/Pre-release_protocol#Login_Success) and use it where success packet is used

(cc @sbh1311 )


rom1504:
fixed it, this commit should pass the test

PR #694, opened 2020-04-13
Switch to ProtoDefNeo + more

Saiv46:
- Fixed .gitignore
- Added `benchmark` script
- Switched to [node-protodef-neo](https://github.com/Saiv46/node-protodef-neo)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 node_modules
 test/npm-debug.log
-test/server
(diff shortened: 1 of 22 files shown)


rom1504:
looks like test are failing

rom1504:
I think that we don't need this one anymore now that we have the protodef compiler, right ?

PR #695, opened 2020-04-13
Deprecate "server-port" and "max-players" + etc

Saiv46:
Changed behavior of version and max player count selection.

Also removed default value for "host" option because Node will handle it by itself: https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -13,20 +13,23 @@ module.exports = createServer
 
 function createServer (options = {}) {
   const {
-    host = '0.0.0.0',
+    host,
     'server-port': serverPort,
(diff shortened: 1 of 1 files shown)


Saiv46:
One word - consistency.

You'll neither need to use camelCased parameters (as is customary in JS development) or dash-cased parameters (like from `server.properties` file).

Saiv46:
Also, I want to implement configuration hot-reloading (`/reload` command in vanilla)

rom1504:
What prevents you to use only one of the 2 today ?

PR #701, opened 2020-04-22
Fix error in example

Heath123:
This fixes an error that occurs with the example echo client:
```TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Object```
This change might be needed in the server as well, but I didn't test that. Or maybe there's a better way to fix this


Heath123:
Oops, I think I reused the same fork for something else

PR #707, opened 2020-05-04
datatypes: properly support optvarints for elytra flying

Mstrodl:
See prismarinejs/minecraft-data#280, adds support for the `optvarints` referenced there


rom1504:
now not needed

PR #714, opened 2020-05-16
Use protodef compiler

Karang:
Goes after https://github.com/ProtoDef-io/node-protodef/pull/102


src/datatypes/nbt.json
```diff
@@ -0,0 +1,114 @@
+{
```
rom1504, line 1: Should be in prismarine-nbt (will need a PR for prismarine-nbt)

src/datatypes/nbt-compound.js
```diff
@@ -0,0 +1,55 @@
+/* global ctx */
```
rom1504, line 1: Should be in prismarine-nbt

src/datatypes/nbt.json
Karang: done

PR #721, opened 2020-05-23
Prevent ping from hanging if the server never replies. #329

Henry00IS:
This fixes the issue where a connection to a server is established, you request a ping, the server never replies but keeps the connection open for hours. I tried to keep your coding style. Using a 30 seconds timeout just like the socket client.

### DIFF
--- a/src/ping.js
+++ b/src/ping.js
@@ -4,6 +4,8 @@ const Client = require('./client')
 const states = require('./states')
 const tcpDns = require('./client/tcp_dns')
 
+const closeTimeout = 30 * 1000
+
 module.exports = ping
 
 function ping (options, cb) {
@@ -14,9 +16,11 @@ function ping (options, cb) {
   const version = mcData.version
   options.majorVersion = version.majorVersion
   options.protocolVersion = version.version
+  var closeTimer = null
 
   const client = new Client(false, version.minecraftVersion)
   client.on('error', function (err) {
+	clearTimeout(closeTimer);
     cb(err)
   })
 
@@ -25,6 +29,7 @@ function ping (options, cb) {
     const start = Date.now()
     client.once('ping', function (packet) {
       data.latency = Date.now() - start
+	  clearTimeout(closeTimer);
       cb(null, data)
       client.end()
     })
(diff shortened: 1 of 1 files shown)


Henry00IS:
Perhaps it would make sense to have a different type of callback than ETIMEDOUT as the connection was established. But your latency check depends on an actual reply of the Minecraft server itself- which you don't get in this case- so that latency result would be different from the usual results. Food for thought here.

src/ping.js
```diff
@@ -14,9 +16,11 @@ function ping (options, cb) {
   const version = mcData.version
   options.majorVersion = version.majorVersion
   options.protocolVersion = version.version
+  var closeTimer = null
```
wvffle, line 19: Unneeded

src/ping.js
```diff
@@ -14,9 +16,11 @@ function ping (options, cb) {
...
 
   const client = new Client(false, version.minecraftVersion)
   client.on('error', function (err) {
+	clearTimeout(closeTimer);
```
wvffle, line 23: - We use 2 spaces rather than tabs.
- Wrong indentation

src/ping.js
```diff
@@ -25,6 +29,7 @@ function ping (options, cb) {
     const start = Date.now()
     client.once('ping', function (packet) {
       data.latency = Date.now() - start
+	  clearTimeout(closeTimer);
```
wvffle, line 32: - Mix between tabs and spaces
- Wrong indentation

src/ping.js
```diff
@@ -45,6 +50,13 @@ function ping (options, cb) {
     })
     client.state = states.STATUS
   })
+  
+  // timeout against servers that never reply while keeping
+  // the connection open and alive.
+  closeTimer = setTimeout(function () {
```
wvffle, line 56: const timeoutId = ...

src/ping.js
```diff
@@ -45,6 +50,13 @@ function ping (options, cb) {
...
+	client.end()
+	cb(new Error("ETIMEDOUT"))
```
wvffle, line 58: Use 2 spaces nad fix the indent

wvffle, line 58: Return more human friendly error

src/ping.js
```diff
@@ -4,6 +4,8 @@ const Client = require('./client')
 const states = require('./states')
 const tcpDns = require('./client/tcp_dns')
 
+const closeTimeout = 30 * 1000
```
wvffle, line 7: this should be an options parameter

rom1504:
For the coding style just run npm run fix, it'll automatically make it right

Saiv46:
Maintainers allowed to edit this PR?

Henry00IS:
That's fine with me.

PR #729, opened 2020-06-10
fix(compression.js): Workaround

### LINKED ISSUE #664: Client connection times out when packet over 2097152 bytes (2mb) is sent to client
In vanilla Minecraft, if a packet over roughly 2mb is sent to the client it will throw the following exception

image

When a large packet like this is sent to a minecraft-protocol client, the client times out and fails to decompress and/or parse the packet.  This can be proven by modifying the /src/transforms/compressor.js file at line 69 to the following:

`if(value < 2097152 - 90000){
    this.push(newBuf)
}`

This will only parse the packet if its size is under 2mb. Upon implementing this patch, the client will successfully connect or remain connected to the server.

I can usually run minecraft-protocol easily on my system, but when an oversized packet is sent like this my computer begins slowing down and the fan begins to whir loudly, leading me to believe the process is allocating too much processing power to decompressing this packet. Below is the full error when the connection times out.  The raw bytes logged in console right before the timeout is a map_chunk packet with too much NBT data.

image

I don't think the client should attempt to modify or repair these oversized packets, as that would cause client/server desync.  If the client can parse it correctly, I see that as a successful fix.

Saiv46:
This should fix #664, I suppose

### DIFF
--- a/src/transforms/compression.js
+++ b/src/transforms/compression.js
@@ -25,7 +25,7 @@ class Compressor extends Transform {
   async _transform (chunk, enc, cb) {
     if (chunk.length >= this.compressionThreshold) {
       try {
-        const newChunk = async compress(chunk)
+        const newChunk = await compress(chunk)
         if (newChunk.length < 2097153) {
           const buf = Buffer.allocUnsafe(sizeOfVarInt(chunk.length) + newChunk.length)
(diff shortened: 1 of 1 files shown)


rom1504:
does this work ? what is it supposed to do ?

rom1504:
combining async and callback doesn't make sense

PR #731, opened 2020-06-16
Switch to @saiv46/protodef

Saiv46:


package.json
```diff
@@ -51,7 +51,7 @@
     "minecraft-data": "^2.44.0",
     "node-rsa": "^0.4.2",
     "prismarine-nbt": "^1.3.0",
-    "protodef": "^1.7.0",
+    "@saiv46/protodef": "^1.7.3",
```
rom1504, line 54: no that's not the right way to do this
you need to put `"protodef": "saiv46/protodef"` there

rom1504:
seems like you have some errors https://github.com/PrismarineJS/node-minecraft-protocol/pull/731/checks?check_run_id=777172409

would be great to identify what and add something that reproduces it as a protodef unit test

src/transforms/serializer.js
```diff
@@ -1,9 +1,9 @@
 'use strict'
 
-const ProtoDef = require('protodef').ProtoDef
-const Serializer = require('protodef').Serializer
-const Parser = require('protodef').FullPacketParser
-const { ProtoDefCompiler } = require('protodef').Compiler
+const ProtoDef = require('@saiv46/protodef').ProtoDef
+const Serializer = require('@saiv46/protodef').Serializer
+const Parser = require('@saiv46/protodef').FullPacketParser
+const { ProtoDefCompiler } = require('@saiv46/protodef').Compiler
```
wvffle, line 6: To improve overall esthetics of the code I'd use destructing there.
```js
const { ProtoDef, ..., Compiler: ProtoDefCompiler } = require(...)
```

Saiv46: I just replaced "protodef" with "@saiv46/protodef" on master branch.

wvffle: Yeah, I know but if you can fix some styling in code you change then it's welcomed :)

rom1504:
looks like you did not continue on this, please reopen if you do

PR #732, opened 2020-06-17
Update version.js

Blueil:
Please update it to the last version!


rom1504:
That's not how this works.
If you're interested to have that last snapshot, you can use https://wiki.vg/Pre-release_protocol to update protocol.json in minecraft-data

rom1504:
done

PR #734, opened 2020-06-26
Add 1.16.1 support

Heath123:
1.16.1 and 1.16 should be the the same. I couldn't make the dependencies work to test this though. examples/server_world/mc.js requires a version of prismarine-chunk that doesn't exist but that was an issue before this too.


src/version.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
 module.exports = {
-  defaultVersion: '1.16',
-  supportedVersions: ['1.7', '1.8', '1.9', '1.10', '1.11.2', '1.12.2', '1.13.2', '1.14.4', '1.15.2', '1.16']
+  defaultVersion: '1.16.1',
+  supportedVersions: ['1.7', '1.8', '1.9', '1.10', '1.11.2', '1.12.2', '1.13.2', '1.14.4', '1.15.2', '1.16', '1.16.1']
```
Saiv46, line 5: Major version duplication (1.16 and 1.16.1) - you just needed to replace `'1.16'` with `'1.16.1'`

Heath123: Fixed

Saiv46:
`node-minecraft-data` update with 1.16.1 not released yet

rom1504:
Ah sorry just saw this PR now, I already did a similar change in master, we have 1.16.1 support

PR #735, opened 2020-06-26
Added workaround

Saiv46:
Some environments do not support necessary ciphers for Minecraft, so I added `aes-js` optional dependency as fallback


package.json
```diff
@@ -29,31 +29,34 @@
...
-    "node-rsa": "^0.4.2",
+    "node-rsa": "^1.0.8",
     "prismarine-nbt": "^1.3.0",
     "protodef": "^1.8.0",
-    "readable-stream": "^3.0.6",
-    "uuid-1345": "^1.0.1"
+    "uuid-1345": "^1.0.2"
+  },
+  "optionalDependencies": {
+    "readable-stream": "^3.6.0",
```
rom1504, line 59: why is this an optional dependency

package.json
```diff
@@ -29,31 +29,34 @@
...
-    "debug": "^4.1.0",
+    "debug": "^4.1.1",
     "endian-toggle": "^0.0.0",
-    "lodash.get": "^4.1.2",
-    "lodash.merge": "^4.3.0",
+    "lodash.get": "^4.4.2",
+    "lodash.merge": "^4.6.2",
     "minecraft-data": "^2.58.0",
-    "node-rsa": "^0.4.2",
+    "node-rsa": "^1.0.8",
```
rom1504, line 53: did you check this still works ? (connect to some online server)

package.json
Saiv46: `readable-stream` resolves problems that it was in ...v0.12? Stream API is stable now and I think code would work even if `readable-stream` package is missing.

package.json
Saiv46: `node-rsa` of course has a breaking change - it just drops support of < 8.11.1

package.json
rom1504: So why not remove it as a dependency completely then ?

package.json
rom1504: So it does work with recent node ? (To connect a client to an online server)

package.json
Saiv46: I'll remove it later

src/client.js
```diff
@@ -79,11 +78,10 @@ class Client extends EventEmitter {
 
     this.deserializer.on('data', (parsed) => {
       parsed.metadata.name = parsed.data.name
-      parsed.data = parsed.data.params
       parsed.metadata.state = state
+      parsed.data = parsed.data.params
       debug('read packet ' + state + '.' + parsed.metadata.name)
-      const s = JSON.stringify(parsed.data, null, 2)
-      debug(s.length > 10000 ? parsed.data : s)
+      debug(parsed.data)
```
rom1504, line 84: Why did you change this ?

src/transforms/encryption.js
```diff
@@ -0,0 +1,53 @@
...
+class Cipher extends Transform {
+  constructor (key) {
+    super()
+    // eslint-disable-next-line
+    this.aes = new aesjs.ModeOfOperation.cfb(key, key, 8)
+  }
+
+  _transform (chunk, enc, cb) {
+    try {
+      cb(null, this.aes.encrypt(chunk))
```
rom1504, line 38: This won't work with a streaming input. Did you try it ?

rom1504:
A general comment : it's a bad idea to change many related things to the main change in the pr. This makes it much slower to merge a PR.
Better make small PRs
(For this time it's already done so let's just finish it)

src/client.js
Saiv46: When running with `DEBUG=*` - formatted JSON output tends to be over-verbose, especially with buffers

```
{
  "type": "Buffer",
  "data": [
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    ...
  ]
}
```

src/transforms/encryption.js
Saiv46: `aes-js` has a bug with CFB encryption (any other algorithms worked fine there)

Marking this PR as draft until they fix that

rom1504: Is there an open issue for this ?

Saiv46: Yes, I'm also making PR

rom1504:
please reopen whenever you want to work on this

PR #739, opened 2020-07-09
Create index.js which redirects to src/index.js

algiuxass:
Older versions of node.js won't read package.json main parameter. This will fix it.


rom1504:
Older versions of node are anyway not supported (we use some recent js features), so this is not needed

PR #740, opened 2020-07-14
Move to prismarine-chat

SiebeDW:
Make client_chat example use prismarine-chat


rom1504:
Please run npm run fix to fix the lint

PR #748, opened 2020-08-02
Fallback to Aes-js if no aes-128-cfb8 in crypto

kukelekuuk:
for BoringSSL-based environments such as Electron

I did some limited testing on this, but I'm not entirely certain if it's a good implementation. I'll have to do more extensive testing at a later time.

### DIFF
--- a/src/transforms/encryption.js
+++ b/src/transforms/encryption.js
@@ -1,0 +1,54 @@
+'use strict'
+
+const crypto = require('crypto')
+const { Transform } = require('readable-stream')
+
+if (!crypto.getCiphers().includes('aes-128-cfb8')) {
+  const Cfb = require('aes-js').ModeOfOperation.cfb
+
+  class Cipher extends Transform {
+    constructor (sharedSecret) {
+      super()
+      this.aes = new Cfb(sharedSecret, sharedSecret, 8)
+    }
+
+    _transform (chunk, encoding, callback) {
+      try {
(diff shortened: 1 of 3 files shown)


src/transforms/encryption.js
```diff
@@ -0,0 +1,54 @@
...
+
+  class Cipher extends Transform {
+    constructor (sharedSecret) {
+      super()
+      this.aes = new Cfb(sharedSecret, sharedSecret, 8)
+    }
+
+    _transform (chunk, encoding, callback) {
+      try {
+        callback(null, this.aes.encrypt(Buffer.concat([chunk, Buffer.alloc(chunk.length % 8)])))
```
rom1504, line 17: I don't see how this can work, this should be using a stream inside aes.js

rom1504: and why `chunk.length % 8` ?

rom1504:
maybe you need https://github.com/ricmoo/aes-js/pull/97/files

rom1504:
this didn't work at https://github.com/PrismarineJS/node-minecraft-protocol/pull/735/files 
anything new here @kukelekuuk ?

kukelekuuk:
while I've made it work on a stream on my local version. But for some reason crypto's encrypt and aesjs' encrypt both give me different values. (also aesjs doesn't accept any values that aren't exactly a multiple of segment-size)

I'm not sure what to do with this for now. There doesn't appear to be a decent pure js implementation of cfb8 anywhere else. I'll close it and leave it alone for now.

Issue #743, opened 2020-07-24
Electron support


rom1504:
Electron is missing the cipher we use because it doesn't use openssl

rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol/pull/735 related

kukelekuuk:
Was just about to make an issue about this. A fallback implementation with aes-js or another library if the client doesn't contain the aes-128-cfb8 cipher would be preferred, personally.

rom1504:
Done

PR #749, opened 2020-08-02
Cipher Fallback using crypto-js

timonschiffmann:
This PR proposes a fallback to the native nodejs crypto module.
This is usefull because some environments like electron/chromium have disabled the default module(OpenSSL) in favor of boringSSL.

While inspired by #735, I tried to implement a solution with crypto-js and cryptojs-extension because I  could not reproduce the same behaviour with aes-js.

The current code does not work and should probably be changed a lot. I have included a file named encryption-fallback-test.js that i used to test my solution. To use this file just change the first expression in encryption.js to true or false and see if the output changes.

I dont know what the current issue is but this still does not work with mineflayer. 
I also do not have the knowledge on how to debug this so any help would be apprechiated.

Minecraft v1.16.1
Electron v9.1.2
Mineflayer v2.25.0

#748 and #743 are also related.

### DIFF
--- a/src/transforms/encryption-fallback-test.js
+++ b/src/transforms/encryption-fallback-test.js
@@ -1,0 +1,91 @@
+const encryption = require('./encryption');
+
+/*
+
+Test for crypto-js fallback
+Change between fallback and nodejs crypto by editing the expression in encryption.js to true or false
+
+########### Nodejs crypto
+
+input     TextToEncryptAndThenDecrypt 27
+encrypted 608e4df02a82de4a03c547e8f8ac3f924a6e962a43e9242ad5eaa5 27
+decrypted TextToEncryptAndThenDecrypt 27
+
+input     ThisIsTheSecondMessageForTheSameStream 38
+encrypted 3379188d66a141f184b5175991fe8c7cfe7b20c7a4b7eb9382f4b645d0cbf5df4f84e298a880 38
+decrypted ThisIsTheSecondMessageForTheSameStream 38
+
+
+########### crypto-js
+
+input     TextToEncryptAndThenDecrypt 27
+encrypted 608e4df02a82de4a03c547e8f8ac3f924a6e962a43e9242ad5eaa5 27
+decrypted TextToEncryptAndThenDecrypt 27
+
+input     ThisIsTheSecondMessageForTheSameStream 38
+encrypted 3379188d66a141f184b5175991fe8c7cfe7b20c7a4b7eb9382f4b645d0cbf5df4f84e298a880 38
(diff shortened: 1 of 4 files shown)


rom1504:
What does work ? Did you identify any difference with your stream test ?

src/transforms/encryption-fallback-test.js
```diff
@@ -0,0 +1,91 @@
+const encryption = require('./encryption');
```
rom1504, line 1: Problem with this example is your are not testing the stream part, ie the crypted bytes can be cut anywhere

src/transforms/encryption.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+
+        _transform(chunk, enc, cb) {
+            try {
+                //console.log("enc:" + enc);
+                //console.log("c-in", chunk)
+
+                let encrypted = CryptoJS.AES.encrypt(
+                    CryptoJS.enc.Hex.parse(chunk.toString('hex')),
```
rom1504, line 90: Going through hex string is highly inefficient

src/transforms/encryption.js
```diff
@@ -0,0 +1,149 @@
...
+                        iv: CryptoJS.enc.Hex.parse(this.iv.toString('hex')),
+                        mode: CryptoJS.mode.CFBb,
+                        segmentSize: 8,
+                        padding: CryptoJS.pad.NoPadding
+                    }
+                );
+
+                let resbuf = Buffer.from(this.wordArrayToByteArray(decrypted));
+                //console.log("dc-out", resbuf)
+                this.iv = chunk.slice(chunk.length - 16, chunk.length);
```
rom1504, line 133: Not convinced this stream can be parsed tcp packet by tcp packet. I encourage you look this up and maybe look at the code of the crypto internal package for that

rom1504:
https://github.com/nodejs/node/blob/1d7be3253f12c9eafaf0ecec378a657893b30852/lib/crypto.js for reference

rom1504:
https://github.com/nodejs/node/blob/1d7be3253f12c9eafaf0ecec378a657893b30852/lib/internal/crypto/cipher.js

rom1504:
https://github.com/nodejs/node/blob/1d7be3253f12c9eafaf0ecec378a657893b30852/lib/internal/crypto/cipher.js#L134

rom1504:
basically something like `update` must be used to maintain the state

rom1504:
https://cryptojs.gitbook.io/docs/

rom1504:
@IdanHo managed to make it work with aes-js in https://github.com/PrismarineJS/node-minecraft-protocol/pull/750 ; I think we can close this

PR #750, opened 2020-08-02
Implement fallback aes-js support for cfb8

IdanHo:
Tested with a flying squid online mode server, works fine.
The 2 lines with disable eslint are due to standard complaining about aes-js's naming convention for classes, which we cannot change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "standard": "^14.3.4"
   },
   "dependencies": {
+    "aes-js": "^3.1.2",
     "yggdrasil": "^1.3.0",
     "buffer-equal": "^1.0.0",
     "debug": "^4.1.0",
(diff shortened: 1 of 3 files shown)


rom1504:
can you also try to connect to a vanilla server using a node-minecraft-protocol client ?

Heath123:
I'm not an expert or anything, maybe I shouldn't be sort of reviewing it but can't you disable specific rules instead of the whole line? And maybe add a comment to explain it?

IdanHo:
> can't you disable specific rules instead of the whole line?
you're right, disabled the specific rule.

rom1504:
also tested with master and confirm this didn't work with electron
I'm going to add a simple example with electron

rom1504:
thanks @IdanHo

Issue #718, opened 2020-05-20
Automatically dump lot of packets to improve benchmark and testing


rom1504:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%247%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/PrismarineJS/node-minecraft-protocol/issues/718)

similar to https://github.com/PrismarineJS/minecraft-chunk-dumper

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $7.00)

- [<img src='https://avatars.githubusercontent.com/u/2346494?v=4' alt='rom1504' width=24 height=24> rom1504](https://issuehunt.io/u/rom1504) ($2.00)
- $5.00 have been anonymously funded.
### Submitted pull Requests
- [#842 Add packet cycletests](https://issuehunt.io/r/PrismarineJS/node-minecraft-protocol/pull/842)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/PrismarineJS/node-minecraft-protocol/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

rom1504:
in progress at https://github.com/PrismarineJS/prismarine-packet-dumper

rom1504:
Idea here :
* Dump a reasonable number of packets per kind (4 maybe) : the raw buffer
* Store that somewhere convenient (some minecraft-protocol-stub repo maybe)
* Build tests for node-minecraft-protocol using that : automatically parse then serialise the packets, check it matches

This can be done first for vanilla, then for some other servers.

This is a good way to fix all the "packet is broken" issue

Heath123:
Would running this on a test server help with packet collection?

rom1504:
Yes but the important thing is to implement the code to do all this,
finding server is easy.

rom1504:
see https://github.com/PrismarineJS/prismarine-packet-dumper/issues to know what to do

rom1504:
done in https://github.com/PrismarineJS/node-minecraft-protocol/pull/842 ; thanks @u9g 
either take the money in issuehunt of reinvest it in other issues ;)

PR #757, opened 2020-08-16
make 1.16.2 the tested version instead of 1.16.1

rom1504:


rom1504:
I'm betting on something wrong in https://github.com/ProtoDef-io/node-protodef/blob/master/src/datatypes/compiler-utils.js#L37
maybe you could have a quick look @Karang to see if you see anything obvious ? 
Problematic bitfield is there https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.16.2/protocol.json#L1587

rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol/issues/718 could help simplify this procedure and confidence in it

Karang:
> I'm betting on something wrong in https://github.com/ProtoDef-io/node-protodef/blob/master/src/datatypes/compiler-utils.js#L37
> maybe you could have a quick look @Karang to see if you see anything obvious ?
> Problematic bitfield is there https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.16.2/protocol.json#L1587

Nothing obvious, I'll run the test manually later.

Karang:
Looks like the problem is in the generated test packet (p1), it doesn't match the schema (x, y, z gets generated inside and outside the bitfield):
```
p1: {
  "x": 1,
  "z": 1,
  "y": 1,
  "chunkCoordinates": {
    "x": 1,
    "z": 1,
    "y": 1
  },
  "notTrustEdges": true,
  "records": [
    1
  ]
}
p2: {
  "chunkCoordinates": {
    "x": 1,
    "z": 1,
    "y": 1
  },
  "notTrustEdges": true,
  "records": [
    1
  ]
}

1) packets 1.16.2
       play,ClientBound,packet_multi_block_change:

      Uncaught AssertionError [ERR_ASSERTION]: field x missing in p2, in p1 it has value 1
```

rom1504:
I don't remember, I just removed this line.
These packets tests should be generalized and cleaned up I think https://github.com/ProtoDef-io/node-protodef/issues/83

rom1504:
mc-server 1.16.2 "clients can log in and chat" failing now apparently

rom1504:
most likely related to packet changes, I'll list them here so it's easier to update mineflayer and flying squid too

rom1504:
reference https://github.com/PrismarineJS/minecraft-data/pull/326

rom1504:
worked a bit on this. Main thing that is annoying is that login packet. The server is now expected to provide a big list of stuff each biome should be doing.
In 1.16.1 they introduced having to send a dimension codec which was already big, but now it's really way too big to put in the example.
I think we should save the default value in a json in minecraft-data. and then it would be used here and in flying-squid in one line.
Will do that another time

GroobleDierne:
> 
> 
> worked a bit on this. Main thing that is annoying is that login packet. The server is now expected to provide a big list of stuff each biome should be doing.
> In 1.16.1 they introduced having to send a dimension codec which was already big, but now it's really way too big to put in the example.
> I think we should save the default value in a json in minecraft-data. and then it would be used here and in flying-squid in one line.
> Will do that another time

When you say `I think we should save the default value in a json in minecraft-data`, are you talking about the whole login packet or just the dimension codec field ?

rom1504:
yes
in the previous versions it was reasonable to store this directly in the example as it was only tens of lines, now it's thousands of lines, it's not reasonable anymore, which is why I'm suggesting to put it in mc data

GroobleDierne:
I am interested to do it, is there an existing file for these data?

rom1504:
no but

> As mentioned in the Discord server, the approach to generating this data would be to log into a 1.16.2 server using mc-protocol and basically save all data from the login packet. Then simply extract the biome data elements into MC-Data.

GroobleDierne:
I meant in MCData

rom1504:
same answer

GroobleDierne:
ok

rom1504:
@GroobleDierne released your file as mc data 2.69.0, you can now use it as .loginPacket here

GroobleDierne:
Thanks. I will do a PR soon for NMP

rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol/pull/765

rom1504:
closing in favor of #765

PR #759, opened 2020-08-30
Add IPv6 support by default

Saiv46:


rom1504 (changes requested):
If the purpose of this change is to set host to undefined by default, I'd better you put host = undefined explicitly even if it means the same

rom1504:
And can you point to the doc saying that undefined means "listen to all
ipv4 and ipv6 adresses" ? That seems unlikely to me

rom1504:
Still better to do host = undefined to make it explicit it's optional

rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol/blob/8e4978f3793504435d1b8d7235173f2075ce94db/src/server.js#L53
host option goes there fyi

rom1504:
And update the doc

rom1504:
https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback
indeed seems like you right, undefined means ipv6+ipv4 on all addresses

src/createServer.js
```diff
@@ -13,7 +13,7 @@ module.exports = createServer
 
 function createServer (options = {}) {
   const {
-    host = '0.0.0.0',
+    host,
```
wvffle, line 16: This looks good to me. I don't see the point of doing `host = undefined`. It is redundant in my opinion.

wvffle (approved):

rom1504:
It's not redundant. That marks the parameter as optional.

Saiv46:
Where? There are no `.d.ts` files to mark this parameter as optional.

rom1504:
Just put host = undefined in the function definition

rom1504:
And update the doc to explain the new behavior by default for this parameter

rom1504:
would be good to finish this

Saiv46:
Any progress? `host = undefined` is useless, and `host` parameter just not documented

rom1504:
Yes I'm waiting for you to make the changes.
host = undefined is a clear indication for the reader of the code that is parameter is optional.
host parameter is documented in the API.md file

rom1504:
I don't really understand what's blocking you here. Seems difficult for you to accept code reviews?
I can do the changes myself, but it seems like a good learning opportunity.

PR #764, opened 2020-09-24
Server Brand

GroobleDierne:
Added a new function, `Client#sendBrand(string)`, this function will send the brand plugin message. It only works for server side clients `client.isServer === true`.
Added a new optional parameter to `createServer(options)` called `serverBrand` which is used as default value for `sendBrand()` if no parameter is given and in the ping response, the default value is `node-minecraft-protocol`.

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -21,7 +21,8 @@ function createServer (options = {}) {
     maxPlayers: maxPlayersNew = 20,
     version,
     favicon,
-    customPackets
+    customPackets,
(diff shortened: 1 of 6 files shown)


rom1504 (changes requested):
hi,
This doesn't belong in node-minecraft-protocol imo but in users of it

GroobleDierne:
In flying-squid instead ?

rom1504:
yes

PR #765, opened 2020-10-04
1.16x

### LINKED ISSUE #762: Support for Minecraft 1.16.3
I'm using the library for `mc.ping` and the request fails when I set the server version to `1.16.3`.

GroobleDierne:
Support for 1.16.2 & 1.16.3
Fix #762


docs/README.md
```diff
@@ -89,78 +89,17 @@ var server = mc.createServer({
...
-    dimension: 'minecraft:overworld',
-    worldName: 'minecraft:overworld',
-    difficulty: 2,
-    hashedSeed: [0, 0],
-    maxPlayers: server.maxPlayers,
-    reducedDebugInfo: false,
-    enableRespawnScreen: true
-  });
+  
+  let loginPacket = mcData.loginPacket
```
rom1504, line 98: Do it the other way round: specify all fields below and extract them all from loginPacket

test/serverTest.js
```diff
@@ -201,21 +201,28 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
...
         })
-        client.write('login', {
-          entityId: client.id,
+        let loginPacket = {
+          entityId: 0,
           levelType: 'default',
           gameMode: 1,
           previousGameMode: 255,
           worldNames: ['minecraft:overworld'],
           dimensionCodec: { name: '', type: 'compound', value: { dimension: { type: 'list', value: { type: 'compound', value: [w] } } } },
```
rom1504, line 210: No reason to keep this line

test/serverTest.js
```diff
@@ -317,21 +324,28 @@ mc.supportedVersions.forEach(function (supportedVersion, i) {
           assert.strictEqual(reason, '{"text":"ServerShutdown"}')
           resolve()
         })
-        client.write('login', {
-          entityId: client.id,
+        let loginPacket = {
+          entityId: 0,
```
rom1504, line 328: Would be cleaner hers to do the same, ie dimensionCodec: loginPacket.dimensionCodec

rom1504 (changes requested):
In addition to the comment I left, the login packet change must happen also in all server examples
And you also need the change I started in the other PR

And you'll also need to change the way map_chunk is sent in examples

test/serverTest.js
GroobleDierne: If I do, the code will look like that
```javascript
let loginPacket = {
  entityId: client.id,
  levelType: 'default',
  gameMode: 1,
  previousGameMode: 255,
  worldNames: ['minecraft:overworld'],
  dimensionCodec: { name: '', type: 'compound', value: { dimension: { type: 'list', value: { type: 'compound', value: [w] } } } },
  dimension: 0,
  worldName: 'minecraft:overworld',
  hashedSeed: [0, 0],
  difficulty: 2,
  maxPlayers: server.maxPlayers,
  reducedDebugInfo: 0,
  enableRespawnScreen: true
}
if (version.version >= 735) { // 1.16x
  loginPacket.dimensionCodec = mcData.loginPacket.dimensionCodec
  loginPacket.dimension = mcData.loginPacket.dimension
  loginPacket.isDebug = false
  loginPacket.isFlat = false
  loginPacket.isHardcore = false
  loginPacket.reducedDebugInfo = false
  loginPacket.viewDistance = 10
  delete loginPacket.levelType
  delete loginPacket.difficulty
}
```
It's much more verbose

rom1504: yes, verbose is not a problem. It's important to be explicit in an example
and this can be simplified a lot by doing ternary operations in each of the fields necessary

plus again `dimensionCodec: { name: '', type: 'compound', value: { dimension: { type:` this should always be taken from mcdata now

rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol/blob/741f9c35ed83c7040952c51af41ff444f7392e39/test/serverTest.js#L240 need to update this line

GroobleDierne:
> 
> 
> https://github.com/PrismarineJS/node-minecraft-protocol/blob/741f9c35ed83c7040952c51af41ff444f7392e39/test/serverTest.js#L240
> need to update this line

What's the problem with this line ?

GroobleDierne:
I will continue another day

rom1504:
> > https://github.com/PrismarineJS/node-minecraft-protocol/blob/741f9c35ed83c7040952c51af41ff444f7392e39/test/serverTest.js#L240
> > 
> > need to update this line
> 
> What's the problem with this line ?

gamemode is not a number anymore

GroobleDierne:
It looks like it is https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.16.2/loginPacket.json#L4

https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.16.2/protocol.json#L2095

rom1504:
Right ok

GroobleDierne:
what is the reason for this condition ?

https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/test/clientTest.js#L156

rom1504:
This test is checking whether the incorrect version error is emitted in case of incorrect version.
For that it sets the version to 1.8.8 for all tested versions except for 1.8.8. for 1.8.8 it sets 1.11.2, so it's different from 1.8.8

GroobleDierne:
So it tests if the server disconnects the client with a decent error message, but it tests it on a vanilla wrapped server ?

rom1504:
it tests that the nmp client emit the right error when the wrapped vanilla server kicks it for incorrect version yes

rom1504:
looks quite reasonable to me, do you think it's ready @GroobleDierne ?

rom1504:
the next step will be doing these adaptation in mineflayer and flying-squid https://github.com/PrismarineJS/node-minecraft-protocol/pull/757#issuecomment-678520061
Would be great if you could also work on this part @GroobleDierne . what do you think ?

rom1504 (approved):

GroobleDierne:
> if you could also work on this part @GroobleDierne . what do you think ?

No problem for me

PR #771, opened 2020-10-25
client.end with full reason

redcarti:


rom1504:
can you also add it in the doc there https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/API.md#clientendreason ?

redcarti:
oh, sorry, forgot about docs

rom1504 (approved):

PR #775, opened 2020-11-05
End the serializer instead of unpiping streams

DeltaEvo:
This will let all the streams end their work before
closing the socket

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -169,10 +169,14 @@ class Client extends EventEmitter {
 
   end (reason) {
     this._endReason = reason
-    if (this.cipher) this.cipher.unpipe()
-    if (this.framer) this.framer.unpipe()
+    /* ending the serializer will end the whole chain
+    serializer -> framer -> socket -> splitter -> deserializer */
+    if (this.serializer) {
+      this.serializer.end()
+    } else {
+      if (this.socket) this.socket.end()
+    }
     if (this.socket) {
-      this.socket.end()
       this.closeTimer = setTimeout(
         this.socket.destroy.bind(this.socket),
         closeTimeout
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -169,10 +169,13 @@ class Client extends EventEmitter {
 
   end (reason) {
     this._endReason = reason
-    if (this.cipher) this.cipher.unpipe()
-    if (this.framer) this.framer.unpipe()
+    /* ending the serializer will end the whole chain
+    serializer -> framer -> socket -> splitter -> deserializer */
+    if (this.serializer) this.serializer.end()
```
wvffle, line 174: Use brackets here

wvffle (changes requested):

DeltaEvo:
gentle ping @wvffle

rom1504:
With this change, the socket would never be closed from node-minecraft-protocol side (both as a server and as a client).
The other party might not close the connection.
This is a breaking change as called of client.end expect the connection to be closed now.
Introducing a new method for this use case might be better.
What do you think ?

DeltaEvo:
> With this change, the socket would never be closed from node-minecraft-protocol side (both as a server and as a client).

No the socket will be still closed but not right away https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options pipe have an `end` option that is by default to true to according to the doc `End the writer when the reader ends. Default: true.`
In practice this means that the writer will be ended when the reader will be finished
I added a comment that explain this `ending the serializer will end the whole chain: serializer -> framer -> socket -> splitter -> deserializer`

ending the serializer will cause the socket to end but not right away it will let the serializer and the framer finish what it was doing before

When we ended the socket and unpiped the stream if the serializer still had packet to serialize they were lost since no one was consuming the stream

And if the socket takes times to end there is still the closeTimeout that will destroy the socket and so force the close if it

DeltaEvo:
And about the code inside end I kept the check for the socket `if (this.socket)` but is there a case in this code where the socket is not defined ? same for the serializer ?

rom1504:
> pipe have an end option that is by default to true to according to the doc End the writer when the reader ends. Default: true.
the serializer is a writer, the socket is a reader
so the writer is ending, that doesn't make the reader end

DeltaEvo:
> the serializer is a writer, the socket is a reader

It's a Duplex stream so it's Readable and Writable and the Readable part is piped inside the Writable part of the framer. The framer is also a Duplex Stream and the Readable part of it is piped into the Writable part of the socket

rom1504 (approved):

Issue #711, opened 2020-05-09
improve nmp perf by using a protodef compiler


rom1504:
see https://github.com/ProtoDef-io/node-protodef/issues/86

rom1504:
this is now done, thanks @Karang

PR #776, opened 2020-11-05
Only use fullReason for the disconnect packet

DeltaEvo:


wvffle (approved):
LGTM though I don't quite understand why the tests changed now.

DeltaEvo:
I wasn't expecting the tests to change when I implemented this, but the tests check the end reason to check the kick reason
Before my change the endReason was transformed into Chat and then used as client end reason and as disconnect packet reason
Now the endReason is only transformed for the disconnect packet so the client end reason is not a JSON string anymore

rom1504:
What is this pr supposed to fix/improve ?

DeltaEvo:
> What is this pr supposed to fix/improve ?
Oh yeah forgot the most important part, explaining what I want to improve x)

Since the merge of #711 there is now a `fullReason` that can be used to use minecraft formated chat as the disconnection reason, the json object to format chat can become quickly big with formatting, so this make the Client end reason very difficult to read

`fullReason` also discard the first `reason` argument of the `end` function, so instead of using a minecraft formated json as client end reason I thinked using the first string reason parameter permit us to have a more readable reason since the `end` reason is mostly used for debugging

rom1504 (approved):

PR #777, opened 2020-11-10
Fix incorrect protocol message

Heath123:
I didn't really make this, so I set it as co-authored
Sets the state to login before sending the message so it gets interpreted properly by the client

### DIFF
--- a/src/server/handshake.js
+++ b/src/server/handshake.js
@@ -8,6 +8,12 @@ module.exports = function (client, server, { version }) {
     client.serverPort = packet.serverPort
     client.protocolVersion = packet.protocolVersion
 
+    if (packet.nextState === 1) {
+      client.state = states.STATUS
+    } else if (packet.nextState === 2) {
+      client.state = states.LOGIN
+    }
+
     if (version === false || version === undefined) {
       if (require('minecraft-data')(client.protocolVersion)) {
         client.version = client.protocolVersion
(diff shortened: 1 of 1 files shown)


src/server/handshake.js
```diff
@@ -8,6 +8,12 @@ module.exports = function (client, server, { version }) {
     client.serverPort = packet.serverPort
     client.protocolVersion = packet.protocolVersion
 
+    if (packet.nextState === 1) {
+      client.state = states.STATUS
+    } else if (packet.nextState === 2) {
+      client.state = states.LOGIN
+    }
+
```
DeltaEvo, line 16: You need to assign `client.version` before `client.state` since `version` is used to build the serializer

DeltaEvo (changes requested):

src/server/handshake.js
DeltaEvo: Or reassign client.state to rebuild the serializer :thinking:

What do you think @rom1504 ?

rom1504:
Does this really work? It seems to me it won't work due to lot of things being done when the state is changed

Heath123:
> Does this really work? It seems to me it won't work due to lot of things being done when the state is changed

I need to update it but I haven't gotten around to it yet

rom1504:
stale, reopen if you want to do it

PR #780, opened 2020-11-30
Add token authentication

### LINKED ISSUE #519: add an option to store and use auth token directly in nmp
[Issuehunt badges](https://issuehunt.io/r/PrismarineJS/node-minecraft-protocol/issues/519)

see https://github.com/pqml/prismarine-tokens and https://github.com/PrismarineJS/node-minecraft-wrap/blob/master/lib/wrap_client.js#L37

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $2.00)

- [rom1504 rom1504](https://issuehunt.io/u/rom1504) ($2.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/PrismarineJS/node-minecraft-protocol/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

ph0t0shop:
(this assumes https://github.com/PrismarineJS/node-yggdrasil/pull/51)
Fixes #519. It allows the user to specify a `profilesFolder` property. If given, a `launcher_profiles.json` file will be read or created if it doesn't yet exist, and it will then act approximately like the minecraft launcher. The given username is searched within the launcher profiles and if found, it attempts to use this information to login. If not found, or the login fails, it can optionally use a password if it's given. If not, it will fail. On any successful login, if the `profilesFolder` property is given, it will save it properly to the file. This works for preexisting and non-preexisting profiles and is compatible with the official launcher. It will only save it if the clientToken matches so as to not corrupt the existing `launcher_profiles.json`. This will match by default, and might only mismatch if another clientToken or session with clientToken is passed as an options property.


src/client/auth.js
```diff
@@ -1,16 +1,67 @@
 const UUID = require('uuid-1345')
 const yggdrasil = require('yggdrasil')
+const fs = require('mz/fs')
```
DeltaEvo, line 3: The node `fs` api have a promisified version https://nodejs.org/api/fs.html#fs_fs_promises_api
You can use it with `require('fs/promises')` or ` require('fs').promises`

src/client/auth.js
```diff
@@ -1,16 +1,67 @@
...
+const promisify = require('es6-promisify')
```
DeltaEvo, line 4: There is already a promisify function in node https://nodejs.org/api/util.html#util_util_promisify_original

src/client/auth.js
```diff
@@ -1,16 +1,67 @@
...
+const mkdirp = promisify(require('mkdirp'))
```
DeltaEvo, line 5: Node `mkdir` function now have a `recursive` option that does the same https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback
And `mkdirp` already return a promise https://www.npmjs.com/package/mkdirp#mkdirpdir-opts---promisestring--undefined

src/client/auth.js
```diff
@@ -1,16 +1,67 @@
...
+module.exports = async function (client, options) {
   const yggdrasilClient = yggdrasil({ agent: options.agent, host: options.authServer || 'https://authserver.mojang.com' })
-  const clientToken = options.clientToken || (options.session && options.session.clientToken) || UUID.v4().toString()
+  const clientToken = (options.profilesFolder && (await getLP()).clientToken) || options.clientToken || (options.session && options.session.clientToken) || UUID.v4().toString().replace("-", "")
   const skipValidation = false || options.skipValidation
   options.accessToken = null
-  options.haveCredentials = options.password != null || (clientToken != null && options.session != null)
+  options.haveCredentials = options.password != null || (clientToken != null && options.session != null) || options.profilesFolder != null
+
+  async function getLP () { // get launcher profile
```
DeltaEvo, line 14: `getLP` is a little short why not the full `getLauncherProfile` ?

src/client/auth.js
```diff
@@ -1,16 +1,67 @@
...
+    let data
+    try {
+      data = JSON.parse(await fs.readFile(options.profilesFolder + '/launcher_profiles.json', 'utf8'))
+    } catch (err) {
+      await mkdirp(options.profilesFolder)
+      await fs.writeFile(options.profilesFolder + '/launcher_profiles.json', '')
+      data = { authenticationDatabase: {} }
+    }
+    return data
```
DeltaEvo, line 23:
```suggestion
    try {
      return JSON.parse(await fs.readFile(options.profilesFolder + '/launcher_profiles.json', 'utf8'))
    } catch (err) {
      await mkdirp(options.profilesFolder)
      await fs.writeFile(options.profilesFolder + '/launcher_profiles.json', '')
      return { authenticationDatabase: {} }
    }
```

src/client/auth.js
```diff
@@ -1,16 +1,67 @@
...
-  options.haveCredentials = options.password != null || (clientToken != null && options.session != null)
+  options.haveCredentials = options.password != null || (clientToken != null && options.session != null) || options.profilesFolder != null
+
+  async function getLP () { // get launcher profile
+    let data
+    try {
+      data = JSON.parse(await fs.readFile(options.profilesFolder + '/launcher_profiles.json', 'utf8'))
+    } catch (err) {
+      await mkdirp(options.profilesFolder)
+      await fs.writeFile(options.profilesFolder + '/launcher_profiles.json', '')
```
DeltaEvo, line 20: Is it a good idea to create an empty json file ?

src/client/auth.js
```diff
@@ -1,16 +1,67 @@
...
+      data = { authenticationDatabase: {} }
+    }
+    return data
+  }
 
   if (options.haveCredentials) {
     // make a request to get the case-correct username before connecting.
-    const cb = function (err, session) {
+    const cb = async function (err, session) {
```
DeltaEvo, line 28: I don't like the idea to have an async callback function since this means that the error will be unhandled

src/client/auth.js
```diff
@@ -1,16 +1,67 @@
 const UUID = require('uuid-1345')
 const yggdrasil = require('yggdrasil')
+const fs = require('mz/fs')
+const promisify = require('es6-promisify')
+const mkdirp = promisify(require('mkdirp'))
 
-module.exports = function (client, options) {
+module.exports = async function (client, options) {
```
DeltaEvo, line 7: If this is a async function maybe we should handle the error ?

src/client/auth.js
```diff
@@ -22,6 +73,34 @@ module.exports = function (client, options) {
...
+          id: uuid
+        };
+
+        options.session = {
+          accessToken: auths.authenticationDatabase[profile].accessToken,
+          clientToken: auths.clientToken,
+          selectedProfile: newProfile,
+          availableProfiles: [newProfile]
+        }
+        options.username = newUsername
```
DeltaEvo, line 100: Can we not set `client.session` and `client.username` instead like in the `cb` used in yggdrasilClient.validate ?

src/client/auth.js
```diff
@@ -1,16 +1,67 @@
...
+    const cb = async function (err, session) {
+      if (options.profilesFolder) {
+        let auths = await getLP()
+        let lowerUsername = options.username.toLowerCase()
+        let profile = Object.keys(auths.authenticationDatabase).find(key => 
+          auths.authenticationDatabase[key].username.toLowerCase() == lowerUsername
+          || Object.values(auths.authenticationDatabase[key].profiles)[0].displayName.toLowerCase() == lowerUsername
+        )
+        if (err) {
+          if (profile) { // profile is invalid, remove
```
DeltaEvo, line 37: Are we sure that the profile is invalid here ? We should check the error since a lot of kind of error can happen here

src/client/auth.js
```diff
@@ -1,16 +1,67 @@
...
   if (options.haveCredentials) {
     // make a request to get the case-correct username before connecting.
-    const cb = function (err, session) {
+    const cb = async function (err, session) {
+      if (options.profilesFolder) {
+        let auths = await getLP()
+        let lowerUsername = options.username.toLowerCase()
+        let profile = Object.keys(auths.authenticationDatabase).find(key => 
+          auths.authenticationDatabase[key].username.toLowerCase() == lowerUsername
+          || Object.values(auths.authenticationDatabase[key].profiles)[0].displayName.toLowerCase() == lowerUsername
```
DeltaEvo, line 34: `Object.values(auths.authenticationDatabase[key].profiles)[0]` why it's only the value of the first key here ?

DeltaEvo (changes requested):
Nice feature !
In your code a lot of let can be transformed to const
And I don't think we need theses extra dependencies

src/client/auth.js
ph0t0shop: Ah okay, this was just mirrored from the minecraft wrapper module, which also used mz

src/client/auth.js
ph0t0shop: Alright. Again, I just mirrored this behavior from the minecraft wrapper module but it should be fine either way

src/client/auth.js
ph0t0shop: Sure, should be fine as well

src/client/auth.js
ph0t0shop: I think so, yeah. It was something suggested in the issue that if it doesn't exist, it creates it. Otherwise this behavior can be easily changed of course

src/client/auth.js
ph0t0shop: It's gonna be hard not to have it, or we'll have to use readFileSync and writeFileSync in getLauncherProfiles

src/client/auth.js
DeltaEvo: I mean more, maybe create it with an empty object inside like `{}` because an empty file is not a valid json file

src/client/auth.js
ph0t0shop: No because we don't want code duplication and we'll load these efforts off to the rest of the code that also handles refreshing the token.

src/client/auth.js
ph0t0shop: I think there's only ever 1 profile for the minecraft launcher

ph0t0shop: Per account, that is. You can still have multiple accounts but for now those all have 1 profile.

src/client/auth.js
DeltaEvo: The problem is not having it but more that the way that it's used here if you have an error in this async function no one is listening to it, I don't think using readFileSync and writeFileSync is the good solution here, maybe just extra the async function elsewhere and use the `.then` on it and handle the error

src/client/auth.js
DeltaEvo: Oh ok :smile:

src/client/auth.js
DeltaEvo: :+1:

ph0t0shop:
@DeltaEvo I think I fixed most of the stuff you mentioned, any thoughts?

src/client/auth.js
```diff
@@ -1,60 +1,146 @@
...
-    const cb = function (err, session) {
-      if (err) {
-        client.emit('error', err)
-      } else {
-        client.session = session
-        client.username = session.selectedProfile.name
-        options.accessToken = session.accessToken
-        client.emit('session', session)
-        options.connect(client)
+  (async function() {
```
wvffle, line 6: Can't you just make the exported function async?

src/client/auth.js
```diff
@@ -1,60 +1,146 @@
...
+      client.username = options.username
+      options.connect(client)
     }
-  } else {
-    // assume the server is in offline mode and just go for it.
-    client.username = options.username
-    options.connect(client)
-  }
+  })().then((res) => {}, (err) => {
+    console.warn("Authentication failed: \n", err)
```
wvffle, line 144: Printing error to the console is not a good approach. For instance it would break layout of mineflayer-dashboard.

src/client/auth.js
```diff
@@ -1,60 +1,146 @@
...
+                  displayName: session.selectedProfile.name
+                }
+                auths.authenticationDatabase[profile] = newProfileObj
+              }
+            }
+            fs.writeFile(options.profilesFolder + '/launcher_profiles.json', JSON.stringify(auths, null, 2)).then(()=>{},(err)=>{
+              console.warn("Couldn't save tokens:\n", err)
             })
+          }, (err) => {
+            console.warn("Skipped saving tokens because of error\n", err)
```
wvffle, line 63: Same as request on L144

wvffle (changes requested):

src/client/auth.js
DeltaEvo: See https://github.com/PrismarineJS/node-minecraft-protocol/pull/780#discussion_r533308810
I think we can make this async but for me we need to handle the promise rejection somewhere, so if this is made async I think we should handle the error in createClient.js

DeltaEvo:
Yep it seems better, only

> In your code a lot of let can be transformed to const

is still valid :)

DeltaEvo (approved):

src/client/auth.js
ph0t0shop: It was like this before, I wasn't sure if that was okay

wvffle:
You should resolve things that you've already done

src/client/auth.js
ph0t0shop: What should be done instead?

src/client/auth.js
ph0t0shop: I'm not sure, I made a nested async function instead to handle the error inside the function, but what is the preferred way to handle errors if not logging to console?

src/client/auth.js
ph0t0shop: Good point

src/client/auth.js
wvffle: Of course we have to handle it. 

If things haven't changed (I've been busy lately), we're using callbacks in pretty much every project so another option is to transform this async function into a callback one at some point.

The rejection can also emit an error of some sort.

ph0t0shop:
Okay I resolved most, the remaining ones have to do with error handling. That's mainly because I'm not sure what the preferred way to go about that is in this project. If I know this I can properly implement it

ph0t0shop:
And transforming `let` to `const` I guess...

ph0t0shop:
Ok I changed that. I still left it in an async function because I prefer the await operator over ugly .then chaining. If you ask me, this pr is ready to merge now

ph0t0shop:
Welp, idk why the java env setup is failling in the tests, but lmk what needs to be done to get this merged. Seems like it would be a useful feature, and it's also on the "big prismarine projects" list I think. If you don't want it just tell me and I'll close this

Gjum:
The java env setup seems to be failling because of [a recent (Oct. 1) security vulnerability fix](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) on Github's CI. It's not really in-scope of this PR, so I've attempted a fix in #785.

src/client/auth.js
```diff
@@ -1,16 +1,68 @@
...
   options.accessToken = null
-  options.haveCredentials = options.password != null || (clientToken != null && options.session != null)
+  options.haveCredentials = options.password != null || (clientToken != null && options.session != null) || options.profilesFolder != null
+
+  async function getLauncherProfiles () { // get launcher profiles
+    try {
+      return JSON.parse(await fs.readFile(options.profilesFolder + '/launcher_profiles.json', 'utf8'))
+    } catch (err) {
+      await fs.mkdir(options.profilesFolder, { recursive: true })
+      await fs.writeFile(options.profilesFolder + '/launcher_profiles.json', '{}')
```
Heath123, line 17: Will this mess with the launcher at all?

ph0t0shop: Yes, 'mess' as in it will modify that file. It's compatible with the official launcher, which means it properly deletes _and_ also saves new accounts to the file for the launcher to use

ph0t0shop: Also, one could just specify a path that is not the launcher folder and it will just use its own isolated file instead

ph0t0shop:
Thanks a lot @Gjum, the tests are all passing now

src/client/auth.js
ph0t0shop: I have thought about this, and it should be fine like this IMO. The minecraft launcher itself does the same thing I think, if your login doesn't succeed for whatever reason, it invalidates it

rom1504:
looks good.
Missing:
* [ ] fix tests in https://github.com/PrismarineJS/node-yggdrasil/pull/51
* [ ] add new option in API.md doc
* [ ] why not use the home folder by default so it's enabled by default ? this would remove hundred of opened issues and discord complaints (ie people getting temporarily banned by mojang due to too often auth requests)

Heath123:
I have a report of this error:
```
(node:22920) UnhandledPromiseRejectionWarning: TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at module.exports (C:\Users\*****\Desktop\pakkit\node_modules\minecraft-protocol\src\client\auth.js:81:30)
```
(happens at https://github.com/ph0t0shop/node-minecraft-protocol/blob/master/src/client/auth.js#L81)

ph0t0shop:
> why not use the home folder by default so it's enabled by default ? this would remove hundred of opened issues and discord complaints (ie people getting temporarily banned by mojang due to too often auth requests)
@rom1504 I was a bit hesitant to do this, firstly because then we need to find the minecraft folder on the user's system, and secondly because then how would one specify that they just don't want to save credentials? Also, it might be unexpected for the user that it suddenly uses their local minecraft folder

rom1504:
https://github.com/PrismarineJS/node-minecraft-wrap/blob/master/lib/wrap_client.js#L11 works pretty well to find the path
To specify not to save the user could put this option as false

About it being unexpected that it would use their folder, I think it's actually what most people would expect by default.
We can add a note in the readme that we're doing that and how to disable it.

I think it's really beneficial to enable this by default, as the current behavior is to call the mojang server too many times which is actually not great.

rom1504:
https://github.com/simonmeusel/minecraft-folder-path/blob/master/index.js
we could also just copy these few lines instead of depending on this package

Karang:
What needs to be enabled by default is saving the token, not necessary the credentials. What I would do is:
* if the user specifies mc directory: use it to load/store credentials and token
* by default: save/load the token only in the local or tmp directory

ph0t0shop:
Okay. I will do this in a bit, and also add some error handling for invalid launcher profile files

rom1504:
Hmmm does the vanilla client save the credentials ? I think it doesn't
Saving the token is good but saving the password not so much

rom1504:
Yeah and I see you also don't store the password.
What did you mean by credentials @Karang  ?

Karang:
> Yeah and I see you also don't store the password.
> What did you mean by credentials @Karang ?

I was thinking about username + password but yeah thats probably not a good idea.

ph0t0shop:
Oh yeah you definitely don't wanna do that. Currently, the password can be specified, and it will be used to login only if it can't use the launcher_profiles for the account, and then saves it as well. This IMO is the most intuitive and mimics the launcher behavior pretty pragmatically

examples/client_token_auth/client_token_auth.js
```diff
@@ -0,0 +1,28 @@
...
+
+if (process.argv.length < 3 || process.argv.length > 5) {
+  console.log('Usage : node echo.js <host> <port> [<name>]')
+  process.exit(1)
+}
+
+const client = mc.createClient({
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4] ? process.argv[4] : 'echo'
```
rom1504, line 11: better remove the default to echo, this example doesn't make sense in offline mode

examples/client_token_auth/client_token_auth.js
```diff
@@ -0,0 +1,28 @@
+const mc = require('minecraft-protocol')
+
+if (process.argv.length < 3 || process.argv.length > 5) {
```
rom1504, line 3: and make this process.argv.length !== 4

rom1504 (approved):

rom1504 (approved):

PR #785, opened 2020-12-04
[CI] update to actions/setup-java@v1.4.3

Gjum:
CI seems to be failing because of this recent change https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

see also the CI results on #780 where I first noticed this


rom1504 (approved):

rom1504:
Thanks, I'll have to do that for other repos too

PR #787, opened 2020-12-06
Fix for auth bug that crashes client after token authentication update

ph0t0shop:
So basically, in the token auth update, the function was changed from synchronous to asynchronous. However, there was an expectation that wasn't accounted for. It's debatable if it's good behavior, but mineflayer depends on the auth function being synchronous. This means that client.username is set _synchronously_. This didn't happen after converting to async code. Some entity iterator that has to do with player_info packets then called toLowercase() on a null username, which doesn't work. So, the fix for this is to set a dummy username synchronously in the auth method, and then asynchronously properly authenticate. You might think that this breaks stuff, but it cannot: before, when authenticating with email&password, the username was immediately set to the email address, which obviously isn't a valid in-game username. But it still worked. And only later was it asynchronously updated through the 'session' event. So this approach of setting an empty username synchronously should work fine. I have checked, and `username` is also the only property that it actually expects. This can be verified by checking the bottom of the original function when authenticating in offline mode. The only line is `client.username = options.username` before calling connect. So, this is now set synchronously to meet these expectations.


src/client/auth.js
```diff
@@ -4,7 +4,12 @@ const fs = require('fs').promises
 const mcDefaultFolderPath = require('minecraft-folder-path')
 const path = require('path')
 
-module.exports = async function (client, options) {
+module.exports = function (client, options) {
+  client.username = ''
```
Karang, line 8: Wouldn't it be safer to initialize it to:
```js
    client.username = options.username
```
?

ph0t0shop: I think that would be just as valid, since neither can be trusted to be an actually valid username, but sure we can do this.

examples/client_echo/client_echo.js
```diff
@@ -1,6 +1,6 @@
 const mc = require('minecraft-protocol')
 
-if (process.argv.length !== 4) {
```
rom1504: This is incorrect. Should be not below 3 and not more than 5 like before.
My comment on this on the previous PR was for the example you had added.
(And that you removed apparently ?)
Can you fix this please ?

rom1504 (changes requested):
Can you add a test checking that username is defined sync ?

rom1504:
Does the mineflayer code using this (invalid ?) username make sense ?
If not we could simply change that

ph0t0shop:
Yeah, that might be the better solution. It's undefined behavior since it will be an invalid username unless the game is started in offline mode, which mineflayer does not check for anyway.

Karang:
What I still find surprising is that mineflayer apparently tries to connect (and receive login packet) while the authentication function has not returned. Do we know what happen ?

ph0t0shop:
Yes, that's because this function is now async. But it was sync before, and still then it is doing stuff before the session has succeeded. This is behavior that needs to be fixed in mineflayer I think.

examples/client_echo/client_echo.js
ph0t0shop: Whoops. I misinterpreted your comment previously. I will solve this.

Karang:
> Yes, that's because this function is now async. But it was sync before, and still then it is doing stuff before the session has succeeded. This is behavior that needs to be fixed in mineflayer I think.

Can you point me to where the auth function is called ? I thought it was the auth function itself that was calling client.connect() ?

ph0t0shop:
The auth.js exports function is called in createClient.js. The auth function itself indeed calls client.connect(), so that is fine. Mineflayer is doing stuff with the player list _before_ even having connected

rom1504:
I checked mineflayer code:
* username existence is needed in findPlayer function
* this function is needed when player_info packet is received (it gives the list of players)

In offline mode there is no auth, so I think play state will happen sooner and player_info will be sent earlier

Karang:
What I don't understand is how it can start receiving packets like `player_info` before calling `connect()`

rom1504:
indeed, I'm checking why `bot.username` is undefined

ph0t0shop:
Because it happens async now, that is the issue. We need some way to check `hasCredentials` synchronously

rom1504:
no async is triggering the issue, but I think it just exposed a bug in mineflayer
https://github.com/PrismarineJS/mineflayer/blob/master/index.js#L99 is where username is being set

Karang:
> no async is triggering the issue, but I think it just exposed a bug in mineflayer
> https://github.com/PrismarineJS/mineflayer/blob/master/index.js#L99 is where username is being set

Ah makes more sense now, so this is where the await should be added then.

ph0t0shop:
Probably. So I think this entire PR is actually unnecessary and we need a fix in mineflayer

rom1504:
I'm looking into it, not understanding fully the chain of events, I'll put it here when I do

rom1504:
https://github.com/PrismarineJS/mineflayer/pull/1486 here's the fix
This PR is indeed unneeded, I'll just revert the revert

rom1504:
what was happening FYI:
1. empty user name was set to bot.username (from bot._client.username)
2. session event was never emitted as it's async
3. login listener was propagating bot.username to bot.entity.username
4. player_info listener was trying to use bot.entity.username

After my change:
1. bot.username is set on login event when username is properly defined, with or without a session event
2. player_info listener can use bot.entity.username

this NMP PR just removed the initial sync set of bot._client.username. Mineflayer should not anyway have relied on it as it's incorrect in online mode.
I believe this change also fixed a bug in mineflayer: bot.entity.username used to stay at the previous value in online mode

rom1504:
superseded by https://github.com/PrismarineJS/node-minecraft-protocol/commit/011b6a189af9c21e2bcb0e387a1f8c29e478bf48

Issue #782, opened 2020-12-01
Authenticating with a Microsoft account


viktordanov:
**Is your feature request related to a problem? Please describe.**
Mojang are transitioning to Microsoft authentication and intend to completely switch over to it in the future. I haven't been able to figure out how to use my new Microsoft account with `#createClient()`. 

**Describe the solution you'd like**
It would be great for the package to treat Microsoft and Mojang accounts equal, i.e. supply an email and a password for the `username` and `password` ClientOptions

**Describe alternatives you've considered**
I've attempted to authenticate my Microsoft account with the access- and clientTokens I extracted from my `.minecraft/launcher-accounts.json` file but I'm not sure what username to supply. I've tried using my Microsoft email, my Minecraft username, and my Xbox username (seeing as this one is also saved in `launcher-accounts.json` I figured it was worth a shot) but none has worked.

rom1504:
As of now, minecraft does not use microsoft accounts and mojang did not
provide any technical details on how this will work.
So there is nothing to do, it will not work to try anything now.
Just wait for microsoft to enable this.

rom1504:
https://wiki.vg/Microsoft_Authentication_Scheme now we have some doc, implementation can start

rom1504:
04
 <rbrick>so i had been rev-engineering the MSA/minecraft login for the last few days. saw minidigger released it. however, i can confirm, you can use your own Azure Application instead of MCs
03:05
 <rbrick>https://gist.github.com/rbrick/be8ed86864fc5d77aa6c979053cfc892#file-msa-go-L20  my test code is a bit of a mess, but works

extremeheat:
I don't think there's much else to reverse-engineer, it looks like standard Xbox auth the same way it's done on Bedrock. There's existing libraries as such that do this auth already, and a lot of complex docs from MS if anyone wants to read them. 

You don't need to even auth like the launcher, there's other non-oauth ways to sign in, as is done in https://github.com/minecraft-linux/msa/blob/0dd792abe72ee0526533765a89017255bb394ad5/src/network/server_config.cpp

Oauth based, https://github.com/Sandertv/gophertunnel/tree/master/minecraft/auth

[...]

PR #788, opened 2020-12-11
Integrate Authentication for Microsoft Accounts

Kashalls:
This pull request adds two new npm modules: `@xboxreplay/xboxlive-auth` and `node-fetch`
This integrates logging into Minecraft Servers for Microsoft Accounts. 
In order to log into a server using a microsoft account, you must specify options.auth = 'microsoft';

This fixes: 
https://github.com/PrismarineJS/node-minecraft-protocol/issues/782
https://github.com/PrismarineJS/mineflayer/issues/1480

// todo: clean up code just a little bit more.


src/client/encrypt.js
```diff
@@ -18,7 +18,7 @@ module.exports = function (client, options) {
         client.end()
         return
       }
-      if (options.haveCredentials) {
+      if ((options.auth === 'microsoft' && !options.haveCredentials) || options.haveCredentials) {
```
Karang, line 21: If you set `options.haveCredentials = true` in microsoftAuth.js, you don't need this change, and the PR will be a bit more self-contained.

Kashalls: Well rom said he wanted to switch between the auth endpoints using auth and options.haveCredentials is a Mojang auth thing and auth is a Microsoft thing but if you want me to ditch the auth check here I can.

Karang: `haveCredentials` is not a mojang thing https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client/auth.js#L22
its just a boolean to say if the user provided a password. If you are using a ms account it should have the same definition.

Karang (approved):
LGTM

Kashalls:
> Does it work ? Does it fake the browser ?
> Can you update API.md with information on how to use it ?

It works, valid session and everything. Any requests to minecraftservices.com uses a simple user-agent, requests made to Xbox Live uses this User-Agent: `Mozilla/5.0 (XboxReplay; XboxLiveAuth/3.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36`

2FA is not supported with this method as it causes authentication issues, however this can be used to temporarily support it until a better solution is found.

Kashalls:
> Can you update API.md with information on how to use it ?
I can, do you want me to create an example on how to use it with Microsoft Authentication?

rom1504:
Yes, basically add the auth option in API.md, and a simple example is useful

rom1504:
Thanks, that's great.
Could you also add a small note about the existence of this auth option in https://github.com/PrismarineJS/node-minecraft-protocol#echo-client-example ?
I think that lot of users will be asking about that and we want to make it as visible as possible

rom1504 (approved):

rom1504:
LGTM, we probably want to support the json storage here too, but this can be done in another PR

Kashalls:
> LGTM, we probably want to support the json storage here too, but this can be done in another PR

Afaik, we do not have enough information to support the `launcher_accounts.json` file. We definitely could do the same thing for Mojang Auth though.

rom1504:
What information are we missing ? Can't we store the tokens ?

craftycram:
is there a way to use an already valid access token to authenticate the user?

kashalls:
> Is there a way to use an already valid access token to authenticate the user?
> 
> I have a web dashboard where I'm logging in using Microsoft Oauth with the XboxLive.signin scope. I want to use the access_token from there to create a minecraft client in the backend. That's why just using the already existing token, instead of using https://www.microsoft.com/link would be awesome.
> 
> EDIT: found your discord server and posted the question there as well. not sure if comments on three year old PR still get noticed

Afaik its been done, you just need to provide your own auth factory. Theres been some mentions in the discord on how to do this.

craftycram:
As far as I understood browsing through the code doing the authentication flows normally I would need to fork prismarine-auth, since the flows are defined there right?  
There is no way for me to provide custom flows with minecraft-protocol client options directly, if I got that correctly.
@kashalls

kashalls:
> As far as I understood browsing through the code doing the authentication flows normally I would need to fork prismarine-auth, since the flows are defined there right? There is no way for me to provide custom flows with minecraft-protocol client options directly, if I got that correctly. @kashalls

https://github.com/PrismarineJS/prismarine-auth/blob/master/src/MicrosoftAuthFlow.js#L54

The auth flow allows you to use a function instead of a caching location, allowing you to use external means of authentication.
Ex: https://discord.com/channels/413438066984747026/1127636567914184826

You can just replicate the data that prismarine-auth generates and store it like that. You can also try switching the auth provider to the old Yggdrasil one, setting all of the session data required and then setting the bot to offline mode. I haven't tried that in a bit however. 

> I have a web dashboard where I'm logging in using Microsoft Oauth with the XboxLive.signin scope. I want to use the access_token from there to create a minecraft client in the backend. That's why just using the already existing token

I would honestly save that token into a database that can be fetched using a cacheFactory like I posted a link of an example. It would make it easier.

craftycram:
Aight thanks. Imma look into that later

PR #790, opened 2020-12-14
Fixed password=false not using offline mode

dada513:
```js
false == null
false
```
False is not null, so the expression `options.password != null` would be true even when password was set to false

### DIFF
--- a/src/client/auth.js
+++ b/src/client/auth.js
@@ -19,7 +19,7 @@ module.exports = async function (client, options) {
   const clientToken = options.clientToken || (options.session && options.session.clientToken) || (options.profilesFolder && (await getLauncherProfiles()).clientToken) || UUID.v4().toString().replace(/-/g, '')
   const skipValidation = false || options.skipValidation
   options.accessToken = null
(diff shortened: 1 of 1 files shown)


src/client/auth.js
```diff
@@ -19,7 +19,7 @@ module.exports = async function (client, options) {
   const clientToken = options.clientToken || (options.session && options.session.clientToken) || (options.profilesFolder && (await getLauncherProfiles()).clientToken) || UUID.v4().toString().replace(/-/g, '')
   const skipValidation = false || options.skipValidation
   options.accessToken = null
-  options.haveCredentials = options.password != null || (clientToken != null && options.session != null) || (options.profilesFolder && await hasProfileCredentials())
+  options.haveCredentials = !!options.password || (clientToken && options.session) || (options.profilesFolder && await hasProfileCredentials())
```
rom1504, line 22: Not sure clientToken has the same behavior

dada513: quite sure it would be still better to check if it isnt undefined not just null but reverted that change

extremeheat (approved):

ph0t0shop (approved):

rom1504 (approved):

PR #793, opened 2020-12-18
Fix proxy example

ph0t0shop:
ProfilesFolder needs to be explicitly set to false now because of a change in behavior with the token authentication update


ph0t0shop:
nvm, this is not the fix for the bug and I'm stupid

PR #797, opened 2020-12-28
Don't stringify every packet if debug not enabled

evan-goode:
The JSON.stringify call here takes up a huge amount of CPU time, especially toward the beginning of a bot's lifetime when the server is sending a lot of data.

There might be other places where expensive debug messages are prepared without debug necessarily enabled, I didn't check, but this one is especially costly.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -83,8 +83,10 @@ class Client extends EventEmitter {
       parsed.data = parsed.data.params
       parsed.metadata.state = state
       debug('read packet ' + state + '.' + parsed.metadata.name)
-      const s = JSON.stringify(parsed.data, null, 2)
-      debug(s && s.length > 10000 ? parsed.data : s)
+      if (debug.enabled) {
+        const s = JSON.stringify(parsed.data, null, 2)
+        debug(s && s.length > 10000 ? parsed.data : s)
+      }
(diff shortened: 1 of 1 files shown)


rom1504 (approved):

rom1504:
thanks, that's great

it would be nice if we could have something in the CI to check how much CPU the client is using. Do you have any idea on this?

evan-goode:
Hmm yeah I guess there are already some benchmarks that get run, but those wall-clock times may not be very consistent across CI executions. We could throw in some calls to https://nodejs.org/api/process.html#process_process_cpuusage_previousvalue to measure the CPU time used by each test, which might be more consistent on the CI server, but it will still vary across CPUs.

rom1504:
> there are already some benchmarks that get run

these benchmarks only measure the serialization time. Here this change would not affect them at all, although it will affect the client performance probably a lot.

yeah I guess that trying to measure the CPU usage of the process might be interesting.

evan-goode:
Oh I meant the end to end tests like in `clientTest.js`, but I guess the times currently measured in those are probably including the time to spin up the Notchian server, so not great benchmarks of NMP.

rom1504:
ah yes indeed, most of the time is spent waiting for the notchian server to boot.

Issue #779, opened 2020-11-18
Investigate the possibility to use es6 proxy to parse lazily only if the object is accessed


rom1504:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

would be needed to still parse enough to know the name of the packet.

Would need an useful benchmark to evaluate the gain

idea from @DeltaEvo

PR #800, opened 2021-01-01
Add support for lazy packet parsing

evan-goode:
Dynamically create a "packet_shallow" type for use with the lazy parser at https://github.com/ProtoDef-io/node-protodef/pull/114. 

See https://github.com/PrismarineJS/node-minecraft-protocol/issues/779

### DIFF
--- a/src/transforms/serializer.js
+++ b/src/transforms/serializer.js
@@ -12,10 +12,29 @@ const get = require('lodash.get')
 
 const protocols = {}
 
+const SHALLOW_PACKET_TYPE = 'packet_shallow'
+
 function createProtocol (state, direction, version, customPackets, compiled = true) {
   const key = state + ';' + direction + ';' + version + (compiled ? ';c' : '')
   if (protocols[key]) { return protocols[key] }
   const mcData = require('minecraft-data')(version)
+
+  if (direction === 'toClient') {
+    const nameField = mcData.protocol.play.toClient.types.packet[1].find((field) => field.name === 'name')
+    mcData.protocol.play.toClient.types[SHALLOW_PACKET_TYPE] = [
+      'container',
+      [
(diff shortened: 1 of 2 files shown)


rom1504:
Can you temporarily put "evan-goode/node-protodef#lazy" as version for protodef in package.json ? that way the CI will run properly

evan-goode:
Yep sure thing!

rom1504:
ah yes you need to disable it for 1.7
1.7 cannot be parsed lazily because packets do not have length prefix there

src/transforms/serializer.js
```diff
@@ -2,7 +2,7 @@
 
 const ProtoDef = require('protodef').ProtoDef
 const Serializer = require('protodef').Serializer
-const Parser = require('protodef').FullPacketParser
+const Parser = require('protodef').LazyPacketParser
```
rom1504, line 5: use the FullPacketParser for 1.7 (you can check how the compiled variable is defined)

rom1504:
Would be cool to finish this

evan-goode:
Yeah, I'd like to, sorry for leaving it dangling. I finally have some evidence that the lazy parsing is faster in some use cases! I'm working on a project that involves logging in batches of ~36 bots at once, so it's a nice stress test, and I noticed they finished their tasks around 10% faster after I switched to lazy parsing.

When I left off, I was trying to write a benchmark and got a little confused by the JSON schema for the protodef tests. But I've just pushed something to the protodef branch that I think is on the right track (https://github.com/ProtoDef-io/node-protodef/pull/114/commits/7d8bd7aec0c41fa18eacc2801353ced07ecf38ad), maybe we should continue on that PR?

rom1504:
reopen if you want to finish it

Issue #736, opened 2020-07-01
MCLeaks authentication not working


hriaksnv5678:
[x] The [FAQ](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - minecraft-protocol: 13.0.0
 - server: vanilla 1.15.2
 - node: 14.0.0

## Detailed description of a problem
My question is how would I go about authenticating a user successfully with mcleaks?

my current attempt:
When I use a session from mcleaks, the client partially connects to the server, but does not join, and the client nmp gets stuck after a promise rejection. If I end the node process, the server shows that the player was disconnected, and has the account name, but no id. 

## Current code
```js
var mc = require('minecraft-protocol');

var client = mc.createClient({
    username: "McLeaks",
    version: "1.15.2",
    session: {
        accessToken: "<Token from response to requesting MCLeaks API to redeem a alt token>",
        selectedProfile: {
            name: "<Alt Username from response to requesting MCLeaks API to redeem a alt token>"
        }
    }
});
```

## Expected behaviour
The client joins the server with the username and account provided from mcleaks

## Additional context

the warning that is shown after the client tries to log on
```
(node:7356) UnhandledPromiseRejectionWarning: Error
    at /Users/vrishank/node_modules/yggdrasil/lib/utils.js:60:10
    at /Users/vrishank/node_modules/phin/lib/phin.min.js:1:1150
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:7356) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function wit
hout a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, us
e the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:7356) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will 
terminate the Node.js process with a non-zero exit code.
```

the message that is shown by the server after the client (failed to connect) terminates

```
[...]

PR #802, opened 2021-01-03
mcleaks

MrGeorgen:
mcleaks support #736
should work with `options.auth='mcleaks'`, but I could not test it because the mcleaks website is broken

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "minecraft-data": "^2.70.0",
     "minecraft-folder-path": "^1.1.0",
     "node-fetch": "^2.6.1",
+    "node-mcleaks": "git+https://github.com/funniray/node-mcleaks.git",
     "node-rsa": "^0.4.2",
     "prismarine-nbt": "^1.3.0",
     "protodef": "^1.8.0",
--- a/src/client/encrypt.js
+++ b/src/client/encrypt.js
@@ -3,6 +3,7 @@
 const crypto = require('crypto')
 const debug = require('debug')('minecraft-protocol')
 const yggdrasil = require('yggdrasil')
+const mcleaks = require('node-mcleaks')
 
 module.exports = function (client, options) {
   const yggdrasilServer = yggdrasil.server({ agent: options.agent, host: options.sessionServer || 'https://sessionserver.mojang.com' })
@@ -19,7 +20,7 @@ module.exports = function (client, options) {
         return
       }
       if (options.haveCredentials) {
-        joinServerRequest(onJoinServerResponse)
+        joinServerRequest(options, onJoinServerResponse)
       } else {
         if (packet.serverId !== '-') {
           debug('This server appears to be an online server and you are providing no password, the authentication will probably fail')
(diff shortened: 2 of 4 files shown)


Pandapip1:
In the meantime, is there a plugin for this?

Pandapip1:
I thought of a cludgy solution that might work; I am going to create a custom mcleaks authserver and sessionserver. That way, you can set the authServer and sessionServer options and log in that way.

ArcticZeroo:
I really do not think this is a good idea. I don't think that this library should be directly supporting the use of stolen accounts.

Kashalls:
Well. The only thing I could see working in that regard is changing the auth server host to something user owned and it would have to post on its behalf...

Yggdrasil would do it but it appends session/minecraft/join so a basic webserver should be able to handle it. See if we can get it to throw a http webserver real quick to post then shut it down? Might be useful for exploring 2fa with microsoft in the future.

Kashalls:
@rom1504 What do you think? @ArcticZeroo has a point, thinking maybe this should be its own fork.

rom1504:
You could add some stuff in node-minecraft-protocol to have the necessary options to implement this, so you can do this without having to include mcleaks in node-minecraft-protocol

Pandapip1:
Update on my idea:

It is a bit cludgy, but it seems to work.

```
axios({
	method: "POST",
	url: "https://mcleaks-serverproxy.glitch.me/setserver",
	data: {
		server: "2b2t.com:25565"
	}
}).then(() => {
	let client = mc.createClient({
		host: "2b2t.org",
		port: 25565,
		username: "token here",
		password: "passwords don't matter",
		authServer: "https://mcleaks-serverproxy.glitch.me",
		sessionServer: "https://mcleaks-serverproxy.glitch.me",
		version: "1.12.2"
	});})```

The proxy I made is open-source: https://glitch.com/edit/#!/mcleaks-serverproxy

Kashalls:
Yeah, what I think I'm going to is add support for sessions but it would be up to the implementating user to figure out how to do auth. Maybe have an option that disables yggd session and that way it can emit session or something.

MrGeorgen:
> Update on my idea:
> 
> It is a bit cludgy, but it seems to work.
> 
> ```
> axios({
> 	method: "POST",
> 	url: "https://mcleaks-serverproxy.glitch.me/setserver",
> 	data: {
> 		server: "2b2t.com:25565"
> 	}
> }).then(() => {
> 	let client = mc.createClient({
> 		host: "2b2t.org",
> 		port: 25565,
> 		username: "token here",
> 		password: "passwords don't matter",
> 		authServer: "https://mcleaks-serverproxy.glitch.me",
> 		sessionServer: "https://mcleaks-serverproxy.glitch.me",
> 		version: "1.12.2"
> 	});})
> ```
> 
> The proxy I made is open-source: https://glitch.com/edit/#!/mcleaks-serverproxy
> 
> EDIT: fixed formatting

ok good solution. I guess we can all agree my code isn't good, so I'll close the pr

crazyaccess:
Maybe

PR #806, opened 2021-01-14
Msa device code auth

### LINKED ISSUE #791: 2fa support for microsoft accounts
Nice to Microsoft accounts are now supported. Would be nice 2 factor authentication could also be added

extremeheat:
Fix #791 

* If the `createClient()` is called with `{ auth: 'microsoft', username: 'anyUsername' }`, it tries to authenticate with device code flow and not username and password.
* It first turns the username into a sha1 hash, takes the first few bytes of it into an identifier, uses it to search for token cache files, if not found it goes through the normal auth flow
* If the cache files are found, it checks if the tokens are valid, if they are they get used otherwise the tokens get refreshed

* Adds a `onMsaCode(data)` callback parameter to client options that can receive code info, see electron example for usage

### Test the PR

You can test this PR by running
```
npm i extremeheat/node-minecraft-protocol#msa
```

where you installed mineflayer or node-minecraft-protocol (install mineflayer or nmp *before* you run the command, do not add it to your package.json)

In your `createClient` or `createBot` options, make sure you set `auth: 'microsoft'`.

### DIFF
--- a/examples/client_microsoft_auth/client_msal_auth.js
+++ b/examples/client_microsoft_auth/client_msal_auth.js
@@ -11,7 +11,7 @@ const client = mc.createClient({
   host: process.argv[2],
   port: parseInt(process.argv[3]),
   username: process.argv[4], // your microsoft account email
-//   password: process.argv[5], // your microsoft account password
+  //   password: process.argv[5], // your microsoft account password
   auth: 'microsoft' // This option must be present and set to 'microsoft' to use Microsoft Account Authentication. Failure to do so will result in yggdrasil throwing invalid account information.
 })
 
--- a/examples/client_microsoft_auth/client_msal_auth.js
+++ b/examples/client_microsoft_auth/client_msal_auth.js
@@ -1,0 +1,32 @@
+'use strict'
+
+const mc = require('minecraft-protocol')
+
+if (process.argv.length < 4 || process.argv.length > 6) {
+  console.log('Usage : node echo.js <host> <port> <email>')
+  process.exit(1)
+}
+
+const client = mc.createClient({
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4], // your microsoft account email
+//   password: process.argv[5], // your microsoft account password
+  auth: 'microsoft' // This option must be present and set to 'microsoft' to use Microsoft Account Authentication. Failure to do so will result in yggdrasil throwing invalid account information.
+})
+
+client.on('connect', function () {
+  console.info('connected')
+})
+client.on('disconnect', function (packet) {
+  console.log('disconnected: ' + packet.reason)
+})
+client.on('chat', function (packet) {
+  const jsonMsg = JSON.parse(packet.message)
+  if (jsonMsg.translate === 'chat.type.announcement' || jsonMsg.translate === 'chat.type.text') {
+    const username = jsonMsg.with[0].text
+    const msg = jsonMsg.with[1]
+    if (username === client.username) return
+    client.write('chat', { message: msg })
+  }
+})
--- a/src/client/microsoftAuth.js
+++ b/src/client/microsoftAuth.js
@@ -2,6 +2,7 @@ const XboxLiveAuth = require('@xboxreplay/xboxlive-auth')
 const msal = require('@azure/msal-node')
 const fetch = require('node-fetch')
 const authConstants = require('./authConstants')
+const crypto = require('crypto')
 const { MsaTokenManager, XboxTokenManager, MinecraftTokenManager } = require('./tokens')
 
 const getFetchOptions = {
@@ -41,11 +42,7 @@ const msalConfig = {
   }
 }
 
-const scopes = ['XboxLive.signin', 'offline_access']
-const msa = new MsaTokenManager({ msalConfig, scopes })
-const relayParty = 'rp://api.minecraftservices.com/'
-const xbl = new XboxTokenManager({ relayParty })
-const mca = new MinecraftTokenManager()
+let msa, xbl, mca
 
 function debug (...message) {
   console.debug(message[0])
@@ -140,7 +137,7 @@ async function getXboxToken () {
     const msaToken = await getMsaToken()
     const ut = await xbl.getUserToken(msaToken)
     const xsts = await xbl.getXSTSToken(ut)
-    return xsts.data
+    return xsts
   }
 }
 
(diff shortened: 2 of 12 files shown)


src/client/microsoftAuth.js
```diff
@@ -1,26 +1,107 @@
...
+      },
+      piiLoggingEnabled: false,
+      logLevel: msal.LogLevel.Verbose,
+    }
   }
 }
 
+let scopes = ['XboxLive.signin', 'offline_access']
+let msa = new MsaTokenManager({ msalConfig, scopes })
+let relayParty = 'rp://api.minecraftservices.com/'
```
Kashalls, line 53: This variable is a duplicate of the const on L4, merge them?

src/client/tokens.js
```diff
@@ -0,0 +1,315 @@
...
+  console.debug(message[0])
+}
+
+// Manages Microsoft account tokens
+class MsaTokenManager {
+  constructor({ msalConfig, scopes, cacheLocation }) {
+    this.msaClientId = msalConfig.auth.clientId
+    this.scopes = scopes
+
+    if (!cacheLocation) cacheLocation = path.join(__dirname, './msa-cache.json')
```
Kashalls, line 17: Maybe assign cacheLocation to each of its respective classes like so? I fear we might bleed into other classes.

`this.cacheLocation = cacheLocation ? cacheLocation : path.join(`${__dirname}/msa_cache.json`)`

Kashalls (commented):
I like this, you did somethings I wouldn't have done. Let me know if you want me to try this on my setup. It should test this out good (around 12 live accounts).

src/client/microsoftAuth.js
extremeheat: Yes, there needs to be a bit of refactoring to avoid code duplication. Still haven't decided how much code should be here and how much can be removed to a external package, like how the yggdrasil auth is currently done.

extremeheat:
Yeah, if you could test it out on an account with Minecraft bought if you have one that would be great. I’ve only tested with normal ms accounts as there’s no way to migrate at the moment. 

There’s still refactoring work to do, and decisions to make on how much code can be removed to a separate package. So far this is just the current progression I’ve made for an MVP. A big missing feature at the moment is the ability to use multiple accounts. Right now there’s only one cache and multiple accounts aren’t accounted for. The easy way to deal with multiple accounts would be to have different caches, the slightly harder way would be to alter the caching logic to be able to handle multiple accounts.

Kashalls:
Well I have one of each right now. Pre-Mojang Auth, Mojang, Microsoft U&P and Microsoft 2FA. I'm not worried about switching profiles at the moment, just look at writing it to the file system if it doesn't already exist. We will have to link the email to a token i guess.

extremeheat:
Added basic support for multiple accounts, right now what I have it doing is getting the first few bytes of the sha1 of the passed in username and using that as a profile identifier, and using separate JSON files for caching. The username can be anything, for device code auth it's completely ignored otherwise. The msa cache file should be created now, can you test that it works @Kashalls ?

Kashalls:
I'll try it today and get back to you.

src/client/microsoftAuth.js
```diff
@@ -52,21 +72,132 @@ module.exports = async (client, options) => {
...
   const session = {
-    accessToken: MineServicesResponse.access_token,
+    accessToken: mcAccessToken,
     selectedProfile: profile,
     availableProfile: [profile]
   }
   client.session = session
   client.username = MinecraftProfile.name
-  options.accessToken = MineServicesResponse.access_token
+  options.accessToken = authConstants.MineServicesResponse.access_token
```
Kashalls, line 81: This line is incorrect and will cause the client to error out. 

It should be `options.accessToken = mcAccessToken` because MineServesResponse is a undefined property in authConstants.
(Found during testing)

src/client/microsoftAuth.js
```diff
@@ -52,21 +72,132 @@ module.exports = async (client, options) => {
...
+  if (MineEntitlements.items.length === 0) throw Error('This user does not have any items on its accounts according to minecraft services.')
+
+  postAuthenticate(client, options, MineServicesResponse.access_token)
+}
+
+async function getMsaToken () {
+  if (await msa.verifyTokens()) {
+    return msa.getAccessToken().token
+  } else {
+    const ret = await msa.authDeviceToken((response) => {
```
Kashalls, line 120: authDeviceToken always returns nothing (in console)

regardless whether this is our first time authenticating or not.

extremeheat: The device_code line is debug logging, the path issues I think were caused by not calling path.join, it should be fixed now

extremeheat:
Can you test that it works now @Kashalls ? With respect to the clientId, it's currently hardcoded to use a clientId from microsoft. In the future we could look into reversing the auth flow used in bedrock that it uses for device token auth on devices like the Nintendo Switch. But in the meantime, using APIs exposed with node-msal should work.

Kashalls:
> Can you test that it works now @Kashalls ? With respect to the clientId, it's currently hardcoded to use a clientId from microsoft. In the future we could look into reversing the auth flow used in bedrock that it uses for device token auth on devices like the Nintendo Switch. But in the meantime, using APIs exposed with node-msal should work.

We are still writing cache documents to the node_modules folder for some reason.

extremeheat:
Caches are stored in .minecraft/nmp-cache by default now, if that fails it defaults to __dirname. Also now it falls back to device code auth when username+password auth fails. For Electron/GUI apps that don't use the console, adding a callback parameter to the client options to send the code info would be a good idea too I think.

extremeheat:
Updated the electron example and added a new `onMsaCode()` callback property to the options, it returns the code info needed for sign-in, also added to the docs.

examples/client_microsoft_auth/client_msal_auth.js
```diff
@@ -0,0 +1,32 @@
...
+if (process.argv.length < 4 || process.argv.length > 6) {
+  console.log('Usage : node echo.js <host> <port> <email>')
+  process.exit(1)
+}
+
+const client = mc.createClient({
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4], // your microsoft account email
+  //   password: process.argv[5], // your microsoft account password
```
Kashalls, line 14: Maybe remove this because with msal you dont specify an account password.

Although it would just get ignored either way.

extremeheat: Yes, there's a lot of debug code at the moment, just leaving it there for now as the PR is still a WIP

src/client/tokens.js
```diff
@@ -0,0 +1,299 @@
+const msal = require('@azure/msal-node')
+const XboxLiveAuth = require('@xboxreplay/xboxlive-auth')
+const fs = require('fs')
+const path = require('path')
+const fetch = require('node-fetch')
+const authConstants = require('./authConstants')
+
+function debug (...message) {
```
Kashalls, line 8: Use const debug = require('debug')('minecraft-protocol') instead for items like sending out the data recieved by msal, xla and node-fetch. Anything that has to be sent should use console log like asking a user to complete a 2fa step.

Kashalls (commented):
Make sure that all steps are easily debuggable if the need arises.

I have tested this on my production system and it works alright for my tastes. Have yet to try it with more than 5 bots running at a time however.

rom1504:
is this ready to review/merge yet?

examples/client_microsoft_auth/client_msal_auth.js
extremeheat: fixed

extremeheat:
Cleaned it up a bit, should be mostly stable now. Was just waiting for some people to test it, but it seems to work fine so far so I think it's ready for review. @Kashalls can you test again to make sure the whole auth flow works?  Just fixed an issue with promises.

Kashalls:
> Cleaned it up a bit, should be mostly stable now. Was just waiting for some people to test it, but it seems to work fine so far so I think it's ready for review. @Kashalls can you test again to make sure the whole auth flow works?  Just fixed an issue with promises.

I'm out of town till tomorrow. Will be that good?

extremeheat:
Okay. I didn't update to xbox live library 4.0 so maybe we can do that in another PR when it's ready? It's more of an optional thing, their APIs did change a bit.

I don't think there's much left to do other than that and maybe some extra docs.

rom1504 (approved):

rom1504:
LGTM

Let's merge, that'll give more opportunity for people to try it out

PR #810, opened 2021-01-21
Refactor/Cleanup TCP_DNS

Kashalls:
This PR just cleans up the file and makes everything more streamlined.

If it can do an srv lookup, do the lookup, otherwise just connect. Added some debug just incase dns does not send the srv record.

### DIFF
--- a/src/client/tcp_dns.js
+++ b/src/client/tcp_dns.js
@@ -1,6 +1,8 @@
 const net = require('net')
 const dns = require('dns')
 
+const debug = require('debug')('minecraft-protocol')
+
 module.exports = function (client, options) {
   // Default options
   options.port = options.port || 25565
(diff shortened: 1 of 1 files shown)


rom1504:
Looks correct.
Can you please try it out with a few servers(public local local network) to make sure it actually works ?

Kashalls:
It's working with a server with an srv record, it's also working with local servers. Local network servers are also working.

extremeheat (approved):

rom1504 (approved):

src/client/tcp_dns.js
```diff
@@ -11,35 +13,25 @@ module.exports = function (client, options) {
...
-          }
+          if (err) debug(`srv lookup failed: ${err}`)
 
-          // SRV Lookup resolved conrrectly
+          // SRV Lookup resolved correctly
           if (addresses && addresses.length > 0) {
+            debug(`srv lookup returned ${addresses}`)
             options.host = addresses[0].name
             options.port = addresses[0].port
-            client.setSocket(net.connect(addresses[0].port, addresses[0].name))
```
extremeheat, line 33: The changes here need to be reverted: you need to wait for the srv records request to resolve before you call `client.setSocket`

Kashalls: I missed this during development because it depends how quickly it resolves and I hadn't experienced this issue at all. Its now become a race condition.

Kashalls:
@rom1504 We need to unmerge this commit until someone can fix it. I'm able to fix it right now.

PR #813, opened 2021-01-26
fix launcher_profiles.json problems

Rob9315:
fixes launcher_profiles.json (formerly launcher_accounts.json) not being written to (all clienttokens being deleted) in specific cases

### DIFF
--- a/src/client/auth.js
+++ b/src/client/auth.js
@@ -50,6 +50,7 @@ module.exports = async function (client, options) {
     const cb = function (err, session) {
       if (options.profilesFolder) {
         getLauncherProfiles().then((auths) => {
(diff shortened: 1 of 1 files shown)


Kashalls:
Does this for sure work? Is there any currently open github issues that this fixes?

Kashalls (approved):

Rob9315:
Does not fix the whole issue, just a part of it. Can assure does not create new issues

extremeheat (approved):
is launcher profiles still used? I believe only launcher_accounts.json is used now.

Kashalls:
I think we are keeping it for backwards compatibility?

Rob9315:
We do, this is just for preventing an error that causes an empty file to be written to disk

rom1504 (approved):

PR #814, opened 2021-02-04
Refactored tcp_dns and made it work again

u9g:
previously was failing in some cases for users trying to join certain servers

### DIFF
--- a/src/client/tcp_dns.js
+++ b/src/client/tcp_dns.js
@@ -1,6 +1,6 @@
 const net = require('net')
-const { Resolver } = require('dns').promises;
-const resolver = new Resolver();
+const { Resolver } = require('dns').promises
+const resolver = new Resolver()
 const debug = require('debug')('minecraft-protocol')
 
 module.exports = (client, options) => {
--- a/src/client/tcp_dns.js
+++ b/src/client/tcp_dns.js
@@ -1,8 +1,7 @@
 const net = require('net')
-const dns = require('dns')
-const util = require('util')
+const { Resolver } = require('dns').promises;
+const resolver = new Resolver();
 const debug = require('debug')('minecraft-protocol')
-const resolveSrv = util.promisify(dns.resolveSrv)
 
 module.exports = (client, options) => {
   // Default options
@@ -28,7 +27,7 @@ async function connect (client, options) {
   ) {
     // Try to resolve SRV records for the comain
     try {
-      const addresses = await resolveSrv(`_minecraft._tcp.${options.host}`)
+      const addresses = await resolver.resolveSrv(`_minecraft._tcp.${options.host}`)
       const [address] = addresses
       const { name, port } = address
       // SRV Lookup resolved conrrectly
--- a/src/client/tcp_dns.js
+++ b/src/client/tcp_dns.js
@@ -1,6 +1,7 @@
 const net = require('net')
 const dns = require('dns')
 const util = require('util')
+const debug = require('debug')('minecraft-protocol')
 const resolveSrv = util.promisify(dns.resolveSrv)
 
 module.exports = (client, options) => {
@@ -22,8 +23,8 @@ async function connect (client, options) {
   // If port was not defined (defauls to 25565), host is not an ip neither localhost
   if (
     options.port === 25565 &&
-        net.isIP(options.host) === 0 &&
-        options.host !== 'localhost'
+    net.isIP(options.host) === 0 &&
+    options.host !== 'localhost'
   ) {
     // Try to resolve SRV records for the comain
     try {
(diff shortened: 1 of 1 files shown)


extremeheat:
What's the benefit of splitting up the internal functions? This is usually only something needed if you intend to export functions for external use or if the functions get really big for readability. If you split up small functions it can have the opposite effect.

u9g:
> What's the benefit of splitting up the internal functions? This is usually only needed if you intend to export functions for external use or if the functions get really big for readability. If you split up small functions it can have the opposite effect.

which one in specific? connect or connectwithoptions

Karang:
I think he meant connectwithoptions. The whole file is pretty short so I don't think there is a real readability issue, but i think the single line of the function is self explanatory and the function doesn't bring any value. Its a detail though.

src/client/tcp_dns.js
```diff
@@ -1,37 +1,50 @@
...
+  if (
+    options.port === 25565 &&
+        net.isIP(options.host) === 0 &&
+        options.host !== 'localhost'
+  ) {
+    // Try to resolve SRV records for the comain
+    try {
+      const addresses = await resolveSrv(`_minecraft._tcp.${options.host}`)
+      const [address] = addresses
+      const { name, port } = address
```
Karang, line 32: If address is undefined, the destructuration will error.

src/client/tcp_dns.js
```diff
@@ -1,37 +1,50 @@
...
+      // SRV Lookup resolved conrrectly
+      if (addresses?.length > 0) {
```
Karang, line 34: length can be undefined ? If not, drop the ?

src/client/tcp_dns.js
```diff
@@ -1,37 +1,50 @@
...
-            options.host = addresses[0].name
-            options.port = addresses[0].port
-          }
-        })
+  // If port was not defined (defauls to 25565), host is not an ip neither localhost
+  if (
+    options.port === 25565 &&
+        net.isIP(options.host) === 0 &&
+        options.host !== 'localhost'
+  ) {
```
Karang, line 27: could you put the condition on 1 line or fix the indentation ?

Karang (changes requested):
Can it connect to all servers now ? If yes can you explain what was the issue ?

src/client/tcp_dns.js
u9g: before the if statement was `if (addresses && addresses.length > 0)` so I assumed this achieves the same idea

Karang: https://nodejs.org/api/dns.html#dns_dns_resolvesrv_hostname_callback says it returns an array so its safe to assume the length is defined. Btw, you could use https://nodejs.org/api/dns.html#dns_dns_promises_api instead of promisify()

src/client/tcp_dns.js
u9g: That's true, but the catch would handle it, I tried entering an arbitrary invalid address, `dhusah.afsasd`, and it was caught by the catch which then attempts to call the error resolving domain, which doesn't work, so you get the error described below. I am also adding debugging however that will tell you when srv lookup fails. This is the same outcome that happened before my pr though.
```js
Error: getaddrinfo ENOTFOUND dhusah.afsasd
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26)
    at GetAddrInfoReqWrap.callbackTrampoline (internal/async_hooks.js:129:14) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'dhusah.afsasd'
}
```

src/client/tcp_dns.js
u9g: I fixed indentation

u9g:
> Can it connect to all servers now ? If yes can you explain what was the issue ?

The problem was on the case of an error, the error handler should try to connect directly, but that didn't happen, so the connect just stalled there. By making the function async, I added a dedicated error handler with the try catch

src/client/tcp_dns.js
u9g: 1. Added in commit `removed optional chaining`
2. Added in commit `using native promisified dns`

Kashalls:
The point of my refactor was to reduce the amount of lines you would need. If your going to do this, its better to revert my PR.

extremeheat:
re the breakage, this was the original code - https://github.com/PrismarineJS/node-minecraft-protocol/blob/8f2a027812931d0bef48ed1da2a50a16878dc536/src/client/tcp_dns.js

in https://github.com/PrismarineJS/node-minecraft-protocol/commit/10984fd5d5656811fd8d1dc6512dc9e3f9208611 a call to `client.setSocket` was incorrectly removed which introduced a race bug where a connection can be created before srv records are able to resolve

rom1504:
I reverted the initial commit in https://github.com/PrismarineJS/node-minecraft-protocol/commit/10984fd5d5656811fd8d1dc6512dc9e3f9208611

If you think it still deserves a refactor, rebase on master, refactor what you want and add tests to make sure the result works

PR #815, opened 2021-02-05
Fixes TCP Race Condition

Kashalls:
This PR fixes the dns resolve race condition that occurs when your dns does not resolve as fast as mine does.

### DIFF
--- a/src/client/tcp_dns.js
+++ b/src/client/tcp_dns.js
@@ -1,5 +1,7 @@
 const net = require('net')
-const dns = require('dns')
+const dns = require('dns').promises
+
+const debug = require('debug')('minecraft-protocol')
 
 module.exports = function (client, options) {
   // Default options
@@ -7,7 +9,7 @@ module.exports = function (client, options) {
   options.host = options.host || 'localhost'
 
   if (!options.connect) {
-    options.connect = (client) => {
+    options.connect = async (client) => {
       // Use stream if provided
       if (options.stream) {
         client.setSocket(options.stream)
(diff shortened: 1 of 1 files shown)


extremeheat:
LGTM, but is the promise needed in a callback context?

Kashalls:
> LGTM, but is the promise needed in a callback context?

I think its just overkill now, because some responses will take longer than others.
On the short hand though, this might pose an issue if the dns takes to long to fetch it or something as the client will "hang" until its resolved.

extremeheat:
I meant the switch from a callback to a promise. It’s not a problem, just comment about the code style.

src/client/tcp_dns.js
```diff
@@ -19,19 +19,24 @@ module.exports = function (client, options) {
...
-            options.port = addresses[0].port
-          }
-        })
+          debug(`[DNS] SRV Lookup: ${name}:${port}`)
+
+          options.host = name
+          options.port = port
+        } catch (error) {
+          if (error.code && error.code === 'ENOTFOUND') debug('[DNS] Host does not seem to use an SRV record...')
+          else debug(`[DNS] SRV Resolve Failed: ${error}`)
```
extremeheat, line 32: Does this debug statement work? Is error not an object? You can probably replace the logging with
```suggestion
          catch (error) { debug(error) }
```

rom1504:
I reverted the initial commit in https://github.com/PrismarineJS/node-minecraft-protocol/commit/10984fd5d5656811fd8d1dc6512dc9e3f9208611

If you think it still deserves a refactor, rebase on master, refactor what you want and add tests to make sure the result works

PR #830, opened 2021-03-26
Throw if mcData is invalid

u9g:


src/transforms/serializer.js
```diff
@@ -15,8 +15,12 @@ const protocols = {}
 function createProtocol (state, direction, version, customPackets, compiled = true) {
   const key = state + ';' + direction + ';' + version + (compiled ? ';c' : '')
   if (protocols[key]) { return protocols[key] }
+  
   const mcData = require('minecraft-data')(version)
-
+  if (mcData === null) {
+    throw new Error(`minecraft-data is null for version ${version}`)
```
rom1504, line 21: I think the error should be something like "no data available for version X"

extremeheat: also the `=== null` won't match if it returns undefined, a simple if would work

rom1504: Minecraft data returns null that's fine https://github.com/PrismarineJS/node-minecraft-data/blob/master/index.js#L33

Better not rely on implicit cast if we can

rom1504 (approved):

rom1504:
You need to run npm run fix

u9g:
> You need to run npm run fix

fixed :D

rom1504 (approved):

PR #839, opened 2021-04-05
Fixed server examples

KaffinPX:
There is no loginPackets in mcdata.


rom1504:
https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.16.2/loginPacket.json

PR #842, opened 2021-04-06
Add packet cycletests

u9g:


rom1504 (changes requested):
you need to use this kind of thing https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/test/packetTest.js#L192 to avoid running the tests in each CI worker for all versions

.github/workflows/ci.yml
```diff
@@ -32,3 +32,28 @@ jobs:
...
         CIRCLE_NODE_TOTAL: 10
       run: npm test
+    non-par-test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x]
+        mcVersionIndex: [0, 1]
```
rom1504, line 42: Don't do them in parallel

rom1504 (approved):

PR #849, opened 2021-04-17
Mc.ping return a promise and use callbacks

AngeryFrog:
It's not a suggestion I've seen however having mc.ping return a promise instead of using a callback will allow for people too use async await on this making for cleaner code.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -219,11 +219,16 @@ Where the key is the numeric metadata key and the value is the value of the
 correct data type. You can figure out the types [here](http://wiki.vg/Entities#Entity_Metadata_Format)
 
 
-## mc.ping(options, callback)
+## mc.ping(options)
 
-Ping a minecraft server and retrieve information about it
+`options` is an object containing the following:
+* host : default too locahost
+* port : default too 25565
+* version: default too most recent version
 
-`callback(err, pingResults)`
+Ping a minecraft server and return a promise containing the information about it
+
+`promise( <pending> ).then(pingResult).catch(err)`
 
 `pingResults`:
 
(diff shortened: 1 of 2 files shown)


u9g:
lint

rom1504:
Can you make it both take a callback and return a promise like in https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/book.js#L86 to avoid making it a breaking change ?

u9g:
if you just wrapped this in a promise and thats it, why cant we wrap this and do `mc.ping = callbackify(promisify(mc.ping))`

AngeryFrog:
Sure give me a few mins

AngeryFrog:
Updated the ping function too both return a promise use a callback function if present

rom1504:
tests failing

AngeryFrog:
after testing on local machine produces 2 errors
```
1) client 1.7.10
       offline
         produce a decent error when connecting with the wrong version:
     Error: read ECONNRESET
      at TCP.onStreamRead (internal/stream_base_commons.js:209:20)

  2) client 1.8.8
       offline
         produce a decent error when connecting with the wrong version:
     Error: read ECONNRESET
      at TCP.onStreamRead (internal/stream_base_commons.js:209:20)
```
 however unsure too the cause, is anyone else able too replicate and/or give a suggestion too the cause?

rom1504 (changes requested):
could you use promise and just use a callbackify function to re-add the callback ?
otherwise it makes the code too complex

AngeryFrog:
I hope this is what was intended, if not please comment and I'll re-review it

rom1504 (approved):

PR #857, opened 2021-05-26
Make cache path creation recursive

DecentralisedTech:
If cache dir specified or fallback sir does not exist it can error if parent dirs are missing.

### DIFF
--- a/src/client/authFlow.js
+++ b/src/client/authFlow.js
@@ -41,7 +41,7 @@ class MsAuthFlow {
     let cachePath = cacheDir || mcDefaultFolderPath
     try {
       if (!fs.existsSync(cachePath + '/nmp-cache')) {
-        fs.mkdirSync(cachePath + '/nmp-cache')
+        fs.mkdirSync(cachePath + '/nmp-cache', { recursive: true })
       }
       cachePath += '/nmp-cache'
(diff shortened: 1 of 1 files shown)


extremeheat:
I don’t believe this is necessary, the fallback was added for this purpose in environments where there is no .minecraft folder. What you can do is debug the log statement below it

rom1504:
@extremeheat can you explain why creating the dir recursively is no good ?

extremeheat:
I don't think there was any problem with that, just the original issue was the logging error below. The errors with the .minecraft not existing or not being writable (eg container) are already handled, so the error printed to the console can just be removed.

DecentralisedTech:
Yes the non-existent .minecraft is already handled, this right now just serves as less friction / developer UX when setting paths, I don't see any harm in recursively creating dirs, if this change is not needed thats fine.

Kashalls:
> Yes the non-existent .minecraft is already handled, this right now just serves as less friction / developer UX when setting paths, I don't see any harm in recursively creating dirs, if this change is not needed thats fine.

Is there any decent gains / losses from implementing this? I don't see a specific reason to do it other than silencing a warning.

rom1504 (approved):

PR #862, opened 2021-06-14
Fix loginPacket in serverTest.js for 1.17

piousdeer:
In 1.17 `previousGameMode` type was changed from `u8` to `i8`, setting it to 255 causes tests to fail.


rom1504:
tests failing

piousdeer:
Tests appeared to fail on the snapshot which you replaced with 1.17 in #866, should all work now.

test/serverTest.js
```diff
@@ -222,6 +222,10 @@ for (const supportedVersion of mc.supportedVersions) {
           delete loginPacket.levelType
           delete loginPacket.difficulty
         }
+        if (version.version >= 755) { // 1.17
```
rom1504, line 225: we only support the last version in examples

rom1504:
> Tests appeared to fail on the snapshot which you replaced with 1.17 in #866, should all work now.

you will need to rebase to get the changes

rom1504:
thanks, @u9g redid it in another pr

PR #864, opened 2021-06-15
Fix slightly malformed chunk in example

Heath123:
Context:
image
image
image
image
image
image


Heath123:
This just makes it blank, which isn't exactly right but it's better than what it was before

rom1504 (approved):

PR #868, opened 2021-06-21
Functions from src/client accessable and createClient use autoversion if version is not defined

AngeryFrog:
It is more convinient too have the functions within src/client be accessable from the main library making it easier too use the function class.
Also using auto version in createClient if version is not defined however I am unsure if this is the best way too do this or if this is needed

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -277,4 +277,6 @@ Returns a minecraft protocol [serializer](https://github.com/roblabla/ProtoDef#s
 
 Returns a minecraft protocol [deserializer](https://github.com/roblabla/ProtoDef#parserprotomaintype) for these parameters.
 
+## mc.clientFunctions 
 
+The functions within /src/client for more advanced use of the Client class without directly importing them
--- a/src/client/clientFunctions.js
+++ b/src/client/clientFunctions.js
@@ -13,15 +13,15 @@ const pluginChannels = require('./pluginChannels')
 const versionChecking = require('./versionChecking')
 
 module.exports = {
-    encrypt: encrypt,
-    keepalive: keepalive,
-    compress: compress,
-    auth: auth,
-    microsoftAuth: microsoftAuth,
-    setProtocol: setProtocol,
-    play: play,
-    tcpDns: tcpDns,
-    autoVersion: autoVersion,
-    pluginChannels: pluginChannels,
-    versionChecking: versionChecking
+  encrypt: encrypt,
+  keepalive: keepalive,
+  compress: compress,
+  auth: auth,
+  microsoftAuth: microsoftAuth,
+  setProtocol: setProtocol,
+  play: play,
+  tcpDns: tcpDns,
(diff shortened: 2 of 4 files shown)


rom1504 (changes requested):
it's not good to expose more things unless really needed

why do you need to access these functions directly ?

AngeryFrog:
Too be able too authenticate a client outside of the createClient function, encrypt and compress packets with a custom client, open a connection between a client and an end point with a custom client. From what I found the best way too do this would be with the functions contained within the src/client folder however I could be wrong

AngeryFrog:
Would it be easier too include these functions with the Client class? As my context extends the client class ,and hence the need for these functions outside of it

u9g:
> Would it be easier too include these functions with the Client class? As my context extends the client class ,and hence the need for these functions outside of it

Why do you make a custom client object?

AngeryFrog:
> make a custom client object?

In this case to include saving packets for ML and extend authentication functionality, however off the top of my head I can think of a few reasons such as implementing custom physics and automating more complex behavior. My issue is without using the createClient function a lot of this functionality is not possible without directly importing from the src/client folder

u9g:
> > make a custom client object?
> 
> In this case to include saving packets for ML and extend authentication functionality, however off the top of my head I can think of a few reasons such as implementing custom physics and automating more complex behavior. My issue is without using the createClient function a lot of this functionality is not possible without directly importing from the src/client folder

why couldnt you use createClient and just add a listener for packets as in client.on('packet', … , also, I dont know if you are aware of it, but https://github.com/prismarinejs/mineflayer has implemented a lot of that complex behavior and has implemented physics (that can be changed easily)

AngeryFrog:
> why couldnt you use createClient and just add a listener for packets as in client.on('packet', … , also, I dont know if you are aware of it, but https://github.com/prismarinejs/mineflayer has implemented a lot of that complex behavior and has implemented physics (that can be changed easily) , re: the authentication, what do you mean by that?

Creating a listener on a packet is a perfectly good idea however in this case I am planning too use the same code and functions over and over again hence it would be cleaner too extend the Client object, I am fully aware that mineflayer has implemented complex behavior and physics however it's my intent too only write what is needed. 
As for authentication I intend too do fetch authentication info outside of the custom Client object and use it within the custom client object when needed, it's a little difficult too explain and I know I've not done a good job at explaining it however the only other way too give a client authentication info from what I know of is too use the createClient function and my issue with that is that it does other things.

If any of this can be worked around from within the library that I've never noticed I'm happy too use them

extremeheat:
Sounds like an XY problem. If you’re just trying to authenticate, why not use Yggdrasil? This is a purpose built lib for the Minecraft protocol.

AngeryFrog:
After research Yggdrasil seems too be able too help me with my issue, thank you for the help and sorry for the pull request without doing full research into the issue.

PR #869, opened 2021-06-22
make port discovery automatic & update java ver for 1.17

u9g:

PR #870, opened 2021-06-29
automatic port "discovery" for packets and server tests

Kashalls:
This PR builds on https://github.com/PrismarineJS/node-minecraft-protocol/pull/869 by making the other two tests for packets and servers also use port "discovery".


u9g (approved):
LGTM, thanks

PR #881, opened 2021-08-01
Fix 1.7.10 auto versioning

Beaness:
This fixes the auto versioning. 
1.7.10 has to be hardcoded since the protocol ids of the first 1.8 snapshots use the 1.7 protocol id.
problem

### DIFF
--- a/src/server/handshake.js
+++ b/src/server/handshake.js
@@ -1,3 +1,4 @@
+const mcData = require('minecraft-data')
 const states = require('../states')
 
 module.exports = function (client, server, { version }) {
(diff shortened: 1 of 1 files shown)


nickelpro (approved):
Correct from the protocol POV, should probably log a warning here about the version ambiguity and assuming 1.7.10 (although, who's running a 1.8 snapshot server?)

LGTM

u9g:
> Correct from the protocol POV, should probably log a warning here about the version ambiguity and assuming 1.7.10 (although, who's running a 1.8 snapshot server?)
> 
> LGTM

There’s similar ambiguity with other versions like this, so I think it’s not necessary.

src/server/handshake.js
```diff
@@ -9,10 +9,14 @@ module.exports = function (client, server, { version }) {
     client.protocolVersion = packet.protocolVersion
 
     if (version === false || version === undefined) {
-      if (require('minecraft-data')(client.protocolVersion)) {
-        client.version = client.protocolVersion
+      if (client.protocolVersion === 5) { // The first snapshot versions of 1.8 uses the 1.7 protocol id, after 1.8 each snapshot has a different protocol id.
```
rom1504, line 12: what is the bug and why can't it be fixed in the data ?

src/server/handshake.js
```diff
@@ -9,10 +9,14 @@ module.exports = function (client, server, { version }) {
...
+        client.version = '1.7.10'
```
u9g, line 13: I know this is the only way to make it works, but this is no longer always a protocolversion anymore, which is odd

src/server/handshake.js
nickelpro: Incompatible protocol version that report the same protocolVersion. Can't be fixed in data because there's nothing to fix, both the snapshots and 1.7.10 version report proto version 5

rom1504: looking at https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/common/protocolVersions.json#L3008 ; what prevent us to take the last version for this protocol version and for this major version ?

rom1504: this should be fixed directly in node-minecraft-data version guessing from protocol version logic

rom1504: https://github.com/PrismarineJS/node-minecraft-data/blob/63e8769d8f3e71d7df0678773ac8e58f9b58c178/index.js#L62 there

nickelpro:
Should be fixed

rom1504:
https://github.com/PrismarineJS/node-minecraft-data/pull/103 fixes it

can you check it works for you now ?

PR #882, opened 2021-08-01
Make raw event pass full buffer

Heath123:
Not sure if this is right but surely the raw data should pass the whole thing before any parsing?


Heath123:
This was completely wrong, ignore this

Issue #833, opened 2021-03-27
Move microsoft auth to yggdrasil or its own package


rom1504:
It's self contained and too big to be here

PR #884, opened 2021-08-04
Implement Prismarine-Auth

Kashalls:
Addresses #833
Everything tested out locally for all three types of accounts.

Pre-mojang, mojang, microsoft (password/devicecode)

Ready for review


Kashalls:
This will not be ready for review until https://github.com/PrismarineJS/prismarine-auth/pull/6 is available and tests are passing.

rom1504:
is this ready @Kashalls @extremeheat ?
we have no online tests, so please checks all paths:
* mojang auth with no cache
* microsoft auth with no cache
* mojang with cache
* microsoft with cache
* offline mode

Kashalls:
> is this ready @Kashalls @extremeheat ?
> we have no online tests, so please checks all paths:
> * mojang auth with no cache
> * microsoft auth with no cache
> * mojang with cache
> * microsoft with cache
> * offline mode

I will test it tonight. Should I write an online test?

src/client/microsoftAuth.js
```diff
@@ -11,113 +10,61 @@ const getFetchOptions = {
...
-  if (!minecraftProfile.id) {
-    debug('[mc] profile', minecraftProfile)
+  const Authflow = new PrismarineAuth(options.username, options.profilesFolder, options, options.onMsaCode)
+  const accessToken = await Authflow.getMinecraftJavaToken()
+
+  // We have a minecraft access token.
+  // Let's verify entitlements.
+
+  getFetchOptions.headers.Authorization = `Bearer ${accessToken}`
+  const MineEntitlements = await fetch(Endpoints.MinecraftServicesEntitlement, getFetchOptions).then(checkStatus)
```
extremeheat, line 25: please use lower camel casing when declaring variables, upper camel case when defining classes. This makes it easier for others to differentiate.

src/client/microsoftAuth.js
```diff
@@ -11,113 +10,61 @@ const getFetchOptions = {
...
+  if (MineEntitlements.items.length === 0) throw Error('This user does not possess any entitlements on this account according to minecraft services.')
+
+  // We know that this account owns minecraft.
+  // Let's fetch the profile data.
+
+  options.haveCredentials = accessToken != null
+
+  const MinecraftProfile = await fetch(Endpoints.MinecraftServicesProfile, getFetchOptions)
```
extremeheat, line 33: endpoints should not need to be exposed, why not do this check in prismarine-auth?

src/client/microsoftAuth.js
```diff
@@ -11,113 +10,61 @@ const getFetchOptions = {
...
-  let minecraftProfile
-  const res = await fetch(authConstants.MinecraftServicesProfile, getFetchOptions)
-  if (res.ok) { // res.status >= 200 && res.status < 300
-    minecraftProfile = await res.json()
-  } else {
-    const user = msa ? msa.getUsers()[0] : options.username
-    throw Error(`Failed to obtain Minecraft profile data for '${user?.username}', does the account own Minecraft Java? Server returned: ${res.statusText}`)
+async function authenticate (client, options) {
+  if (!options.profilesFolder && options.profilesFolder !== false) {
+    options.profilesFolder = require('minecraft-folder-path')
```
extremeheat, line 15: put this require at the top of the file

src/client/microsoftAuth.js
```diff
@@ -11,113 +10,61 @@ const getFetchOptions = {
...
-
-  let minecraftProfile
-  const res = await fetch(authConstants.MinecraftServicesProfile, getFetchOptions)
-  if (res.ok) { // res.status >= 200 && res.status < 300
-    minecraftProfile = await res.json()
-  } else {
-    const user = msa ? msa.getUsers()[0] : options.username
-    throw Error(`Failed to obtain Minecraft profile data for '${user?.username}', does the account own Minecraft Java? Server returned: ${res.statusText}`)
+async function authenticate (client, options) {
+  if (!options.profilesFolder && options.profilesFolder !== false) {
```
extremeheat, line 14: why not just check `if (!options.profilesFolder)` ?

extremeheat (changes requested):

src/client/microsoftAuth.js
Kashalls: https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client/auth.js#L10

extremeheat: I don't get it. The code underneath that sets the folder path to .minecraft is never hit if the option is unspecified (undefined)

extremeheat: OK, nevermind, it's set to false. Why this specific exception?

Kashalls: Was going for consistency here, but I'll do that over what we have.

src/client/microsoftAuth.js
Kashalls: Reason being is that prismarine-auth handles authentication and caching tokens. I can merge this into prismarine-auth if you feel the need to have it there instead of here. This just checks for entitlements as you would know.

src/client/microsoftAuth.js
extremeheat: ok, if there's a valid reason we can keep it. I just wanted some clarification for it, do you know the intention behind it ? The comment doesn't make it entirely clear there.

src/client/microsoftAuth.js
extremeheat: ok, do you think it would make sense for others to have this information if they're also using minecraft auth ? That way people don't have to have copies of this in their code that need mc auth. I think it would be a helpful additional check, maybe optional option to the getMinecraftToken call ?

src/client/microsoftAuth.js
Kashalls: Basically it prevents people from setting profileFolder to true and having it crash.

src/client/microsoftAuth.js
Kashalls: I'm not sure what you're getting at here The only thing I can think of is an extra check to see if you're actually entitled to the game or not

Kashalls:
Currently depending on https://github.com/PrismarineJS/prismarine-auth/pull/7

rom1504:
did you check it works in all cases I mentionned above ?

Kashalls:
> did you check it works in all cases I mentionned above ?

I was able to test everything without issues, I still would like other to briefly test this to ensure it.

Here's my test code. I would just change options to fit other accounts.
```js
const { Authflow } = require('prismarine-auth')
const direct = path.join(require('minecraft-folder-path'), 'nmp-cache')
Authflow.resetTokenCaches(direct)

const doAuth = async () => {
  const flow = new Authflow('username', direct, { password: 'wowzers' })
  const response = await flow.getMinecraftJavaToken()
  console.log(response)
}

doAuth()
```

Kashalls:
I'm not sure if we want to remove these console.logs from https://github.com/PrismarineJS/prismarine-auth/blob/master/src/MicrosoftAuthFlow.js#L94 and handle the code in nmp itself.

rom1504:
Did you test with nmp createClient directly ?

Kashalls:
> Did you test with nmp createClient directly ?

Yes

rom1504 (approved):

rom1504:
Ok I'm merging it. Code looks good.

if we get complaints I'll revert and you can try again.

rom1504:
would be great to do this again @Kashalls 
related https://github.com/PrismarineJS/prismarine-auth/issues/10

PR #888, opened 2021-08-09
Improve Typescript definitions and API.md

GhqstMC:
This started out as a small PR to copy missing properties and methods from API.md to the typescript definition file, but as I discovered things in the typescript definitions that aren't in API.md I decided to try my best at adding those in too. The overall goal is to reach a state where API.md and the typescript definitions have no differences.

There are several places in API.md where I need confirmation on what some properties are/what some methods do. I've also reorganized some things so I think it makes sense, and I've ordered the typescript definitions for Server, ServerOptions, Client, and ClientOptions to match the order of API.md so discrepencies can be noticed and fixed more easily.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -134,7 +134,7 @@ If `fullReason` is not defined, then the `reason` will be used.
 
 ### client.connect(port, host)
 
-TODO
+Used by the [Client Class](https://github.com/PrismarineJS/node-minecraft-protocol/blob/d9d01c8be4921bb38e7b59d9c18afd771615ba22/src/client.js) to connect to a server, done by createClient automatically
 
 ### client.setSocket(socket)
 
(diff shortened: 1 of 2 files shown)


docs/API.md
```diff
@@ -120,13 +121,37 @@ Takes a minecraft `version` as second argument.
...
+
+Registers a plugin channel with the given `name` and protodef `typeDefinition`
+
+### client.unregisterChannel(name)
+
+Unregisters a plugin channel.
+
+### client.writeChannel(channel, params)
+
+TODO: clarify what the heck this does exactly
```
rom1504, line 153: Read the code or remove this part

rom1504: I mean the description not the section

u9g: This is now done

rom1504 (approved):

PR #889, opened 2021-08-12
Client throw on write error

nickelpro:
Don't ignore write errors, propagate them

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -218,7 +218,7 @@ class Client extends EventEmitter {
     if (!this.serializer.writable) { return }
     debug('writing packet ' + this.state + '.' + name)
     debug(params)
-    this.serializer.write({ name, params })
+    this.serializer.write({ name, params }, (e) => { if (e) throw e })
   }
 
   writeRaw (buffer) {


src/client.js
```diff
@@ -218,7 +218,7 @@ class Client extends EventEmitter {
     if (!this.serializer.writable) { return }
     debug('writing packet ' + this.state + '.' + name)
     debug(params)
-    this.serializer.write({ name, params })
+    this.serializer.write({ name, params }, (e) => { if (e) throw e })
```
rom1504, line 221: This doesn't happen here
It's the error event which is already handled

nickelpro: This is the callback that is propagated from errors in _transform. The write method is here:
https://github.com/nodejs/readable-stream/blob/master/lib/_stream_writable.js#L288-L310

nickelpro: If you don't pass a callback the errors in transform get dropped

rom1504: ok then we should emit an error with the err event in this case
crashing the program is not the best imo
users could have a larger program on top of this, that could for example decide to restart the client with another version/server

PR #897, opened 2021-08-22
Change how we export states enum

u9g:


rom1504:
what do you think about removing it instead ?

rom1504:
stale

PR #898, opened 2021-08-23
Update Online Mode Exceptions

u9g:


extremeheat:
Why?

extremeheat:
Sounds like this should be a Map and/or a Set

PR #902, opened 2021-09-02
Make examples work for 1.17.1

u9g:


u9g:
I fixed all examples but really we should just take the whole login packet from minecraft-data

rom1504 (approved):

rom1504:
> I fixed all examples but really we should just take the whole login packet from minecraft-data

yes!

PR #903, opened 2021-09-03
Make all events have promiselike returns for async

u9g:


rom1504 (approved):

DeltaEvo:
I'm not sure it's a good thing to have changed the return types to PromiseLike since the EventEmitter will not handle rejections of Promises
This would be possible if `captureRejections` is enabled on the EventEmitter but it's still experimental https://nodejs.org/api/events.html#events_capture_rejections_of_promises

PR #913, opened 2021-10-24
Add the ability to send a fake hostname in set_protocol

GhqstMC:
This is a change made for my own purposes but U9G said it was worth PRing. The usecase for this is something like connecting to hypixel with the raw ip but saying you connected with "mc.hypixel.net" (otherwise it would kick you.)


u9g (approved):
LGTM

PR #920, opened 2021-11-05
add default for authTitle

U5B:


u9g:
please run `npm run fix`

U5B:
> please run `npm run fix`

linted.

should this go into prismarine-auth instead as a default?

src/client/microsoftAuth.js
```diff
@@ -8,6 +8,10 @@ async function authenticate (client, options) {
     options.profilesFolder = path.join(minecraftFolderPath, 'nmp-cache')
   }
 
+  if (!options.authTitle) {
```
extremeheat, line 11: Make this check === undefined, in pauth false is also allowed

rom1504:
@extremeheat does that look correct?

extremeheat:
Lgtm

PR #922, opened 2021-11-10
Fixed TypeScript error TS2769 when authenticating with mojang

liquiddevelopmentnet:
Fixed TypeScript Typing Error TS2769: No overload matches this call.


u9g (approved):
Thanks

PR #929, opened 2021-12-04
Remove entitlement check in microsoftAuth

extremeheat:
redundant with the existing profile check


Kashalls (commented):
Are we still able to fetch entitlements if they don't have any? We can still log it if they don't...

extremeheat:
Per the issue above, the entitlement check is not necessary or relevant to if the user can play Minecraft (for example via xbox game pass), only having a profile is.

Kashalls (approved):

PR #930, opened 2021-12-09
1.18 support

rom1504:


u9g:
need to update workflow

rom1504:
* [ ] make https://github.com/PrismarineJS/node-minecraft-protocol/blob/6095c81a34a88b4c6e5f0d9aac63eaa5a31e3959/test/non-par-test.js#L10 use https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/version.js#L5 
* [ ] update data in https://github.com/PrismarineJS/minecraft-packets/tree/main/data/pc

PR #931, opened 2021-12-10
Make packetCycleTest run for all versions & seperate lint step from test step

u9g:


rom1504:
why is it not failing? it should fail until mc packets is updated

u9g:
> why is it not failing? it should fail until mc packets is updated

didn't add env to yml, for now nothing is in env, that way while nothing is in the env var it runs all

u9g:
> why is it not failing? it should fail until mc packets is updated

Okay, failing now :)

u9g:
Not understanding why it's just not running the test now?

test/packetCycleTests.js
```diff
@@ -0,0 +1,48 @@
...
+const serializers = {}
+const deserializers = {}
+const makeClientSerializer = version => createSerializer({ state: states.PLAY, version, isServer: true })
+const makeClientDeserializer = version => createDeserializer({ state: states.PLAY, version })
+
+describe('Packet cycle tests', () => {
+  for (const supportedVersion of supportedVersions) {
+    const mcData = require('minecraft-data')(supportedVersion)
+    const version = mcData.version.minecraftVersion
+    it('Packet cycle tests for ${version}', () => {
```
rom1504, line 16: Is that the right thing

u9g: fixed, but now the grep isn't being called: https://github.com/PrismarineJS/node-minecraft-protocol/runs/4488855401?check_suite_focus=true#step:6:13

rom1504:
* [ ] use the nearest major available data ; or remap/redump in minecraft packets
* [ ] dump 1.17 and 1.18

rom1504:
stuck pr

rom1504:
would be nice to finish

rom1504:
please re-open if you want to finish it

PR #934, opened 2021-12-12
fix mc 1.18

rom1504:


rom1504:
chunkBlockEntity making things fail (see https://github.com/PrismarineJS/node-minecraft-protocol/runs/4498961835?check_suite_focus=true )

PR #937, opened 2021-12-16
Session doc fix

IceTank:


docs/API.md
```diff
@@ -88,8 +88,10 @@ Returns a `Client` instance and perform login.
    is blank, and `profilesFolder` is specified, we auth with the tokens there instead.
    If neither `password` or `profilesFolder` are specified, we connect in offline mode.
  * host : default to localhost
- * clientToken : generated if a password is given
- * accessToken : generated if a password or microsoft account is given
+ * session : An object holding clientToken and accessToken. Generated after logging in using username + password with mojang auth or after logging in using microsoft auth. `clientToken`, `accessToken` and `selectedProfile: {name: '<username>'}` can be set inside of `session` when using createClient to login with a client and access Token instead of a password. 
+   * clientToken : generated if a password is given or can be set when when using createClient
+   * accessToken : generated if a password or microsoft account is given or can be set when using createBot
+   * selectedProfile : generated if a password or microsoft account is given. Can be set as a object with one property `name` that specifies the selected profile. The selected profile is the same name as the in game username.
```
U5B, line 94: `selectedProfile` has two properties 
`id:` which is the uuid of the account
`name:` which is the username of the account

I do not know if the uuid is required but it would match what is generated in:
https://github.com/PrismarineJS/node-minecraft-protocol/blob/010d6b229c88b7cc0d7c51efb01a5ea510bdb58a/src/client/auth.js#L127-L132

U5B (changes requested):

IceTank:
I updated the documentation to include selected profile. I also updated the typings for clientOptions.

docs/API.md
IceTank: It looks like the uuid is required. I don't know if the name is required but node-minecraft-protocol does not like it if you do not provide the name.

PR #948, opened 2022-01-05
docs(readme): add SteveProxy to projects list

MrZillaGold:
Also added dots at the end of descriptions of other projects.


rom1504:
is that proxy meant to be used client side or server side?

MrZillaGold:
@rom1504 Both client and server side, when used on the server side, some plugin functionality will be unavailable

PR #950, opened 2022-01-05
Add callback  for beforeLogin function on server

C1assic:
This addscallback to beforeLogin, which is useful forusing deferred functions. 
For example: database queries.


rom1504:
I don't understand the purpose of this, and I'm pretty sure the proposed code doesn't do anything useful

Can you explain an use case and what you want to do ?

rom1504:
too old

PR #952, opened 2022-01-20
Fix proxy example not working for servers running in offline mode

IceTank:
Some yggdrasil authentication attempt somehow fails when trying to connect to an offline server making the proxy fail to connect.

### DIFF
--- a/examples/proxy/proxy.js
+++ b/examples/proxy/proxy.js
@@ -96,7 +96,8 @@ srv.on('login', function (client) {
     port: port,
     username: client.username,
     keepAlive: false,
-    version: version
+    version: version,
+    skipValidation: true
   })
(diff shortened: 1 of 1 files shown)


rom1504:
that's a much larger issue in mineflayer

we should make it much more explicit whether offline or online mode is used

IceTank:
>we should make it much more explicit whether offline or online mode is used

So make the example ask for credentials instead? And then try to connect as an online or offline client?

rom1504:
No, what I meant is this shows an issue in createClient
As this PR shows it is now confusing to know whether offline mode will be used or not
I think we should improve the doc on the topic or improve the API itself

rom1504:
this is not needed now, issue fixed in nmp

PR #958, opened 2022-02-06
uses microsoft auth by default

rom1504:


rom1504:
is that correct @extremeheat @Kashalls ?

Kashalls:
> is that correct @extremeheat @Kashalls ?

https://github.com/PrismarineJS/node-minecraft-protocol/blob/4691abd0365721d0c7981c4140c9009b434bc18f/src/createClient.js#L37 ?

extremeheat:
Hmm, once mojang auth is removed the auth options might as well be `true` or `false` I think. But I do agree with making mojang auth explicit, that way later on when the mojang auth endpoints are disabled an error can be printed explaining it's no longer possible to do mojang auth

rom1504:
So what about that mode "offline" ? It was never really a Mojang auth and it's definitely not a Microsoft auth, it's simply no auth. I think making it explicit is good.

Regarding true/false vs named, i think named is better because who knows if there won't be yet a new auth scheme we want to support in the future ;)

rom1504:
And should offline be the default?
It used to be that way.

(IE no password = offline)

Now it's pretty confusing

extremeheat:
Either way I think this will break things. Since some people don't currently put anything in for `auth`, switching it to offline would silently cause the bot to fail until connecting to the server (but I think with hideErrors false should be less of an issue), with 'microsoft' it could require user action.

I'm not sure what most people are running, but if most people are running/testing bots locally (this is what I'm guessing) I think it makes sense to default to offline.

Kashalls:
Looks like by March 10th 2022, all mojang accounts will no longer authenticate. (Sounds like they are just taking the auth servers down.)

As we previously used mojang as the default authentications scheme, I think we should default it to Microsoft. If the player doesn't want to use a real account, let them use offline. 

> Either way I think this will break things. Since some people don't currently put anything in for `auth`, switching it to offline would silently cause the bot to fail until connecting to the server (but I think with hideErrors false should be less of an issue), with 'microsoft' it could require user action.
> 
> I'm not sure where most people are running bots, but if most people are running/testing bots locally (this is what I'm guessing) I think it makes sense to default to offline.

I like this as well. I only have a couple bots deployed in production, where I have explicitly set `auth: 'microsoft'` in their configurations. Maybe we can instead of sending the output to debug, we can pipe to stdin and just display a little `Hey! Unless specified, your bot won't attempt to auth without specifying the auth type!` ?

extremeheat:
I think we should switch to offline by default. If people want to do Microsoft authentication the auth field is required anyway, so it seems to me like the least breaking path.

Kashalls:
> I think we should switch to offline by default. If people want to do Microsoft authentication the auth field is required anyway, so it seems to me like the least breaking path.

I changed it to use offline mode as the default and fallback.

rom1504:
need a redo

PR #965, opened 2022-02-27
fix plugin channel registration

RichardDorian:
I hope this is "mergable" with my other pull request. I finally fixed the plugin registration so now it's working even if the channel registred is custom.

### DIFF
--- a/src/client/pluginChannels.js
+++ b/src/client/pluginChannels.js
@@ -23,7 +23,7 @@ module.exports = function (client, options) {
 
   function registerChannel (name, parser, custom) {
     if (custom) {
-      client.writeChannel('REGISTER', name)
+      client.writeChannel('REGISTER', [name])
     }
     if (parser) proto.addType(name, parser)
     channels.push(name)
@@ -32,7 +32,7 @@ module.exports = function (client, options) {
 
   function unregisterChannel (channel, custom) {
     if (custom) {
-      client.writeChannel('UNREGISTER', channel)
+      client.writeChannel('UNREGISTER', [channel])
     }
     const index = channels.find(function (name) {
       return channel === name
@@ -87,7 +87,7 @@ module.exports = function (client, options) {
   function writeDumbArr (value, buf, offset) {
     // TODO: Remove trailing \0
     value.forEach(function (v) {
-      offset += this.write(v, buf, offset, 'cstring', {})
+      offset += buf.write(v, offset)
     })
     return offset
   }
--- a/src/client/pluginChannels.js
+++ b/src/client/pluginChannels.js
@@ -17,13 +17,15 @@ module.exports = function (client, options) {
   client.registerChannel('REGISTER', ['registerarr', []])
   client.registerChannel('UNREGISTER', ['registerarr', []])
 
-  if (options.protocolVersion >= 385) { // 1.13-pre3 (385) added Added Login Plugin Message (https://wiki.vg/Protocol_History#1.13-pre3)
+  const above385 = options.protocolVersion >= 385
+  if (above385) { // 1.13-pre3 (385) added Added Login Plugin Message (https://wiki.vg/Protocol_History#1.13-pre3)
     client.on('login_plugin_request', onLoginPluginRequest)
   }
+  const channelNames = above385 ? ['minecraft:register', 'minecraft:unregister'] : ['REGISTER', 'UNREGISTER']
 
   function registerChannel (name, parser, custom) {
     if (custom) {
-      client.writeChannel('REGISTER', [name])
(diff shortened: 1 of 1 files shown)


rom1504:
can you explain what this is doing?

RichardDorian:
Oh yeah sorry. As referenced [here](https://wiki.vg/Plugin_channels) the payload for the register and unregister channel should be a null separated array of strings and not just string. A pull request fixed this problem a long time ago. But when registering a custom plugin channel the `registerChannel` function was passing a string and not an array of strings. The `writeDumbArr` function was broken as well so I fixed it.

src/client/pluginChannels.js
```diff
@@ -87,7 +89,7 @@ module.exports = function (client, options) {
   function writeDumbArr (value, buf, offset) {
     // TODO: Remove trailing \0
     value.forEach(function (v) {
-      offset += this.write(v, buf, offset, 'cstring', {})
+      offset += buf.write(v, offset)
```
rom1504, line 92: this is a change of meaning for the value arg, it now assumes that value is a bugger, is that true?

RichardDorian: No, value is the array of strings

rom1504: ok so that code is not going to work then

RichardDorian: It's working and tested..

RichardDorian: `this.write` throw an error. I mean, it says `this.write is not a function`. Which is weird because if you do a `console.log(this.write)` above a function is correctly logged. But I'm sure that `buf.write` should be used here.

rom1504: this.write is indeed wrong because it's not a function
However buf.write is also wrong because it's for writing raw bytes

You need to use protodef to serialize that array of strings

rom1504: No actually pretty sure that this.write is defined, the this is the protodef instance

RichardDorian: > However buf.write is also wrong because it's for writing raw bytes

Unless you set the encoding

RichardDorian: > No actually pretty sure that this.write is defined, the this is the protodef instance

`this` refers to the protodef instance for some reason. But again `this.write` throws an error when called in `writeDumbArr`

RichardDorian: And `buf.write` should be used instead of `this.write` even if the method was working because the `writeDumbArr` is used to serialize the actual array

rom1504: https://nodejs.org/api/buffer.html#bufwritestring-offset-length-encoding 
Buffer.write takes a string as argument, not an array of strings

RichardDorian: It's not the array of strings passed inside the big.write it's the v variable (value.forEach)

rom1504: I see that's true
still missing the zero terminator

RichardDorian: The zero terminator is not needed, should I add it anyway?

rom1504: how do you know it's not needed?

rom1504: if there's no 0 termination then you need length prefixing
you would definitely need to adapt read and size of as well

RichardDorian: A null separated list of strings (for me) looks like this (hex)
```
63 68 61 6E 6E 65 6C 31 00 CA FE BA BE
<----------------------------------> <--> <--------------->
               Channel1.                     Channel 2
                                      Separator

There's no need for a terminator at the end

Comment (RichardDorian) [reply src/client/pluginChannels.js:92]: Oh wait are you telling me that if i put two strings there's no separator?

Comment (RichardDorian) [reply src/client/pluginChannels.js:92]: Ok I get it. I'll fix this when I'll be home. I'm still wondering how the previous code was working.

Comment (RichardDorian) [reply src/client/pluginChannels.js]: Updated! Now the null separator is present

File: src/client/pluginChannels.js [hunk 2]
```diff
@@ -87,7 +89,7 @@ module.exports = function (client, options) {
   function writeDumbArr (value, buf, offset) {
     // TODO: Remove trailing \0
     value.forEach(function (v) {
-      offset += this.write(v, buf, offset, 'cstring', {})
+      offset += buf.write(v + '\0', offset)
```
rom1504, line 92: can you do it by using protodef instead ? that's what `cstring` does

RichardDorian:
Ok done! I hope it's the last time lol

PR #975, opened 2022-03-28
Add basic deserializer and serializer typings

IceTank:


IceTank:
@jojomatik try this patch https://github.com/PrismarineJS/node-minecraft-protocol/pull/982

PR #976, opened 2022-03-29
fix: cross version ping

jojomatik:
Enable dynamic cross version support for ping by responding with the client version and protocol version if dynamic cross version support is enabled.

Additional preparatory steps:
- **docs: explain version parameter https://github.com/PrismarineJS/node-minecraft-protocol/commit/006aabf20a17795ebe86ad382e1b16675430e902**
  Explain version parameter more explicitly. Remove reference to outdated versions. Describe dynamic cross version support with parameter value `false`.
- **fix(types): allow boolean value for version parameter https://github.com/PrismarineJS/node-minecraft-protocol/commit/a4bda0f9a95f01686fd114062607ec5b20951bc2**
  Allow boolean value for version parameter. This makes dynamic cross version support usable in typescript projects.

References https://github.com/PrismarineJS/node-minecraft-protocol/pull/508#issuecomment-1082092675

Please be extra cautious while reviewing. I have not been able to test this out properly. Thank you very much :)

### DIFF
--- a/src/server/ping.js
+++ b/src/server/ping.js
@@ -1,15 +1,23 @@
 const endianToggle = require('endian-toggle')
 
-module.exports = function (client, server, { beforePing = null }) {
+module.exports = function (client, server, { beforePing = null, version }) {
   client.once('ping_start', onPing)
   client.once('legacy_server_list_ping', onLegacyPing)
 
   function onPing () {
+    // Use client version if dynamic cross version support is enabled.
+    const responseVersion = (version === false)
+      ? {
+          name: client.version,
+          protocol: client.protocolVersion
+        }
+      : {
+          name: server.mcversion.minecraftVersion,
+          protocol: server.mcversion.version
+        }
+
     const response = {
-      version: {
-        name: server.mcversion.minecraftVersion,
-        protocol: server.mcversion.version
-      },
+      version: responseVersion,
       players: {
         max: server.maxPlayers,
         online: server.playerCount,
--- a/docs/API.md
+++ b/docs/API.md
@@ -20,7 +20,7 @@ automatically logged in and validated against mojang's auth.
  * motd : default to "A Minecraft server"
  * maxPlayers : default to 20
  * keepAlive : send keep alive packets : default to true
- * version : 1.8 or 1.9 : default to 1.8
+ * version : the version of the server, defaults to the latest version. Set version to `false` to enable dynamic cross version support.
  * favicon (optional) : the favicon to set, base64 encoded
  * customPackets (optional) : an object index by version/state/direction/name, see client_custom_packet for an example
  * errorHandler : A way to override the default error handler for client errors. A function that takes a Client and an error.
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -142,7 +142,7 @@ declare module 'minecraft-protocol' {
 		motd?: string
 		maxPlayers?: number
 		keepAlive?: boolean
-		version?: string
+		version?: string | false
 		favicon?: string
 		customPackets?: any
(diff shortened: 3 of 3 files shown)


rom1504:
> I have not been able to test this out properly

can you say what's a blocker for you to test it?

jojomatik:
> can you say what's a blocker for you to test it?

Mainly time constraints, if I find some time I'll try it out later. Just wanted to get the push done last night. Am I able to specify my branch as version in my `package.json`? I think I've seen people doing this before, if that works it should be a matter of minutes to test the functionality.

jojomatik:
@rom1504 have you yet found time to review this PR and take a look at the questions above? 

How long will this take approximately? I'd like to use this change in my project and wonder if it makes sense to temporarily include my fork as the package source instead of the npm package

src/server/ping.js
```diff
@@ -1,15 +1,23 @@
 const endianToggle = require('endian-toggle')
 
-module.exports = function (client, server, { beforePing = null }) {
+module.exports = function (client, server, { beforePing = null, version }) {
   client.once('ping_start', onPing)
   client.once('legacy_server_list_ping', onLegacyPing)
 
   function onPing () {
+    // Use client version if dynamic cross version support is enabled.
+    const responseVersion = (version === false || version === undefined)
```
rom1504, line 9: let's keep undefined meaning "default version" from https://github.com/PrismarineJS/node-minecraft-protocol/blob/78f68d1ff98e82e35f85678f96a9fa3cf2d6ba10/src/version.js#L4

rom1504:
hey, can you make undefined mean default version? it means that in the rest of the code

jojomatik:
> hey, can you make undefined mean default version? it means that in the rest of the code

I've force-pushed some changes, you can take a look at those [here](https://github.com/PrismarineJS/node-minecraft-protocol/compare/e3afa2bed7e3269108c9860b07369946bc19ac82..58ad3a311192d194f59fe647cb679321469d3da3).

`version: undefined` -> `1.18.2`
`version: "1.18.2"` -> `1.18.2`
`version: "1.16.5"` -> `1.16.5`
`version: undefined` or no version parameter -> `1.18.2`
`version: false` -> client version
`version: true` -> type error

May I critically question why the default version is locked at `1.17.1` in your snippet above? Should I really use that value for `version: undefined` instead of the most recent version?

rom1504:
> May I critically question why the default version is locked at 1.17.1 in your snippet above? Should I really use that value for version: undefined instead of the most recent version?

I think we should use that default version parameter *and* update it to 1.18.2

jojomatik:
> I think we should use that default version parameter _and_ update it to 1.18.2

Sounds reasonable.

We need the protocol version for the response though, should that one still be `server.mcversion.version`?

jojomatik:
Just checked: `1.18.2` already is the default since `[1.31.0](https://github.com/PrismarineJS/node-minecraft-protocol/releases/tag/1.31.0)` (https://github.com/PrismarineJS/node-minecraft-protocol/commit/9e94d9d9019250b2f4bfa098f937464cccf7e4b8)

jojomatik:
> I think we should use that default version parameter _and_ update it to 1.18.2

It seems like this requirement is already fully implemented:
https://github.com/PrismarineJS/node-minecraft-protocol/blob/9cff34efc0ac35fb3693a766133ca988ce40b93c/src/createServer.js#L30-L38

It takes the supplied version or the default version and uses `minecraft-data` to get the corresponding protocol version.

jojomatik:
Thanks for merging @rom1504. Really appreciate the fast response times :)

PR #982, opened 2022-04-11
Fix missing readable-stream types

IceTank:
This should fix an issue with typescript for people that do not have readable-stream installed globally

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
   "browser": "src/browser.js",
   "devDependencies": {
     "@types/node": "^17.0.4",
-    "@types/readable-stream": "^2.3.13",
     "espower-loader": "^1.0.0",
     "intelli-espower-loader": "^1.0.0",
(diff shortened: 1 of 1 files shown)


jojomatik:
This fix doesn't work for me as well unfortunately.

`devDependencies` are not considered for projects using this library.

I could however add `"@types/readable-stream": "^2.3.13",` as a dev dependency of my project, but that doesn't sound quite right to me. I never had to install some the types of a dependencies dependency before.

I'm not sure how, but I think it should be somehow possible to include that type.

IceTank:
I think adding it to normal dependencies should be fine

jojomatik:
Could you create a new release @rom1504, so that I can use this in the production version of my project?

rom1504:
You can make a release PR, then i just need to press merge

jojomatik:
> You can make a release PR, then i just need to press merge

#988

PR #983, opened 2022-04-12
Fallback version for dynamic versions

matthi4s:
This adds an optional `fallbackVersion` server option which is used as protocol version if dynamic version support is enabled (through `version: false`) but the client version is not supported. This allows to at least attempt communication with the client if they are on a different version, e.g. a new release version that isn't yet supported but might have a similar basic protocol.


jojomatik (commented):
I'd like to see the same/ consistent behaviour for ping. Could you implement the same behaviour there?

PR #984, opened 2022-04-12
Export defaultVersion

matthi4s:
This adds the `defaultVersion` to the exported fields of this module. This can be useful to set the version or the fallback version (see https://github.com/PrismarineJS/node-minecraft-protocol/pull/983) to the current default version. I've also replaced the type definition of `supportedVersions` with the correct type.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -325,6 +325,10 @@ The minecraft protocol states.
 
 The supported minecraft versions.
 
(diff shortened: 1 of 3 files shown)


jojomatik (approved):
Looks good to me, I'd also like to use the `defaultVersion` in my project.

PR #985, opened 2022-04-12
Added optional callback to individually decide about verification

matthi4s:
This adds an optional `shouldVerifyClient` server option to specify a callback function that's called in the login process to individually decide for each client if the online mode verification is required or not. There already is an array of username exceptions `server.onlineModeExceptions` but this doesn't allow deciding based on other factors like the used server address.

### DIFF
--- a/src/server/login.js
+++ b/src/server/login.js
@@ -32,7 +32,7 @@ module.exports = function (client, server, options) {
     if (typeof options.shouldVerifyClient === 'function') {
       needToVerify = await options.shouldVerifyClient(client)
     }
-    if (typeof needToVerify !== 'function') {
+    if (typeof needToVerify !== 'boolean') {
       const isException = !!server.onlineModeExceptions[client.username.toLowerCase()]
       needToVerify = (onlineMode && !isException) || (!onlineMode && isException)
     }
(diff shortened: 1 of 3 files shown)


src/server/login.js
```diff
@@ -14,6 +14,7 @@ module.exports = function (client, server, options) {
   } = options
 
   let serverId
+  let needToVerify
```
u9g, line 17: this shouldn't be a top level variable, instead it should be decided on a case-by-case basis (inside the onLogin function) based on the exceptions array and the result of the login function in that specific case

matthi4s: The result is required in `onEncryptionKeyResponse()` and `loginClient()` as well during the login process of a single client. Calling the function three times per login is inefficient and probably also unexpected. This is only the top level inside the function that's called individually for each client, so it's still a case-by-case basis for each client.

rom1504:
can you add in the doc the use case of this?
It's not immediately obvious to me

src/server/login.js
rom1504: I agree

rom1504:
I don't know what's the use case
re-open if you want to finish it

PR #986, opened 2022-04-12
Allow false as beforePing callback result to ignore pings

matthi4s:
This adds the ability to return `false` in the `beforePing` callback or passing `false` to the callback in the third argument to ignore pings by terminating the connection. This can be used to block pings to invalid or unwanted server addresses.


u9g (approved):
LGTM

PR #988, opened 2022-04-13
1.34.0

jojomatik:


jojomatik:
Sorry to bother you @rom1504, but could you please merge this?

IceTank:
LGTM

jojomatik:
Well that didn't go as intended. Seems like the commit has to be named `Release x`... Sorry about that :grimacing:

extremeheat:
You can make another release

rom1504:
I fixed it

jojomatik:
> I fixed it

Thanks @rom1504 :) 

In case you're interested in automating the release process: 

Take a look at semantic-release and the angular commit convention. I use it in all my projects and even though it took some time to get used to it, I think it is totally worth it. It basically creates new minor releases for commits that start with `feat: ` and patch releases for commits that start with `fix: ` (and much more) and is highly configurable. It can also create release notes and update changelog files.

You can take a look at one of my repos or hit me up for any input on this :)

rom1504:
> Take a look at semantic-release and the angular commit convention. I use it in all my projects and even though it took some time to get used to it, I think it is totally worth it. It basically creates new minor releases for commits that start with feat: and patch releases for commits that start with fix: (and much more) and is highly configurable. It can also create release notes and update changelog files.

I saw that a while back and I am not convinced.

The main problem of this setup is it forces *every* contributor to name their commits that way.
I don't know how to enforce that.

jojomatik:
> I saw that a while back and I am not convinced.

Yeah it has some issues, and it took some time for me to get used to, but I believe it's worth it.

> The main problem of this setup is it forces _every_ contributor to name their commits that way. I don't know how to enforce that.

As far as I can tell, you squash the commits from PRs anyway. GitHub gives you the option to adjust the commit message in that case. Depending on the effort you'd like to put in you could just add the relevant commit type and continue merging.

rom1504:
Yeah but that means every person merging commit need to do that

rom1504:
What does it do if you don't specify the commit type by error ?

jojomatik:
> What does it do if you don't specify the commit type by error ?

Basically nothing.

It tries to find a commit that starts with `feat:`  or `fix:`  between the last tag and the most recent commit. If it doesn't find one, it does nothing. No commit type should equal commit types like `docs: ` that won't trigger a release (although that is configurable).

If a commit that provides a new feature is accidentally merged without the correct type, you can still create a release commit manually and let the current actions setup handle it.

PR #997, opened 2022-05-06
Use offline mode as default authentication, fallback to offline mode if invalid option.

Kashalls:
This PR is meant to add into https://github.com/PrismarineJS/node-minecraft-protocol/pull/958 and any additional input can be used to better this.

PR #998, opened 2022-05-09
Use offline mode as default authentication, fallback to offline mode if invalid option.

Kashalls:
As requested by https://github.com/PrismarineJS/node-minecraft-protocol/pull/997#issuecomment-1121661464 . It has been the collective agreement by most contributors to default auth to offline mode unless specified by the end user. This is to prevent unnecessary logins to xbox live so a user doesn't get banned faster than needed.

### DIFF
--- a/src/createClient.js
+++ b/src/createClient.js
@@ -38,7 +38,6 @@ function createClient (options) {
     case 'mojang':
       auth(client, options)
       break
-    case undefined:
     case 'microsoft':
       microsoftAuth.authenticate(client, options)
       break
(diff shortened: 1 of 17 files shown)


rom1504:
please undo changes to examples

src/createClient.js
```diff
@@ -34,10 +34,19 @@ function createClient (options) {
...
+    case 'mojang':
+      console.warn('[deprecated] mojang auth servers no longer accept mojang accounts to login. convert your account.\nhttps://help.minecraft.net/hc/en-us/articles/4403181904525-How-to-Migrate-Your-Mojang-Account-to-a-Microsoft-Account')
+      auth(client, options)
+      break
+    case 'microsoft':
+      microsoftAuth.authenticate(client, options)
+      break
+    case 'offline':
+    default:
+      client.username = options.username
```
rom1504, line 47: why that change?

Kashalls: > why that change?

We need to set a username for the login_start packet in setProtocol.

PR #1004, opened 2022-06-06
Ipc connection

IceTank:


docs/API.md
```diff
@@ -31,6 +31,7 @@ automatically logged in and validated against mojang's auth.
  * hideErrors : do not display errors, default to false
  * agent : a http agent that can be used to set proxy settings for yggdrasil authentication confirmation (see proxy-agent on npm)
  * validateChannelProtocol (optional) : whether or not to enable protocol validation for custom protocols using plugin channels for the connected clients. Defaults to true
+ * socketType (optional) : either `tcp` or `ipc`. Switches from a tcp connection to a ipc socket connection (or named pipes on windows). With the `ipc` option `host` becomes the path off the ipc connection on the local filesystem. Example: `\\.\pipe\minecraft-ipc` (Windows) `/tmp/minecraft-ipc.sock` (unix based systems). See the ipcConnection example for an example.
```
rom1504, line 34: can't we just make port optional ?

IceTank: Yeah sure

IceTank: Well actually that won't work. `createServer` already has port as an optional argument. If port is not set it defaults to 25565. Making it do something else would be a breaking change.

rom1504:
can you rebase

rom1504:
please re-open if you want to finish it

PR #1007, opened 2022-06-08
Implement Realm joining

LucienHH:
Currently attempts to join the Realm as expected but can't fully be sure until 1.19 support as Realms are always on the latest version. Opening this PR now for review on implementation.


examples/client_realms/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "node-minecraft-protocol-example",
+  "version": "0.0.0",
+  "private": true,
+  "dependencies": {
```
andriycraft, line 5: Remove
```suggestion
  {
     "name": "node-minecraft-protocol-example",
     "version": "0.0.0",
      "private": true,
      "description": "A node-minecraft-protocol example"
```

LucienHH: This is the case in other examples so believe it should stay as is.

andriycraft: ok

rom1504:
is it ready ? can you fix conflicts ?

LucienHH:
Done

examples/client_realms/client_realms.js
```diff
@@ -0,0 +1,34 @@
...
+  auth: 'microsoft' // This option must be present and set to 'microsoft' to  join a Realm.
+})
+
+client.on('connect', function () {
+  console.info('connected')
+})
+client.on('disconnect', function (packet) {
+  console.log('disconnected: ' + packet.reason)
+})
+client.on('chat', function (packet) {
```
extremeheat, line 26: this is outdated now, maybe add `realms` as an option to the existing chat example instead ? This would avoid needing to keep a separate chat example up to date (especially as 1.19.2 and 1.19.3 will change chat further)

LucienHH: Ah ok, I think it'd be better to just remove the on chat event as the example's only really here to show the available options for connecting to a Realm. If we merge it with the current chat example it could get cumbersome with the host/port options etc.

rom1504:
does this work, should we merge?

extremeheat (approved):

rom1504:
lot of conflicts now I'm afraid, let's fix them

LucienHH:
Conflixts fixed inn https://github.com/PrismarineJS/node-minecraft-protocol/pull/1056

PR #1008, opened 2022-06-15
Fix issues with servers sending un-escaped newlines

rtm516:
If a server sent a response like the below it would cause an `Uncaught SyntaxError: Unexpected token x in JSON at position x` error. Escaping the newlines fixes this issue and more accurately follows the vanilla clients parsing of it.
```
{"version": {"name":"§9gameshield.gg", "protocol":-1}, "description":"§7Cannot connect to §951.222.139.92\n§7For more information visit §9gameshield.gg","text":"§9GameShield\n\n§7The server you are trying to connect to is not found, or is not responding.\nPlease try again later.\n\nFor more information visit §9gameshield.gg"}
```


src/ping.js
```diff
@@ -38,6 +38,7 @@ function ping (options) {
       reject(err)
     })
     client.once('server_info', function (packet) {
+      packet.response = packet.response.replace(/\n/g, '\\n')
```
rom1504, line 41: Can you put it into a function like parseLenientJson ?

Apparently this is not json

rom1504:
can you fix lint / ci ?

rtm516:
> can you fix lint / ci ?

Doesnt look like there is an issue that was introduced here, have you tried re-triggering it?

rom1504 (changes requested):
fix ci please

rom1504:
doesn't make sense to me

let's instead maybe add a try catch to avoid crashing if the server sends invalid content

PR #1018, opened 2022-07-19
add custom minecraft type `varlong` which aliases to varint

rob9315:
the purpose is to mark fields which are actually `varlong`s in the minecraft protocol as such in [`minecraft-data`](https://github.com/PrismarineJS/minecraft-data) to allow for more accurate implementations in lower level languages in which the specific number type actually matters.


rom1504:
Merged that but actually might need more thoughts

rob9315:
should i implement a more correct varlong here?
can then be moved to node-protodef if required

PR #1020, opened 2022-07-21
[1.19] fix tests and library session code

rob9315:
blocked by [mcd 1.19 features pr](https://github.com/PrismarineJS/minecraft-data/pull/599)

- tests now work again with the new chat packet names
- added todo's in places where further work is needed like
  - missing signature encryption
  - adding version check for benchmark test chat packet
- adds varlong and command_node packetTest types

### DIFF
--- a/src/client/encrypt.js
+++ b/src/client/encrypt.js
@@ -49,7 +49,6 @@ module.exports = function (client, options) {
         const encryptedVerifyTokenBuffer = crypto.publicEncrypt({ key: pubKey, padding: crypto.constants.RSA_PKCS1_PADDING }, packet.verifyToken)
 
         if (mcData.supportFeature('signatureEncryption')) {
-          // 1.19+
           // todo: add signature encryption
           // starting 1.19.1 we will not be able to join
           // the default server configuration without it
@@ -61,7 +60,6 @@ module.exports = function (client, options) {
             }
           })
         } else {
-          // pre 1.19
           client.write('encryption_begin', {
             sharedSecret: encryptedSharedSecretBuffer,
             verifyToken: encryptedVerifyTokenBuffer
--- a/src/client/encrypt.js
+++ b/src/client/encrypt.js
@@ -42,13 +42,31 @@ module.exports = function (client, options) {
       }
 
       function sendEncryptionKeyResponse () {
+        const mcData = require('minecraft-data')(client.version)
+
         const pubKey = mcPubKeyToPem(packet.publicKey)
         const encryptedSharedSecretBuffer = crypto.publicEncrypt({ key: pubKey, padding: crypto.constants.RSA_PKCS1_PADDING }, sharedSecret)
         const encryptedVerifyTokenBuffer = crypto.publicEncrypt({ key: pubKey, padding: crypto.constants.RSA_PKCS1_PADDING }, packet.verifyToken)
-        client.write('encryption_begin', {
-          sharedSecret: encryptedSharedSecretBuffer,
-          verifyToken: encryptedVerifyTokenBuffer
-        })
+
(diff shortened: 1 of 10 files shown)


test/packetTest.js
```diff
@@ -176,7 +177,30 @@ const values = {
   tags: [{ tagName: 'hi', entries: [1, 2, 3, 4, 5] }],
   ingredient: [slotValue],
   particleData: null,
-  chunkBlockEntity: { x: 10, y: 11, z: 12, type: 25 }
+  chunkBlockEntity: { x: 10, y: 11, z: 12, type: 25 },
+  command_node: {
```
rom1504, line 181: I guess that works but it's not a very general fix

Would be much better to just propagate the recursion to the custom types

src/client/encrypt.js
```diff
@@ -42,13 +42,31 @@ module.exports = function (client, options) {
...
         const pubKey = mcPubKeyToPem(packet.publicKey)
         const encryptedSharedSecretBuffer = crypto.publicEncrypt({ key: pubKey, padding: crypto.constants.RSA_PKCS1_PADDING }, sharedSecret)
         const encryptedVerifyTokenBuffer = crypto.publicEncrypt({ key: pubKey, padding: crypto.constants.RSA_PKCS1_PADDING }, packet.verifyToken)
-        client.write('encryption_begin', {
-          sharedSecret: encryptedSharedSecretBuffer,
-          verifyToken: encryptedVerifyTokenBuffer
-        })
+
+        if (mcData.supportFeature('signatureEncryption')) {
+          // 1.19+
```
rom1504, line 52: I think specifying the version range next to each support feature call is not necessary

src/server/login.js
```diff
@@ -55,18 +55,50 @@ module.exports = function (client, server, options) {
...
+    let signature
+
+    if (mcData.supportFeature('signatureEncryption')) {
+      // 1.19+
+      if (packet.hasVerifyToken) {
+        packetVerifyToken = packet.crypto.verifyToken
+      } else {
+        signature = packet.crypto
+      }
+    } else {
```
rom1504, line 70: I think it would be better to avoid else here and just have a if signatureNotEncrypted
It's not that there was no feature before, it's that there was a different feature

rom1504: Optional though, not absolutely needed to do

rom1504:
Congrats on fixing this, will look in more details later

rom1504:
Ah CI is not running

rom1504:
Just re-open towards master

src/server/login.js
```diff
@@ -55,18 +55,50 @@ module.exports = function (client, server, options) {
...
+      } catch (e) {
         client.end('DidNotEncryptVerifyTokenProperly')
         return
       }
-      sharedSecret = crypto.privateDecrypt({ key: server.serverKey.exportKey(), padding: crypto.constants.RSA_PKCS1_PADDING }, packet.sharedSecret)
-    } catch (e) {
-      client.end('DidNotEncryptVerifyTokenProperly')
+    } else {
+      // todo: signature encryption
+      client.end('signature encryption not implemented')
```
rom1504, line 97: So this is missing then ?

test/benchmark.js
```diff
@@ -7,7 +7,8 @@ const states = mc.states
 
 const testDataWrite = [
   { name: 'keep_alive', params: { keepAliveId: 957759560 } },
-  { name: 'chat', params: { message: '<Bob> Hello World!' } },
+  // TODO: 1.19+ `chat` -> `player_chat` feature toggle
```
rom1504, line 10: Would be better to fix that

PR #1022, opened 2022-07-21
Fix unhandled promise rejection on ms auth

IceTank:
The authenticate function that does ms auth returns a promise that can reject on error. If this occurs in your code the node process will exit due to an unhandledRejection error.
This pull requests fixes this and instead emits a error event on the client.

### DIFF
--- a/src/createClient.js
+++ b/src/createClient.js
@@ -43,7 +43,7 @@ function createClient (options) {
         auth(client, options)
         break
       case 'microsoft':
-        microsoftAuth.authenticate(client, options)
+        microsoftAuth.authenticate(client, options).catch((err) => client.emit('error', err))
         break
       case 'offline':
       default:


rom1504:
can you please resolve conflicts ?

lelboatz:
Looks good to me, would be great if this could be merged.

src/createClient.js
```diff
@@ -43,7 +43,7 @@ function createClient (options) {
         auth(client, options)
         break
       case 'microsoft':
-        microsoftAuth.authenticate(client, options)
+        microsoftAuth.authenticate(client, options).catch((err) => client.emit('error', err))
```
rom1504, line 46: what about other auth methods? why is it different here ?

IceTank: With the old mojang login the auth function was synchronous. Whenever it had an error it would emit the error event on the bot. The new microsoft authenticate function is async. When it has an error it rejects the promise. This would cause a unhandled promise rejection warning in older node versions. But new node versions treat unhandled promise rejections as process errors and exit the node process. This change proberly handles the promise returned by the authenticate function.

Issue #1025, opened 2022-07-27
Custom Authentication Function


vlouboos:
**Is your feature request related to a problem? Please describe.**
I can't join some servers which use non-vanilla auth

**Describe the solution you'd like**
Add an interface to use own auth-server, like custom auth-lib.
Or just simply forward the authentication-packet(hand_shake, login_start)

Kashalls:
What authentication are you talking about? Are you referring to https://github.com/PrismarineJS/node-minecraft-protocol/pull/572 ?

vlouboos:
Including but also others, like some mc-launchers build a local server for auth when logging in servers, clients call the servers before sending the hand_shake packet, the game profile is returned from the server.
The most effective way is to forward the hand_shake and login_start packet to the target server in order to prevent unsupported auth.

vlouboos:
Or another way, use 'authlib.jar' for auth, then give an interface for us to custom using authlib

Kashalls:
I'm against directly putting it minecraft-protocol but I can look into adding an interface for you to do your own authentication. It can be as simple as making auth available as a function so that if you don't set Microsoft or offline, you can handle it your way. Let me look into it.

rob9315:
> Add an interface to use own auth-server
 
i can get behind this, there should also be an easy option for using yggdrasil-compatible ('mojang') auth servers by setting the custom /auth endpoint url.

> Another way, use 'authlib.jar' for auth, then give an interface for us to custom using authlib

this will not turn into a wrapper of authlib, it's been completely java-free so far and will stay that way

Kashalls:
>  > Add an interface to use own auth-server
>  
> i can get behind this, there should also be an easy option for using yggdrasil-compatible ('mojang') auth servers by setting the custom /auth endpoint url.
> 
> > Another way, use 'authlib.jar' for auth, then give an interface for us to custom using authlib
> 
> this will not turn into a wrapper of authlib, it's been completely java-free so far and will stay that way

You can set session server in the client options to change the session server.

rob9315:
no, that doesn't work because it's not forwarded to yggdrasil

rob9315:
*yes, it is*, but there is no option for servers

Kashalls:
https://github.com/PrismarineJS/node-minecraft-protocol/commit/60379eb7d2e102c180b425178f8e525fe0417ad6 Solves original issue by exposing an auth function for non-official auth services.

PR #1026, opened 2022-07-27
Provide interface for using not at all supported alternative accounts.

Kashalls:
This provides an interface for the people desperate enough not wanting to spend money on a good game (that was until Microsoft got their grubby hands on it) to use alternative services to allow game play. I cannot heed this enough that using these alternative services may or may not get you into trouble. Do not hold me accountable.

This provides a solution for: https://github.com/PrismarineJS/mineflayer/issues/1116, https://github.com/PrismarineJS/mineflayer/issues/1331, https://github.com/PrismarineJS/mineflayer/issues/790, https://github.com/PrismarineJS/node-minecraft-protocol/issues/736, https://github.com/PrismarineJS/node-minecraft-protocol/issues/1025 and a bunch more that I am not bothering to look for.

This proposal is to check if the user has specified a function for `options.auth` rather than `mojang` | `microsoft`. If it is a function, we let that function handle setting the username, accessToken... making any related api calls. However, when it comes to joining, if they are not using the official session server, they will have to create a local proxy to convert the expected format that usually is sent to mojang's session server to their custom auth server.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -93,7 +93,7 @@ Returns a `Client` instance and perform login.
 `options` is an object containing the properties :
  * username
  * port : default to 25565
- * auth : the type of account to use, either `microsoft`, `mojang`, or `offline`. default to 'offline'
+ * auth : the type of account to use, either `microsoft`, `mojang`, `offline` or `function (client, options) => void`. defaults to 'offline'.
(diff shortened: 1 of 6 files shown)


rom1504:
can you rebase on master

PR #1027, opened 2022-07-28
1.19.0 support

extremeheat:
With chat signing support

**Todo**
- Verify logic is correct for offline mode

**Adds**
- new client.signMessage and client.verifyMessage utility methods
- binary concatenation function (could possibly fit  into protodef)


src/client/microsoftAuth.js
```diff
@@ -34,6 +37,19 @@ async function authenticate (client, options) {
   }
   client.session = session
   client.username = profile.name
+  client.profileKeys = {}
```
extremeheat, line 40: Delete

src/server/login.js
```diff
@@ -27,9 +29,39 @@ module.exports = function (client, server, options) {
...
+        if (!crypto.verify('RSA-SHA1', Buffer.from(signable, 'utf8'), mojangPubicKey, packet.signature.signature)) {
+          client.end('multiplayer.disconnect.invalid_public_key_signature')
+          return
+        }
+        client.profileKeys = { public: publicKey, publicPEM }
+      } catch (err) {
+        client.end('multiplayer.disconnect.invalid_public_key')
+        return
+      }
+    } else if (options.enforceSecureProfile) {
```
extremeheat, line 60: Dupe condition of above L38

src/client/microsoftAuth.js
```diff
@@ -34,6 +37,19 @@ async function authenticate (client, options) {
...
+
+  if (certificates) {
+    client.profileKeys = {
+      publicPEM: certificates.keyPair.publicKey,
```
extremeheat, line 44: Maybe some of this transformation could be handled in prismarine-auth

Kashalls: Crypto should be done in its own class/util in nmp.

extremeheat: Will move this to pauth as it should not be returning unabstracted 1:1 web request data

extremeheat:
Need to handle offline mode on server also. I'm assuming server disables chat signing for the client if the server is in offline mode, maybe through setting hasVerifyToken to true. Need to investigate that, cc @rob9315

rom1504:
this looks almost ready now

extremeheat:
Passing. Should be possible to add some sort of test for the chat signing, but it's a bit tricky because that would involve a bit of boilerplate and patching mojang public key constant.

xtexChooser:
Is there any way to use the 1.19 support directly?

I knew this is unstable, but I need 1.19.

Thanks.

vincss:
Hi !
When I try to login using this version from https://github.com/vincss/mcsleepingserverstarter, I got this error message : 
image
 And in my logs : 
Chunk size is 848 but only 831 was read ; partial packet : {"name":"login_start","params":{"username":"Vincouille","signature":{"timestamp":[386,-1950364232],"publicKey":{"type":"Buffer","data":[48,130,1,34,48,13,6,9,42,134,72,134
Hope it helps !

extremeheat:
Are your dependencies up to date and are you using 1.19.0 (not 1.19.1) ? Sending an updated login packet allows my client to connect to an empty world

image

vincss:
> @vincss
> 
> Are your dependencies up to date and are you using 1.19.0 (not 1.19.1) ? Sending an updated login packet allows my client to connect to an empty world
> 
> image

I was on a 1.19.2, but indeed in 1.19 it's working better. Thanks.

rom1504:
how far is this from ready ?

rom1504:
ok looked at the code

looks good to me

I would appreciate if a few more people could try to boot it up and check everything seems to work for them

SpencerBurgess:
This doesn't update the README.md to include 1.19.0. Is that intentional or is that added with the next release PR?

rom1504:
No, please add it

PR #1033, opened 2022-08-18
Fix type definition breaking typescript projects

IceTank:
This option should be optional as it is treated as optional in the code.


src/index.d.ts
```diff
@@ -180,7 +180,7 @@ declare module 'minecraft-protocol' {
 		validateChannelProtocol?: boolean
 		// 1.19+
 		// Require connecting clients to have chat signing support enabled
-		enforceSecureProfile: boolean
```
extremeheat, line 183: Need to do the same for Client

PR #1047, opened 2022-12-07
move chat implementation to src/

u9g:


src/client/play.js
```diff
@@ -5,12 +5,14 @@ const concat = require('../transforms/binaryStream').concat
...
 
+  const EMPTY_BUFFER = Buffer.from([])
+
   function onLogin (packet) {
     client.state = states.PLAY
     client.uuid = packet.uuid
     client.username = packet.username
     client.signMessage = (message, timestamp, salt = 0) => {
-      if (!client.profileKeys) throw Error("Can't sign message without profile keys, please set valid auth mode")
+      if (!client.profileKeys) return EMPTY_BUFFER
```
extremeheat, line 15: this should be an exception as no signature is returned

src/client/chat.js
```diff
@@ -0,0 +1,46 @@
...
+
+  function onChat (packet, isMessageFromServer = false) {
+    let message = packet.message ?? packet.unsignedChatContent ?? packet.signedChatContent
+    if (isMessageFromServer) message = packet.content
+
+    let verified = isMessageFromServer ? null : false // default to false if this is a system chat
+    if (!isMessageFromServer && client.profileKeys != null && packet.signature) {
+      verified = client.verifyMessage(players[packet.senderUuid].publicKey, packet)
+    }
+    client.emit('chat_received', { verified, message, isMessageFromServer })
```
extremeheat, line 15: Why not a chatMessage and serverMessage event ? In mineflayer we don't handle both in one function and the event emitted has a type in it. If chat abstraction is moving to nmp just copy over the existing mineflayer implementation / events

src/client/play.js
u9g: Fixed in latest commit

rom1504:
done in another pr

PR #1049, opened 2022-12-16
Implement 1.19.1 - 1.19.3

### LINKED ISSUE #1037: 1.19.1-1.19.3
Support for 1.19.1-1.19.2

Any updates on this issue? no pressure or anything just wondering where your at with it!

Thank you, PandaUprising

### LINKED ISSUE #1046: Implement 1.19.2 chat signing
chat_signin

From @Kashalls

frej4189:
Depends on https://github.com/PrismarineJS/minecraft-data/pull/661

Chat signing is fully implemented for all versions 1.19 through 1.19.3.

This closes #1046 and closes #1037

### DIFF
--- a/docs/HISTORY.md
+++ b/docs/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 1.37.0
+* 1.19.1/2 signed chat support 
+
 ## 1.36.2
 * Throw error on minecraft-data protocol version mismatch (#1044)
 * Make "fakeHost" option working
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "minecraft-protocol",
-  "version": "1.36.2",
+  "version": "1.37.0",
   "description": "Parse and serialize minecraft packets, plus authentication and encryption.",
   "main": "src/index.js",
   "types": "src/index.d.ts",


src/client/profileKeys.js
```diff
@@ -0,0 +1,69 @@
...
+    raw = raw.substring(65)
+  }
+
+  return crypto.createPublicKey(pem + suf)
+}
+
+module.exports = function (client, options) {
+  if (!options.accessToken) throw new Error('Invalid user access token')
+  return new Promise((resolve, reject) => {
+    const req = Https.request('https://api.minecraftservices.com/player/certificates', {
```
extremeheat, line 31: This is already handled in prismarine-auth

frej4189: Alright. My bad. It didn't seem to work when I ran online tests. It will likely prompt a few changes but removing that file won't affect the signing aspect.

src/client/play.js
```diff
@@ -5,15 +5,37 @@ const concat = require('../transforms/binaryStream').concat
...
     client.uuid = packet.uuid
     client.username = packet.username
-    client.signMessage = (message, timestamp, salt = 0) => {
+    client.signMessage = (message, timestamp, salt = 0n, isCommand, preview) => {
       if (!client.profileKeys) throw Error("Can't sign message without profile keys, please set valid auth mode")
-      const signable = concat('i64', salt, 'UUID', client.uuid, 'i64',
-        timestamp / 1000n, 'pstring', JSON.stringify({ text: message }))
-      return crypto.sign('RSA-SHA256', signable, client.profileKeys.private)
+      if (mcData.supportFeature('chainedSignature')) { // 1.19.1/1.19.2
+        const hashable = crypto.createHash('sha256').update(concat('i64', salt, 'i64', timestamp / 1000n, 'pstring', message, 'i8', 70))
```
extremeheat, line 17: What is the 70 referring to here? And is this verified by the server ?

src/client/play.js
```diff
@@ -5,15 +5,37 @@ const concat = require('../transforms/binaryStream').concat
...
   client.once('success', onLogin)
 
+  const mcData = require('minecraft-data')(client.version)
+
   function onLogin (packet) {
     client.state = states.PLAY
     client.uuid = packet.uuid
     client.username = packet.username
-    client.signMessage = (message, timestamp, salt = 0) => {
+    client.signMessage = (message, timestamp, salt = 0n, isCommand, preview) => {
```
extremeheat, line 14: put new fields at the end to not be breaking to current API

src/client/play.js
```diff
@@ -5,15 +5,37 @@ const concat = require('../transforms/binaryStream').concat
...
+        if (!!client.lastSignature && !isCommand) proto.unshift('buffer', client.lastSignature)
+        const signable = concat(...proto)
+        const signed = crypto.sign('RSA-SHA256', signable, client.profileKeys.private)
+        if (!isCommand) client.lastSignature = signed
+        return signed
+      } else if (mcData.supportFeature('sessionSignature')) { // 1.19.3
+        if (!client.sessionUUID) throw Error("Can't sign message before initializing chat session")
+        if (!client.sessionIndex) client.sessionIndex = 0
+        const length = Buffer.byteLength(message, 'utf8')
+        const signable = concat('i32', 1, 'UUID', client.uuid, 'UUID', client.sessionUUID, 'i32', client.sessionIndex, 'i64', salt, 'i64', timestamp / 1000n, 'i32', length, 'pstring', message, 'i32', 0)
```
extremeheat, line 30: explain the 1 and 0 at the end?

extremeheat:
Will the message log be in this PR?

src/client/play.js
frej4189: Client adds an unsigned byte with the value of 70 before hashing. Server does the same thing when verifying. No clue why.

src/client/play.js
frej4189: Same thing as with the 70 in proto 760. For some reason those values are padded in before signing (and before verifying). They must be present or the client will get kicked for invalid or missing signature

src/client/play.js
frej4189: Is that not what I did?

extremeheat: Ah, this is different from what you put in the types

frej4189: I see. My bad.

extremeheat:
This does partially implement client to server chat signing, but it's still missing many things. Some things that come off top of my head:

- Server side processing and validation of client sent messages
- server side validation of secure profile on 1.19.1+
- client side validation
- handling for delete message packet, clientbound chat header
- serverboundack
- rolling memory message logs

Tests can be implemented inside here just like the additions in 1.19 to verify this is all working properly

src/client/play.js
frej4189: Gonna have to add: Trailing 0 is actually the length of the last seen messages array. client.signMessage currently does not account for chaining at all (as the server doesn't seem to mind that the client does not send any acknowledgements), although it should definitely be implemented to mimic the vanilla client

rom1504:
all methods that are new and exposed should be:
* documented
* throw an error for versions where they don't make sense

ideally we would make it easy for higher level API to hide away the version difference

I see many such new methods

src/server/chat.js
```diff
@@ -0,0 +1,292 @@
...
+      raise('multiplayer.disconnect.invalid_public_key_signature')
+      return
+    }
+
+    if (!client.profileKeys) {
+      client.verifyMessage = (packet, acknowledgements) => {
+        const length = Buffer.byteLength(packet.message, 'utf8')
+        const previousMessages = acknowledgements.length > 0 ? ['i32', acknowledgements.length, 'buffer', Buffer.concat(...acknowledgements)] : ['i32', 0]
+
+        const signable = concat('i32', 1, 'UUID', client.uuid, 'UUID', client.session.uuid, 'i32', client.session.index, 'i64', packet.salt, 'i64', packet.timestamp / 1000n, 'i32', length, 'pstring', packet.message, ...previousMessages)
```
rom1504, line 72: I see this kind of packaging at many places
is it really different every time ?

rom1504:
Could you summarize what are the new features from a high level point of view ?
What they bring  / how they should be used ?

src/client/chat.js
```diff
@@ -0,0 +1,538 @@
...
+            packetData.signedPreview = true
+            const preview = await fetchPreview(message)
+            const acknowledgements = client.refreshAcknowledgements()
+            const timestamp = BigInt(Date.now())
+            packetData.timestamp = timestamp
+            packetData.signature = client.signMessage(message, timestamp, salt, acknowledgements, false, preview)
+            packetData.previousMessages = client.chat_log ? client.chat_log.acknowledgements : []
+            packetData.lastMessage = client.chat_log ? client.chat_log.lastUntracked : undefined
+          } catch {
+            // Discard message
```
extremeheat, line 116: Should not silently fail

src/client/chat.js
```diff
@@ -0,0 +1,538 @@
...
+  client.on('player_chat', handlePlayerChat)
+
+  client.on('system_chat', handleSystemChat)
+
+  client.on('hide_message', handleHideChat)
+
+  client.on('message_header', handleChatHeader)
+
+  client.on('should_display_chat_preview', packet => {
+    client.useChatPreview = packet.should_display_chat_preview
```
extremeheat, line 242: this is not the same thing as useChatPreview, that's sent in server_data packet

src/client/chat.js
```diff
@@ -0,0 +1,538 @@
...
+  })
+
+  client.on('server_data', packet => {
+    client.useChatPreview = packet.previewsChat
+    client.signedChat = packet.enforcesSecureChat
+  })
+
+  function handleChatPreview (packet) {
+    if (packet.queryId === client.pendingPreview.id) {
```
extremeheat, line 251: the previews are not necessarily synchronously ordered, for example you might send two previews # 1 and # 2 and the server can respond in the order of # 2 and # 1

src/client/chat.js
```diff
@@ -0,0 +1,538 @@
...
+          delete client.players[player.UUID]
+        })
+      }
+    }
+  }
+
+  function handleHideChat (packet) {
+    if (mcData.supportFeature('acknowledgeUntracked')) {
+      const signature = packet.signature || resolveSignature(packet.id)
+      if (signature) removeMessage(packet.signature)
```
extremeheat, line 301: I don't think it's removed from the lastSeen messages array, in 1.19.2 it appears to run [`removeFromDelayedMessageQueue`](https://github.dev/extremeheat/extracted_minecraft_data/blob/client1.19.1/client/net/minecraft/client/multiplayer/ClientPacketListener.java#L783) and only removes from inbound message throttle

src/client/chat.js
```diff
@@ -0,0 +1,538 @@
...
+
+  client.on('message_header', handleChatHeader)
+
+  client.on('should_display_chat_preview', packet => {
+    client.useChatPreview = packet.should_display_chat_preview
+  })
+
+  client.on('server_data', packet => {
+    client.useChatPreview = packet.previewsChat
+    client.signedChat = packet.enforcesSecureChat
```
extremeheat, line 247: enforcesSecureChat does not exist on 1.19.2, it's secure profile

src/client/chat.js
```diff
@@ -0,0 +1,538 @@
...
+        new: 0,
+        untracked: 0,
+        acknowledgements: []
+      }
+    }
+
+    const id = client.chat_log.new++
+    if (id >= 1024) client.chat_log.old++
+
+    client.chat_log.log[id % 1024] = { type: log.SYSTEM_MESSAGE, packet, timestamp: Date.now() }
```
extremeheat, line 451: One thing to keep in mind using `id % 1024` is this results in index wrapping. The ordering in chat_log no longer implies leftward for older and rightward for newer, which can make lookups more complex as you need to sort by timestamp.

src/client/microsoftAuth.js
```diff
@@ -34,7 +34,9 @@ async function authenticate (client, options) {
     availableProfile: [profile]
   }
   Object.assign(client, certificates)
+  client.profileKeys.expireTime = client.profileKeys.expiresOn.getTime()
   client.session = session
+  client.uuid = profile.id.replace(/(\w{8})(\w{4})(\w{4})(\w{4})(\w{12})/, '$1-$2-$3-$4-$5')
```
extremeheat, line 39: why are the dashes needed?

src/client/microsoftAuth.js
```diff
@@ -34,7 +34,9 @@ async function authenticate (client, options) {
     availableProfile: [profile]
   }
   Object.assign(client, certificates)
+  client.profileKeys.expireTime = client.profileKeys.expiresOn.getTime()
```
extremeheat, line 37: should be fixed in prismarine-auth

src/client/setProtocol.js
```diff
@@ -24,15 +25,17 @@ module.exports = function (client, options) {
...
-        signature: client.profileKeys
+        playerUUID: mcData.supportFeature('signatureOnLogin') ? client.uuid : undefined,
+        signature: (mcData.supportFeature('signatureOnLogin') && client.profileKeys)
           ? {
-              timestamp: BigInt(client.profileKeys.expiresOn.getTime()), // should probably be called "expireTime"
+              timestamp: BigInt(client.profileKeys.expireTime),
               publicKey: client.profileKeys.publicDER,
-              signature: client.profileKeys.signature
+              signature: mcData.supportFeature('profileKeySignatureV2') ? client.profileKeys.signatureV2 : client.profileKeys.signature
             }
```
extremeheat, line 37: missing `profileId` field in login_start

src/server/chat.js
```diff
@@ -0,0 +1,292 @@
...
+    })
+  }
+
+  function handleChatSession (packet) {
+    client.session = {
+      uuid: packet.sessionUUID,
+      index: 0
+    }
+
+    const publicKey = crypto.createPublicKey({ key: packet.publicKey, format: 'der', type: 'spki' })
```
extremeheat, line 58: Pretty sure the public key should already be in DER format, 3 lines here should not be needed

src/server/chat.js
```diff
@@ -0,0 +1,292 @@
...
+    const publicDER = publicKey.export({ type: 'spki', format: 'der' })
+    const publicPEM = mcPubKeyToPem(publicDER)
+
+    if (!crypto.verify('RSA-SHA1', concat('UUID', client.uuid, 'i64', packet.expireTime, 'buffer', publicDER), mojangPubKey, packet.signature)) {
+      raise('multiplayer.disconnect.invalid_public_key_signature')
```
extremeheat, line 63: Only kick if server is enforcing secure profile in options

src/server/chat.js
extremeheat: yeah, and this will not work for 1.19.2. Not sure what the `...previousMessages` part is for but it may be a 1.19.3 thing I've not looked into

src/server/login.js
```diff
@@ -53,14 +51,27 @@ module.exports = function (client, server, options) {
 
       try {
         const publicKey = crypto.createPublicKey({ key: packet.signature.publicKey, format: 'der', type: 'spki' })
-        const publicPEM = mcPubKeyToPem(packet.signature.publicKey)
-        const signable = packet.signature.timestamp + publicPEM // (expires at + publicKey)
-
-        if (!crypto.verify('RSA-SHA1', Buffer.from(signable, 'utf8'), mojangPubKey, packet.signature.signature)) {
-          raise('multiplayer.disconnect.invalid_public_key_signature')
-          return
+        const publicDER = publicKey.export({ type: 'spki', format: 'der' })
```
extremeheat, line 54: Not sure what part of this is 1.19.3 and 1.19.2, but looks like this can be simplified (see how it's handled in the other PR)

src/server/login.js
```diff
@@ -197,11 +209,23 @@ module.exports = function (client, server, options) {
     pluginChannels(client, options)
 
     if (client.profileKeys) {
-      client.verifyMessage = (packet) => {
-        const signable = concat('i64', packet.salt, 'UUID', client.uuid, 'i64',
-          packet.timestamp, 'pstring', packet.message)
-
-        return crypto.verify('sha256WithRSAEncryption', signable, client.profileKeys.public, packet.crypto.signature)
+      client.verifyMessage = (packet, acknowledgements) => {
```
extremeheat, line 212: dupe of server/chat.js ?

extremeheat (commented):
Left some comments that came to mind. I haven't looked much into 1.19.3 so some of the things pointed out could be changes between the versions that need to be handled

src/client/chat.js
extremeheat: actually looks like it was renamed in 1.19.1 to secureChat. Same thing per https://github.dev/extremeheat/extracted_minecraft_data/blob/client1.19.1/client/net/minecraft/server/MinecraftServer.java#L668

src/client/chat.js
frej4189: This code only runs for 1.19.3. That's what the feature check is for

src/client/microsoftAuth.js
frej4189: I just copied the server implementation as I don't know how protodef treats UUIDs

src/client/chat.js
frej4189: No sorting needs to be done as it tracks index of oldest and newest element. Not pretty, but it works.

src/client/setProtocol.js
frej4189: It has playerUUID, which is the same thing. My minecraft-data is not in sync with yours

src/server/login.js
frej4189: Code in chat.js is for the chat session packet, which only exists in 1.19.3. This code is for 1.19/1.19.2

src/server/chat.js
frej4189: This is for the 1.19.3 chat session packet. The function in login.js is for the 1.19/1.19.2 login start packet. It is needed in both places.

src/client/chat.js
extremeheat: Is this the correct behavior on .3?

src/client/microsoftAuth.js
extremeheat: Client.uuid should already be defined elsewhere

src/client/setProtocol.js
extremeheat: Yeah, missed the above field

src/client/chat.js
frej4189: Yes

src/client/microsoftAuth.js
frej4189: It usually is set after login success packet. So it is not defined before login start

src/server/chat.js
extremeheat: actually, nevermind -- nodejs assumes PEM by default in verify https://github.com/nodejs/node/blob/4830a6cf2aa14144a1c36ae92dac7eed66fb8c13/lib/internal/crypto/keys.js#L520

src/client/chat.js
frej4189: What do you propose instead? An event that identifies that a message has been discarded?

src/client/chat.js
frej4189: I see that you're right. The packets name simply indicates the behaviour that I assumed.

As I am reading it, the packet should actually be interpreted like this:

- If chat preview is currently enabled and should_display_chat_preview is false, chat preview should be disabled
- If chat preview is currently disabled and should_display_chat_preview is true, chat preview should be set to the value that was initially read from the server data packet

In the two other cases (useChatPreview === should_display_chat_preview), nothing changes

src/client/chat.js
frej4189: While this is true, the vanilla client makes it virtually impossible to send 2 preview requests at once (essentially by treating responses the same way I do), and therefore I have not added that possibility either

src/client/microsoftAuth.js
frej4189: Noted. Will remove this from the PR

src/server/chat.js
frej4189: Right. Gonna add a check for that

src/server/login.js
frej4189: The other PR's validation does not work as the signature is generated from the public key without padding, but client provides it with padding.

src/server/chat.js
frej4189: > Pretty sure the public key should already be in DER format, 3 lines here should not be needed

Public key provided by client is padded. Signature is created with an unpadded key

src/client/chat.js
extremeheat: You're getting messages here from the server and the client implements the ticketing system on purpose to avoid asynchronous problems. Since the protocol is designed asynchronously it's not inconceivable that flooding messages to a server and then getting previews back would come out of order, depending on how the server is processing the message.

src/client/chat.js
extremeheat: Throw an exception

src/client/chat.js
extremeheat: This code doesn't run on 1.19.2 clients right ?

src/client/chat.js
frej4189: My point is the vanilla client does not support that either. The client only awaits one preview at once. If they come out of order the client will ignore them

extremeheat: Can you link to the source code for that?

src/client/chat.js
frej4189: Correct

src/client/chat.js
frej4189: https://github.com/extremeheat/extracted_minecraft_data/blob/client1.19.1/client/net/minecraft/client/gui/chat/ChatPreviewRequests.java

https://github.com/extremeheat/extracted_minecraft_data/blob/client1.19.1/client/net/minecraft/client/gui/chat/ClientChatPreview.java

ChatPreviewRequests only tracks one message at a time. ClientChatPreview only tracks one ChatPreviewRequests. It does attempt to send out more previews every 100ms but when the response comes back it only compares to the latest. If the preview response is not the latest it just returns null

moom0o:
Chunk size is 83 but only 5 was read ; partial packet : {"name":1575012541}; buffer :bd9183ef6525ff9d0a7131f446ef53241ff5bfdd68cb53690a007be4c868dace3fa49ac7a37d84de8103e3d7170bb79b82a70e41eece022858396db72694527999d1ae00386d0b6bb9f01b6e7df0b6f23eb661

Got this after installing your fork.

frej4189:
> Chunk size is 83 but only 5 was read ; partial packet : {"name":1575012541}; buffer :bd9183ef6525ff9d0a7131f446ef53241ff5bfdd68cb53690a007be4c868dace3fa49ac7a37d84de8103e3d7170bb79b82a70e41eece022858396db72694527999d1ae00386d0b6bb9f01b6e7df0b6f23eb661
> 
> Got this after installing your fork. server constantiam.net 1.19.2

Make sure to update minecraft-data aswell

vincss:
> > Chunk size is 83 but only 5 was read ; partial packet : {"name":1575012541}; buffer :bd9183ef6525ff9d0a7131f446ef53241ff5bfdd68cb53690a007be4c868dace3fa49ac7a37d84de8103e3d7170bb79b82a70e41eece022858396db72694527999d1ae00386d0b6bb9f01b6e7df0b6f23eb661
> > Got this after installing your fork. server constantiam.net 1.19.2
> 
> Make sure to update minecraft-data aswell

Hi !
I'm trying to do so with npm overrides : 
    "overrides": {
        "minecraft-protocol": {
            "minecraft-data": "https://github.com/frej4189/minecraft-data"
        }
    },
But it doesn't work. I tried several version of npm (v8, v9) but it doesn't work either. 
Do you have a working configuration ?

frej4189:
> > > Chunk size is 83 but only 5 was read ; partial packet : {"name":1575012541}; buffer :bd9183ef6525ff9d0a7131f446ef53241ff5bfdd68cb53690a007be4c868dace3fa49ac7a37d84de8103e3d7170bb79b82a70e41eece022858396db72694527999d1ae00386d0b6bb9f01b6e7df0b6f23eb661
> > > Got this after installing your fork. server constantiam.net 1.19.2
> > 
> > 
> > Make sure to update minecraft-data aswell
> 
> Hi ! I'm trying to do so with npm overrides :
> 
> ```
>     "overrides": {
>         "minecraft-protocol": {
>             "minecraft-data": "https://github.com/frej4189/minecraft-data"
>         }
>     },
> ```
> 
> But it doesn't work. I tried several version of npm (v8, v9) but it doesn't work either. Do you have a working configuration ?

This configuration works for me.

src/server/login.js
```diff
@@ -197,11 +209,23 @@ module.exports = function (client, server, options) {
...
+        if (mcData.supportFeature('chainedSignature')) { // 1.19.1/1.19.2
+          const hashable = crypto.createHash('sha256').update(concat('i64', packet.salt, 'i64', packet.timestamp / 1000n, 'pstring', packet.message, 'i8', 70))
```
rom1504, line 214: please implement this serialization only once

src/server/chat.js
extremeheat: Yeah, probably a good idea to do this in the client also.

src/server/login.js
```diff
@@ -53,14 +51,27 @@ module.exports = function (client, server, options) {
...
-        const signable = packet.signature.timestamp + publicPEM // (expires at + publicKey)
-
-        if (!crypto.verify('RSA-SHA1', Buffer.from(signable, 'utf8'), mojangPubKey, packet.signature.signature)) {
-          raise('multiplayer.disconnect.invalid_public_key_signature')
-          return
+        const publicDER = publicKey.export({ type: 'spki', format: 'der' })
+        const publicPEM = mcPubKeyToPem(publicDER)
+        if (mcData.supportFeature('profileKeySignatureV2')) {
+          if (!packet.playerUUID) {
+            client.end('Invalid packet format', JSON.stringify({ translate: 'disconnect.genericReason', with: [{ text: 'Invalid packet format' }] }))
```
extremeheat, line 58: You can use the raise function

frej4189: Raise does not allow for translation with arguments.

rom1504:
Hey, thanks for the work

Will award you and extreme the funding money

I merged https://github.com/PrismarineJS/node-minecraft-protocol/pull/1050

Let's try and get 1.19.3 going in another PR

rom1504:
Hey, thanks for the work

Will award you and extreme the funding money

I merged https://github.com/PrismarineJS/node-minecraft-protocol/pull/1050

Let's try and get 1.19.3 going in another PR

PR #1050, opened 2022-12-29
Initial 1.19.1/2 signed chat support

extremeheat:
prereqs:
* https://github.com/PrismarineJS/prismarine-auth/pull/65
* https://github.com/PrismarineJS/minecraft-data/pull/665

missing:
* server example
* tests

changes:
* adds new client.chat() wrapper to handle signing/changes to chat packet
* new `playerChat` event is emitted that handles verification internally
* server does internal verification also


src/server/chat.js
```diff
@@ -0,0 +1,103 @@
...
+    validateMessage(packet)
+  })
+
+  client.on('message_acknowledgement', (packet) => {
+    validateMessage(packet)
+  })
+
+  // Send a signed message from one player to another player on the server
+  client.logSentMessageFromPeer = (chatPacket) => {
+    if (!options.enforceSecureProfile) {
```
extremeheat, line 84: this function can be simplified

tecbeat:
When using the server you get the error `Chunk size is 847 but only 830 was read`. And an `Invalid signature for the profile's public key. Try restarting your game` in Minecraft itself.I am using this library in another project so unfortunately I have no idea how to solve it. But I imported this fix as a test. I am using this library in another project so unfortunately I have no idea how to solve it.

extremeheat:
> When using the server you get the error `Chunk size is 847 but only 830 was read`. And an `Invalid signature for the profile's public key. Try restarting your game` in Minecraft itself. I am using this library in another project so unfortunately I have no idea how to solve it. But I imported this fix as a test. I am using this library in another project so unfortunately I have no idea how to solve it.

Thanks, can confirm on the server, looking into it.

extremeheat:
>@tecbeat: When using the server you get the error `Chunk size is 847 but only 830 was read`. And an `Invalid signature for the profile's public key. Try restarting your game` in Minecraft itself. I am using this library in another project so unfortunately I have no idea how to solve it. But I imported this fix as a test. I am using this library in another project so unfortunately I have no idea how to solve it.

The `Invalid signature for the profile's public key. Try restarting your game` kick should be fixed if you update the minecraft-data dep with https://github.com/PrismarineJS/minecraft-data/pull/665 and pull the latest changes here.

The server implementation here needs some updates for chat, but with the new login packet in mc-data the server example shouldn't crash.

docs/chat.md
```diff
@@ -0,0 +1,54 @@
...
+message the contents, the sender UUID, timestamp, and signature are all sent so the Mojang server can verify the message 
+and send it for moderator review.
+
+Note: Since the server sends the public key, it's possible that the server can spoof the key and return a fake one, so
+only Mojang can truly know if a message came from a client (as it stores its own copy of the clients' chat key pair).
+
+## 1.19.1
+
+Starting with 1.19.1, instead of signing the message itself, a SHA256 hash of the message and last seen messages are
+signed instead. In addition, the hash is prepended with the signature of the previous message sent by the same client,
```
extremeheat, line 28: *the payload of the hash

src/server/login.js
```diff
@@ -186,6 +197,14 @@ module.exports = function (client, server, options) {
     })
     // TODO: find out what properties are on 'success' packet
     client.state = states.PLAY
+    client.settings = {}
+
+    if (client.protocolVersion >= 760) { // 1.19.1+
```
rom1504, line 202: Could be nicer to use support feature

src/server/chat.js
```diff
@@ -0,0 +1,166 @@
...
+          hash.update(concat('i8', 70, 'UUID', messageSender))
+          hash.update(messageSignature)
+        }
+        // Feed hash back into signing payload
+        verifier.update(hash.digest())
+      }
+      client._lastChatSignature = packet.signature
+      return verifier.verify(client.profileKeys.public, packet.signature)
+    } else {
+      const signable = concat('i64', packet.salt, 'UUID', client.uuid, 'i64', packet.timestamp, 'pstring', packet.message)
```
rom1504, line 105: Could be nice to put this serialization in a common function

rom1504:
Ok let's just add to the doc client.chat and playerChat event and merge

Manual testing will be done by people

Automated testing will be done later

I'm looking at the dependencies PRs

docs/API.md
```diff
@@ -45,6 +45,10 @@ Write a packet to all `clients` but encode it only once.
 
 Verifies if player's chat message packet was signed with their Mojang provided key
 
+### client.logSentMessageFromPeer(packet)
```
rom1504, line 48: Why isn't this done internally by the nmp client?

extremeheat: This one is for a client connected to a server, so the way to handle this internally would be to capture `client.write('player_chat')` as it gets sent to the server and do the handling there. That seemed too hacky so I added a separate call that server code needs to explicitly call into before broadcasting a player message to other players

extremeheat: An alternative could be to replace this with a "sendMessageFromPeer" function taking in a chat_message received from another player, but the payload between chat_message (serverbound) and player_chat (clientbound) is pretty different. Since chat formatting / filters and some other extra fields exist in the player_chat packet. So I thought it made sense to not overcomplicate nmp with lots of handling for this and just let the library users handle that and just call into this just for verification bookkeeping after sending off a message

rom1504: Ok makes sense, let's keep it that way

src/client/play.js
```diff
@@ -1,25 +1,43 @@
...
     client.uuid = packet.uuid
     client.username = packet.username
-    client.signMessage = (message, timestamp, salt = 0) => {
-      if (!client.profileKeys) throw Error("Can't sign message without profile keys, please set valid auth mode")
-      const signable = concat('i64', salt, 'UUID', client.uuid, 'i64',
-        timestamp / 1000n, 'pstring', JSON.stringify({ text: message }))
-      return crypto.sign('RSA-SHA256', signable, client.profileKeys.private)
+
+    if (mcData.supportFeature('signedChat')) {
+      if (!options.disableChatSigning && client.serverFeatures.enforcesSecureChat) {
```
frej4189, line 22: Shouldn't this be `options.disableChatSigning && client.serverFeatures.enforcesSecureChat` instead?

src/client/chat.js
```diff
@@ -0,0 +1,287 @@
...
+    const hash = crypto.createHash('sha256')
+    hash.update(concat('i64', packet.salt, 'i64', packet.timestamp / 1000n, 'pstring', packet.plainMessage, 'i8', 70))
+    if (packet.formattedMessage) hash.update(packet.formattedMessage)
+    for (const previousMessage of packet.previousMessages) {
+      hash.update(concat('i8', 70, 'UUID', previousMessage.messageSender))
+      hash.update(previousMessage.messageSignature)
+    }
+
+    // Chain integrity remains even if message is considered unverified due to expiry
+    const tsDelta = Date.now() - packet.timestamp
```
frej4189, line 105: packet.timestamp is a bigint so Date.now must be wrapped in the BigInt constructor

PR #1052, opened 2023-01-14
Fix breaking issues in 1.19.2

frej4189:
1. Type mismatch between Date.now (type number) and packet.timestamp (type bigint) causes program to throw when attempting to send a chat message
2. (Assumably) typo in play.js means program throws when server enforces secure chat and client has chat signing enabled (and does not throw when client has chat signing disabled)

### DIFF
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -102,7 +102,7 @@ module.exports = function (client, options) {
     }
 
     // Chain integrity remains even if message is considered unverified due to expiry
-    const tsDelta = Date.now() - packet.timestamp
+    const tsDelta = BigInt(Date.now()) - packet.timestamp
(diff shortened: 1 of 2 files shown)


rom1504:
thanks

src/client/chat.js
```diff
@@ -102,7 +102,7 @@ module.exports = function (client, options) {
     }
 
     // Chain integrity remains even if message is considered unverified due to expiry
-    const tsDelta = Date.now() - packet.timestamp
+    const tsDelta = BigInt(Date.now()) - packet.timestamp
```
extremeheat, line 105: This doesn't seem correct, will the `tsDelta > messageExpireTime` work ? The latter is a JS Number. packet.timestamp could instead be handled as a Number

frej4189: BigInt and Number can be compared as usual

PR #1054, opened 2023-01-15
Fix chat previews not working

frej4189:
- Chat preview packet handler now uses the correct name for fields from protocol definition (`packet.query` is now `packet.queryId`)
- Pass original message to `_signedChat` instead of previewed message
- Create message signature using previewed message

### DIFF
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -206,7 +206,7 @@ module.exports = function (client, options) {
           hash.update(previousMessage.signature)
         }
 
-        if(!!preview) hash.update(preview);
+        if (preview) hash.update(preview)
         // Feed hash back into signing payload
         signer.update(hash.digest())
       }
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -9,6 +9,7 @@ function isFormatted (message) {
     for (const key in comp) {
       if (key !== 'text') return true
     }
+    if(comp.text && comp.text !== message) return true
     return false
   } catch {
     return false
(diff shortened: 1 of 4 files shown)


src/client/chat.js
```diff
@@ -158,7 +158,7 @@ module.exports = function (client, options) {
         message,
         timestamp: options.timestamp,
         salt: options.salt,
-        signature: client.profileKeys ? client.signMessage(message, options.timestamp, options.salt) : Buffer.alloc(0),
+        signature: client.profileKeys ? client.signMessage(options.preview || message, options.timestamp, options.salt) : Buffer.alloc(0),
```
extremeheat, line 161: Is this correct? The signature check on the server will fail with this.

extremeheat (changes requested):

src/client/chat.js
frej4189: Then the server signature check is incorrect aswell. Vanilla server accepts this and will kick with "Invalid or missing signature" if we use message instead

src/server/chat.js
```diff
@@ -42,6 +42,8 @@ module.exports = function (client, server, options) {
   const raise = (translatableError) => client.end(translatableError, JSON.stringify({ translate: translatableError }))
   const pending = new Pending()
 
+  if (!options.generatePreview) options.generatePreview = message => message
```
extremeheat, line 45: needs to be added to docs

rom1504:
CI failing

PR #1055, opened 2023-01-15
Update convenience chat events so higher-level clients can properly parse chat

frej4189:
- Emit systemChat event in all versions (not just versions before 1.19)
- Emit plainMessage and unsignedContent as a part of the playerChat event
- Fix client throwing when trying to verify messages in version 1.19

### DIFF
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -113,7 +113,8 @@ module.exports = function (client, options) {
     const expired = !packet.timestamp || tsDelta > messageExpireTime || tsDelta < 0
     const verified = updateAndValidateChat(packet.senderUuid, packet.previousSignature, packet.signature, hash.digest()) && !expired
     client.emit('playerChat', {
-      message: packet.plainMessage || packet.unsignedChatContent,
+      plainMessage: packet.plainMessage,
+      unsignedMessage: packet.unsignedChatContent,
       formattedMessage: packet.formattedMessage,
       type: packet.type,
       sender: packet.senderUuid,
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -228,7 +228,7 @@ module.exports = function (client, options) {
   }
 
   client.verifyMessage = (pubKey, packet) => {
-    if (!mcData.supportFeature('chainedChatWithHashing')) { // 1.19.0
+    if (mcData.supportFeature('chainedChatWithHashing')) { // 1.19.1+
       // Verification handled internally in 1.19.1+ as previous messages must be stored to verify future messages
       throw new Error("Please listen to the 'playerChat' event instead to check message validity. client.verifyMessage is deprecated and only works on version 1.19.")
     }
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -69,6 +69,13 @@ module.exports = function (client, options) {
     }
   })
 
+  client.on('system_chat', (packet) => {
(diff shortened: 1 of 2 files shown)


src/client/chat.js
```diff
@@ -69,6 +69,13 @@ module.exports = function (client, options) {
     }
   })
 
+  client.on('system_chat', (packet) => {
+    client.emit('systemChat', {
```
rom1504, line 73: Can you document that event ?

src/client/chat.js
```diff
@@ -106,7 +113,8 @@ module.exports = function (client, options) {
     const expired = !packet.timestamp || tsDelta > messageExpireTime || tsDelta < 0
     const verified = updateAndValidateChat(packet.senderUuid, packet.previousSignature, packet.signature, hash.digest()) && !expired
     client.emit('playerChat', {
-      message: packet.plainMessage || packet.unsignedChatContent,
+      plainMessage: packet.plainMessage,
```
rom1504, line 116: Can you change the documentation to reflect that change ?

src/client/chat.js
frej4189: It is not a new event so should be documented already

src/client/chat.js
frej4189: Done in latest commit

src/client/chat.js
rom1504: https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/API.md seems it's not

Everything that is exposed / used by users should be documented so it's clear if we make breaking changes or not

frej4189: Interesting. Should have been included in the initial chat signing PR. 

Gonna add it to this one

src/client/chat.js
extremeheat: The point of the event is to abstract internals like this, why separate this to two fields?

frej4189: Because both are needed to properly format the chat message

extremeheat: Can you explain?

extremeheat: The chat example here is also broken with this change

frej4189: Where?

extremeheat: examples/chat

frej4189: > Can you explain?

Passing them the previous way means we will have to check whether `message` is a plaintext string or stringified JSON.

This not only adds unnecessary complications, but also fails to account for the rare case where a user might send a message that is also valid JSON, in which case it will be parsed incorrectly.

extremeheat: Which is which? Can you add this explanation to the doc and handling to the examples?

frej4189: > Which is which? Can you add this explanation to the doc and handling to the examples?

unsignedContent is stringified JSON. plainMessage is plain text

Docs should already be updated.

Going to update examples in a new PR

extremeheat: Don't see this explained in the doc, this is important for users listening to chat to know. To fix the examples you can also revert the `message` -> `plainMessage` change and keep the new unsigned field

frej4189: It is there. https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/API.md#playerchat-event

PR #1056, opened 2023-01-15
Realm Joining

LucienHH:


src/client/microsoftAuth.js
```diff
@@ -42,6 +46,38 @@ async function authenticate (client, options) {
   options.connect(client)
 }
 
+async function realmAuthenticate (client, options) {
+  validateOptions(options)
+
+  client.authflow = new PrismarineAuth(options.username, options.profilesFolder, options, options.onMsaCode)
```
extremeheat, line 52: this will be rewritten in L21

LucienHH: Ah of course, fixed now.

PR #1068, opened 2023-01-19
Add more fields to playerChat event

frej4189:
This adds senderName and targetName to playerChat event for version 1.19.2 so higher level libraries can properly render all chat types


Febzey (commented):
nice

Issue #1092, opened 2023-03-11
Chat signing possible problems on 1.19.3


MatteCarra:
Hello,
I was reviewing the new 1.19.3 code for chat signing.
I found some strange things that I wanted to point out.

What is the `index > player.sessionIndex`? player.sessionIndex is set to true at line 112 and never touched as far as I can tell.
https://github.com/PrismarineJS/node-minecraft-protocol/blob/9c767a0b33f5fe7eb5e109d10d6a9b61881c9c36/src/client/chat.js#L47
<br>
I don't see _signatureCache being modified (push called) anywhere. Is it normal?
https://github.com/PrismarineJS/node-minecraft-protocol/blob/9c767a0b33f5fe7eb5e109d10d6a9b61881c9c36/src/client/chat.js#L30

PR #1093, opened 2023-03-11
1.19.3 chat fixes

frej4189:
cc #1092 

A few things were not being properly updated. Only related to validation so the issue is not so noticeable (unless you actually use the `verified` field from the `playerChat` event).


rom1504:
Thanks

PR #1098, opened 2023-04-15
Added multiprotocol example for 1.19.0 - 1.19.3 + Better code in some examples

andriycraft:


andriycraft:
image

andriycraft:
You should switch from `standardjs` to `prettier`

andriycraft:
@rom1504

andriycraft:
@rom1504

rom1504:
Please say what it is
Looks like like of package.json removed. Why ?

andriycraft:
> Looks like like of package.json removed. Why ? 

Whats the point of containing package json that has no dependencies and basically useless

rom1504:
to keep the same pattern in all examples

also your file new names don't follow the snake case convention

faaxz:
> > Looks like like of package.json removed. Why ?
> 
> Whats the point of containing package json that has no dependencies and basically no usage

Well, actually i kinda agree with you here

IceTank:
Why have the old files been renamed? I don't see a reason why they should be renamed.

andriycraft:
> Why have the old files been renamed? I don't see a reason why they should be renamed.

To follow camelCase file naming

rom1504:
old

andriycraft:
old👍

PR #1224, opened 2023-05-20
chore(runtime): use LTS V18

AugustinMauroy:
V14 will become deprecated.


AugustinMauroy:
If you go on docs we use `node:`
`node:` mean it's an internal module.

and new internal module can only found using `node:` example `test`

rom1504:
ok, but why?

when will nodejs reverse this?

AugustinMauroy:
Now `test` module is only available with `node:test`

rom1504:
yeah I read https://stateful.com/blog/node-18-prefix-only-modules

I disagree, it's a bad change, it makes all future nodejs code not work with previous versions, for no strong reason.

Let's not use these prefixes

AugustinMauroy:
If you want the project to remain "healthy" you have to stay with LTS, so in any case the anicien code will be deprecated.

rom1504:
refer to https://github.com/PrismarineJS/prismarine-meta/issues/8 for why

PR #1226, opened 2023-05-27
1.19.4

extremeheat:
* Rename 'session' (packet) to chat_session_update to fix auth event conflict
* Implement 1.19.4 bundle packet to process several packets at once

after https://github.com/PrismarineJS/minecraft-data/pull/703


src/client/play.js
```diff
@@ -19,7 +19,7 @@ module.exports = function (client, options) {
         uuid: uuid.v4fast()
       }
 
-      client.write('session', {
+      client.write('chat_session_update', {
```
rom1504, line 22: why did we rename this packet ?

rom1504:
thanks for looking into it

src/client/play.js
extremeheat: The auth/microsoftAuth code emit the "session" event after authenticating, so it's not possible to disambiguate in event handler if it's the auth session or packet session. So I updated the naming in mcdata to match the vanilla packet name

Maks-gaming:
when bot getting damage
image

extremeheat:
> thanks for looking into it
> 
> please say when it's ready

Should be ready after the mc-data PR is merged. The bundle handling part can be removed if needed, I don't believe it's absolutely required

rom1504:
thanks

PR #1067, opened 2023-01-19
Handle signed commands in 1.19.0 and 1.19.2

extremeheat:
* Fixes commands that require message signing (/me, /kick, /ban, etc.). For commands with complex serialization this may break due  to handling spaces, but for now makes all the vanilla commands that need signing work.
* fixes field naming issue with chat header handling
* update types
* after https://github.com/PrismarineJS/minecraft-data/pull/671

PR #1241, opened 2023-07-19
Send chat commands as chat commands instead of chat messages for 1.19.3-1.20.1

lkwilson:
# motivation
Current behavior is that useChatSession capable protcols (1.19.3+) will send messages starting with '/' as a chat message (`client.write('chat_message', ...)`) instead of a chat command (`client.write('chat_command', ...)`) packet. This would cause chat commands to be sent as normal chat msgs and not be executed.
# change
This change makes it so that useChatSession protocols send chat commands as chat command packets.

~This change also prevents messages that have already been sent to the server from being sent again.~
# issues
~It fixes command messages for offline servers, but signing doesn't always seem to work for secure servers, so it continues to not work for those.~

### DIFF
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -388,7 +388,7 @@ module.exports = function (client, options) {
         timestamp: options.timestamp,
         salt: options.salt,
         argumentSignatures: signaturesForCommand(command, options.timestamp, options.salt),
-        messageCount: client._lastSeenMessages.capacity,
+        messageCount: client._lastSeenMessages.pending,
         acknowledged
       })
       return
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -338,7 +338,7 @@ module.exports = function (client, options) {
     options.timestamp = options.timestamp || BigInt(Date.now())
     options.salt = options.salt || 1n
 
-    if (message.startsWith('/')) {
+    if (message.startsWith('/') && !mcData.supportFeature('useChatSessions')) {
       const command = message.slice(1)
       client.write('chat_command', {
         command,
@@ -346,13 +346,43 @@ module.exports = function (client, options) {
         salt: options.salt,
         argumentSignatures: signaturesForCommand(command, options.timestamp, options.salt),
         signedPreview: options.didPreview,
-        messageCount: 0,
-        acknowledged: Buffer.alloc(3),
         previousMessages: client._lastSeenMessages.map((e) => ({
           messageSender: e.sender,
           messageSignature: e.signature
         })),
         lastRejectedMessage: client._lastRejectedMessage
+      })
+      return
+    }
+
+    if (message.startsWith('/') && mcData.supportFeature('useChatSessions')) {
+      const command = message.slice(1)
+
+      let acc = 0
+      const acknowledgements = []
+
+      for (let i = 0; i < client._lastSeenMessages.capacity; i++) {
+        const idx = (client._lastSeenMessages.offset + i) % 20
+        const message = client._lastSeenMessages[idx]
+        if (message) {
+          acc |= 1 << i
+          acknowledgements.push(message.signature)
+          message.pending = false
+        }
+      }
+
+      const bitset = Buffer.allocUnsafe(3)
+      bitset[0] = acc & 0xFF
+      bitset[1] = (acc >> 8) & 0xFF
+      bitset[2] = (acc >> 16) & 0xFF
+
+      client.write('chat_command', {
+        command,
+        timestamp: options.timestamp,
+        salt: options.salt,
+        argumentSignatures: signaturesForCommand(command, options.timestamp, options.salt),
+        messageCount: client._lastSeenMessages.capacity,
+        acknowledged: bitset
       })
       return
     }
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -375,6 +375,7 @@ module.exports = function (client, options) {
           messageCount: client._lastSeenMessages.pending,
           acknowledged
         })
+        client._lastSeenMessages.pending = 0
       } else {
         client.write('chat_command', {
           command,
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -338,7 +338,7 @@ module.exports = function (client, options) {
     options.timestamp = options.timestamp || BigInt(Date.now())
     options.salt = options.salt || 1n
 
-    if (message.startsWith('/') && !mcData.supportFeature('useChatSessions')) {
+    if (message.startsWith('/')) {
       const command = message.slice(1)
       client.write('chat_command', {
         command,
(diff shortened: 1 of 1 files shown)


src/client/chat.js
```diff
@@ -346,6 +346,8 @@ module.exports = function (client, options) {
         salt: options.salt,
         argumentSignatures: signaturesForCommand(command, options.timestamp, options.salt),
         signedPreview: options.didPreview,
+        messageCount: 0,
+        acknowledged: Buffer.alloc(3),
```
frej4189, line 350: This code block only runs in versions without chat sessions (1.19 through 1.19.2), so no need to add fields for later versions

lkwilson: Do chat commands still use chat sessions / chat_message packets?

I have this pull request sending chat commands to nmp for building the chat packet: https://github.com/PrismarineJS/mineflayer/pull/3110. But then that changes chat commands from being sent as chat_command to chat_message because of useChatSessions. If I remove the useChatSessions check for chat_commands, then these fields are required.

It seems that if I remove `&& !mcData.supportFeature('useChatSessions')` it makes the pull request work

frej4189: You're gonna want to create a separate block for versions that use chat sessions. Currently you are not actually providing the message acknowledgements, which will break things on servers enforcing secure chat

extremeheat: This is copying the implementation change inside mineflayer that should have been in nmp, so there wouldn't be a change from the status quo on implementation. But yes in a later PR that can be done after the command signing #1067 change is handled in mineflayer

frej4189: For acknowledgements we can definitely just copy what we do for regular chat signing

lkwilson: I created a new block mixing the two. So it sends a chat_command, but does the acknoledgement stuff. I didn't know what to put for messageCount however. I created a 1.19.3 server, and the bot gets kicked on master and on this version.

src/client/chat.js
```diff
@@ -355,6 +355,38 @@ module.exports = function (client, options) {
       return
     }
 
+    if (message.startsWith('/') && mcData.supportFeature('useChatSessions')) {
+      const command = message.slice(1)
+
+      let acc = 0
```
extremeheat, line 361: Since you're copying this from elsewhere, you should use a function instead of duplicating code

src/client/chat.js
```diff
@@ -355,32 +380,29 @@ module.exports = function (client, options) {
...
-          acknowledgements.push(message.signature)
-          message.pending = false
-        }
-      }
-
-      const bitset = Buffer.allocUnsafe(3)
-      bitset[0] = acc & 0xFF
-      bitset[1] = (acc >> 8) & 0xFF
-      bitset[2] = (acc >> 16) & 0xFF
+    if (message.startsWith('/') && mcData.supportFeature('useChatSessions')) {
```
extremeheat, line 383: Did you review the code ? Why are there 2 different handlings here for chat_command? This is likely what's failing the mineflayer PR. 
image

lkwilson: I think it was messageCount being set to capacity instead of pending that was causing the other pr to fail

lkwilson: The two different handlings are because the first is for <1.19.3 and the second (useChatSessions) is for 1.19.3+

lkwilson: Mineflayer is passing with the fix

extremeheat: Yeah, ok. I think the readability here is very poor with the current code, as it's not obvious at a glance the two conditions are different. Instead I think this should be restructured like

```js
if (message.startsWith('/')) {
  if (supportFeature(a)) { return ... }
  else { return ... }
}
```

for better clarity

lkwilson: Agreed, ok, I'll do that. I wasn't sure if it was preferred to have two separate blocks or nested ones

rom1504:
does it work?

lkwilson:
These changes are ready

lkwilson:
Hey, I said this was ready, but I think I fixed chat signing for 1.19.3+, so I pushed those changes to this pr. It was sending signatures of messages / commands that were already sent. This should close out https://github.com/PrismarineJS/mineflayer/issues/2992 and https://github.com/PrismarineJS/mineflayer/issues/2874. If you all don't mind taking another look at this, I would really appreciate it. Thank you

src/client/chat.js
```diff
@@ -222,7 +222,7 @@ module.exports = function (client, options) {
       })
 
       if (packet.signature) {
-        client._lastSeenMessages.push(packet.signature)
+        client._lastSeenMessages.push({ signature: packet.signature, pending: false })
```
frej4189, line 225: This won't work as signature needs to be a Buffer. The push method on lastSeenMessages also handles pending state internally already.

Furthermore, signatures must be pending: true by default or chat signing will break after a hide_chat packet

lkwilson: The LastSeenMessages object was sending signatures for messages with pending set to false. Is that what it was supposed to be doing? If you set pending to true here, it sends the signature containing only msgs it hasn't sent before. However, in both of these cases, I get kicked from the vanilla server when I send commands. If I set pending to false, then i never get kicked from offline or online vanilla servers, but that's because it doesn't send signatures, so I'm not sure what else that breaks. Do you know of things I should test to make sure that this doesn't break anything that is currently working?

As for hidden messages, those are handled by setting the whole entry to null, and those don't use pending as far as I can tell. That logic should still be the same.

lkwilson: Wait, I found what I did wrong. Chat messages set pending to 0 afterwards, I forgot that for commands. Adding that makes everything work. I just pushed that version

src/client/chat.js
```diff
@@ -334,53 +334,75 @@ module.exports = function (client, options) {
...
-        })),
-        lastRejectedMessage: client._lastRejectedMessage
-      })
+      if (mcData.supportFeature('useChatSessions')) {
+        const { acknowledged } = getAcknowledgements()
+        client.write('chat_command', {
+          command,
+          timestamp: options.timestamp,
+          salt: options.salt,
+          argumentSignatures: signaturesForCommand(command, options.timestamp, options.salt),
```
lkwilson, line 374: `signaturesForCommand` calls `client.signMessage(signable, ts, salt)`, and `signMessage` accepts `acknowledgements`. Should I pass `acknowledgements` to `signaturesForCommand` to pass to `signMessage`?

frej4189: Where you pass them from is up to you

lkwilson: Ok, I fed them through

rom1504:
I think it would be nice if we had some tests here as it's quite complicated

but if you're confident it's good, merging, we can revert if it breaks things

PR #1246, opened 2023-07-24
Update index.d.ts to have supported versions

MalkistCoder:
I replaced `version: string` in the `index.d.ts` file to have all the supported versions in a union type instead so that TypeScript only allows the version to be one of the supported versions.


rom1504:
please refactor this as an enum, otherwise it's hard to maintain

rom1504:
Please reopen if you want to finish it

PR #1254, opened 2023-08-17
Allow to create custom client & communication between clients

zardoy:
For https://github.com/PrismarineJS/flying-squid/issues/491
Also allow creating custom clients so don't need to patch it with webpack


src/client.js
```diff
@@ -13,9 +13,11 @@ const createDecipher = require('./transforms/encryption').createDecipher
 const closeTimeout = 30 * 1000
 
 class Client extends EventEmitter {
-  constructor (isServer, version, customPackets, hideErrors = false) {
+  constructor (isServer, version, customPackets, hideErrors = false, customCommunication = undefined) {
```
zardoy, line 16: Hm, pretty sure it would be better if it was an object like
```suggestion
  constructor (isServer, version, additionalOptions) {
```

src/client.js
```diff
@@ -230,10 +249,15 @@ class Client extends EventEmitter {
...
   write (name, params) {
-    if (!this.serializer.writable) { return }
-    debug('writing packet ' + this.state + '.' + name)
+    if (!this.customCommunication && !this.serializer.writable) { return }
+    debug(`[${this.state}] from ${this.isServer ? 'server' : 'client'}: ` + name)
     debug(params)
-    this.serializer.write({ name, params })
+
+    if (this.customCommunication) {
+      this.customCommunication.sendData({ name, params, state: this.state })
```
zardoy, line 257: It doesn't support writeBundle and writeRaw as I'm not sure how flying-squid could use it

rom1504:
please open a pr on flying squid showing how you use it

zardoy:
> please open a pr on flying squid showing how you use it

Oke! I've included some other important changes to the same branch, so I have to rebase it first I guess. Will look into it tomorrow!

src/client.js
```diff
@@ -39,100 +41,115 @@ class Client extends EventEmitter {
...
-      }
-      if (parsed.metadata.name === 'bundle_delimiter') {
-        if (this._mcBundle.length) { // End bundle
-          this._mcBundle.forEach(emitPacket)
-          emitPacket(parsed)
-          this._mcBundle = []
-        } else { // Start bundle
+
+    if (this.customCommunication) {
+      this.customCommunication.receiverSetup.call(this, (/** @type {{name, params, state?}} */parsed) => {
```
rom1504, line 94: The fact customCommunication should have a receiverSetup property in it with a call function should be documented

src/client.js
```diff
@@ -230,10 +249,15 @@ class Client extends EventEmitter {
...
   write (name, params) {
-    if (!this.serializer.writable) { return }
-    debug('writing packet ' + this.state + '.' + name)
+    if (!this.customCommunication && !this.serializer.writable) { return }
+    debug(`[${this.state}] from ${this.isServer ? 'server' : 'client'}: ` + name)
     debug(params)
-    this.serializer.write({ name, params })
+
+    if (this.customCommunication) {
+      this.customCommunication.sendData.call(this, { name, params, state: this.state })
```
rom1504, line 257: sendData should be documented

src/client.js
rom1504: I don't think this whole change makes a lot of sense.

What about instead letting the user provide its own Client class ?

This is overriding pretty much everything so I don't think it really fits in Client

src/createClient.js
```diff
@@ -32,7 +32,8 @@ function createClient (options) {
   options.protocolVersion = version.version
   const hideErrors = options.hideErrors || false
 
-  const client = new Client(false, version.minecraftVersion, options.customPackets, hideErrors)
+  const Client = options.customClient ?? DefaultClient
```
rom1504, line 35: Why isn't that option enough? Seems like it is to me

zardoy: Isn't it to allow users to provide custom implementation of the Client class? Like you said in https://github.com/PrismarineJS/node-minecraft-protocol/pull/1254#discussion_r1327909079

zardoy: I think I got it. I refactored all it. See latest commits

zardoy:
Also FYI `setProtocol` call should be moved before `tcpDns` since the latter is emitting `connect` event before the an event listener for it is added in `setProtocol` (when you use `socket` option)

extremeheat:
Why can't you implement this functionality with a custom `options.customClient` in createClient defined outside nmp? If the client and server are both embedded there is no reason to go through the authentication, compression or serialization steps here, as you can simply share one EventEmitter between the two.

zardoy:
You are totally right! I'm sorry I missed the intention to have custom client implementation outside of nmp (node minecraft protocol) in https://github.com/PrismarineJS/node-minecraft-protocol/pull/1254#discussion_r1327909337. Since then we can have just custom server and client options here. And yes, I don't skip auth currently, most probably I could just emit login event directly..

zardoy:
done. I simply moved this custom client to pwc, I'm still not sure about skipping auth even in the context of the local server, I feel it can break mineflayer...

docs/API.md
```diff
@@ -34,7 +34,8 @@ automatically logged in and validated against mojang's auth.
  * enforceSecureProfile (optional) : Kick clients that do not have chat signing keys from Mojang (1.19+)
  * generatePreview (optional) : Function to generate chat previews. Takes the raw message string and should return the message preview as a string.  (1.19-1.19.2)
  * socketType (optional) : either `tcp` or `ipc`. Switches from a tcp connection to a ipc socket connection (or named pipes on windows). With the `ipc` option `host` becomes the path off the ipc connection on the local filesystem. Example: `\\.\pipe\minecraft-ipc` (Windows) `/tmp/minecraft-ipc.sock` (unix based systems). See the ipcConnection example for an example.
- 
+ * Server : You can pass a custom server class to use instead of the default one.
```
rom1504, line 37: it would be really good to say here what methods this class must expose

docs/API.md
```diff
@@ -129,21 +130,22 @@ Returns a `Client` instance and perform login.
...
-   * realmId : The id of the Realm to join. 
-   * pickRealm(realms) : A function which will have an array of the user Realms (joined/owned) passed to it. The function should return a Realm. 
+ * realms : An object which should contain one of the following properties: `realmId` or `pickRealm`. When defined will attempt to join a Realm without needing to specify host/port. **The authenticated account must either own the Realm or have been invited to it**
+   * realmId : The id of the Realm to join.
+   * pickRealm(realms) : A function which will have an array of the user Realms (joined/owned) passed to it. The function should return a Realm.
+ * Client : You can pass a custom client class to use instead of the default one, which would allow you to create completely custom communication. Also note that you can use the `stream` option instead where you can supply custom duplex, but this will still use serialization/deserialization of packets.
```
rom1504, line 148: it would be good to say here what methods this class must expose

rom1504:
left a comment regarding the doc but otherwise lgtm

docs/API.md
zardoy: I guess there is the typing for that (with all signatures etc)

rom1504:
ok

PR #1257, opened 2023-10-05
chat: Only sign command args when profile keys defined

evan-goode:
I run a custom authentication server and I was getting the error:

```
Error: Can't sign message without profile keys, please set valid auth mode
```

when my bot issues the `/me` command. This patch changes the `argumentSignatures` field of the `chat_command` packet to only be calculated if `client.profileKeys` is valid, like the current behavior of the `chat_message` packet.

### DIFF
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -371,7 +371,7 @@ module.exports = function (client, options) {
           command,
           timestamp: options.timestamp,
           salt: options.salt,
-          argumentSignatures: signaturesForCommand(command, options.timestamp, options.salt, options.preview, acknowledgements),
+          argumentSignatures: client.profileKeys ? signaturesForCommand(command, options.timestamp, options.salt, options.preview, acknowledgements) : [],
           messageCount: client._lastSeenMessages.pending,
           acknowledged
         })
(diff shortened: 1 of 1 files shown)


evan-goode:
Hmm, that CI failure looks unrelated, not sure though:

```
1) client 1.20.1
       "before all" hook in "client 1.20.1":
     FetchError: request to https://launchermeta.mojang.com/mc/game/version_manifest.json failed, reason: getaddrinfo EAI_AGAIN launchermeta.mojang.com
      at ClientRequest.<anonymous> (node_modules/node-fetch/lib/index.js:1501:11)
      at ClientRequest.emit (node:events:517:28)
      at TLSSocket.socketErrorListener (node:_http_client:501:9)
```

extremeheat:
You can rerun the CI with `git push upstream master --force`

extremeheat:
Have you tested this to work in-game?

evan-goode:
Looks like the CI is happy now.

Sorry for the delay, yes this works for me in game, on 1.20.1, with `enforce-secure-profile=false` in `server.properties` and with an authentication server that does not sign player public keys.

I also tested regular Microsoft authentication with this patch and it didn't seem to break anything.

PR #1261, opened 2023-11-06
Update to 1.20.2

SuchBlue:


Parker2991:
i dont think its just that i think its also something with compiler-minecraft.js and minecraft.js in node-minecraft-protocol

rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol/pull/1264

rom1504:
moving to #1265

PR #1263, opened 2023-11-22
UnsignedContent correction for playerChat event.

Ynfuien:
When emitting one of `playerChat` events, there was used nonexisting `packet.unsignedContent`, to set `unsignedContent` field. So when using `playerChat` event, `unsignedContent` was `unknown`.

### DIFF
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -198,7 +198,7 @@ module.exports = function (client, options) {
       if (verified) client._signatureCache.push(packet.signature)
       client.emit('playerChat', {
         plainMessage: packet.plainMessage,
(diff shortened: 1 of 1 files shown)


frej4189:
The field _is_ called `unsignedContent` - at the very list it is in 1.19.2

https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.19.2/protocol.json#L3319

If other versions have a different field name you should fix that in minecraft-data instead

rom1504:
@frej4189 I read unsignedChatContent where you are pointing. What am I missing?

rom1504:
ah this PR https://github.com/PrismarineJS/minecraft-data/pull/795/files

PR #1264, opened 2023-12-12
Initial 1.20.2 support

crux153:
Credits to @xZarex and @wgaylord.

This PR adds preliminary support for version 1.20.2 or later, including support for the configuration stage.

Before merging this PR, we need https://github.com/PrismarineJS/prismarine-nbt/pull/78 to be merged and released.

I have tested that a bot can successfully log in to a server and interact with chunks using this patch. However, please note that I haven't tested [all changes](https://www.minecraft.net/en-us/article/minecraft-java-edition-1-20-2) introduced in version 1.20.2.


extremeheat:
There is no need to make changes to the NBT like this. There are no changes to the NBT format minus a small special case inside the root compound. Instead, this can be fixed by adding a new parameter to the existing UUID type inside minecraft-data specifying the expected root type, and then implementing that inside nmp here. Inside prismarine-nbt, for the new root type, simply omit the name field in nbt.json.

rom1504:
@crux153 can you please

1. take into account the above comment from @extremeheat to implement this better ?
2. change the version of prismarine-nbt here towards your PR to make the tests succeed

extremeheat:
Also to elaborate on how to do the change to minecraft-data, see how "string" is a "pstring" with a lengthType prop on it. Just like that you can do 

```
"nbt": { tagType: "anonymousCompound" }
```

anonymousCompound would be alternative to "compound" in nbt.json, but without a name.

crux153:
@extremeheat I'm looking at minecraft-data and prismarine-nbt and trying to follow your instruction. But I'm quite confused since it seems like [the compound type is actually defined using the nbt type in prismarine-nbt](https://github.com/PrismarineJS/prismarine-nbt/blob/1c62910515ab78b0599ea0700371c34811f5b0a9/compiler-compound.js#L20). 

Also, according to my understanding, adding a parameter like [pstring's countType](https://github.com/ProtoDef-io/node-protodef/blob/912683467d4eec4df852a4ae66030338416b48b1/src/datatypes/compiler-utils.js#L3) would require defining parametrizable type.

I guess your intention is to avoid feature detection in prismarine-nbt and node-minecraft-protocol's protocol handling part, making it configurable in minecraft-data's protocol.json.

So in order to achieve something like
```json
"nbt": [
  "nbt",
  {
    "tagType": "anonymousCompound"
  }
]
```
in [minecraft-data/data/pc/1.20.2/protocol.json](https://github.com/PrismarineJS/minecraft-data/blob/b1088724d565819f1f1c8ae4931ca3b2fe609b3a/data/pc/1.20.2/protocol.json),

it would require defining an additional anonymousCompound under prismarine-nbt's nbt.json without name
```json
"nbt": [
  "container",
  [
    {
      "name": "type",
      "type": "nbtMapper"
    },
    {
      "name": "name",
      "type": "nbtTagName"
    },
    {
      "name": "value",
      "type": ["nbtSwitch",{"type":"type"}]
    }
  ]
],
"anonymousCompound": [
  "container",
  [
    {
      "name": "type",
      "type": "nbtMapper"
    },
    {
      "name": "value",
      "type": ["nbtSwitch",{"type":"type"}]
    }
  ]
]
```
like this,

and then define a parametrizable `nbt` type in [node-minecraft-protocol/src/datatypes/compiler-minecraft.js](https://github.com/PrismarineJS/node-minecraft-protocol/blob/788bff289030fa66c980de82d82cb953bf76332b/src/datatypes/compiler-minecraft.js#L19) that resolves to `nbt` or `anonymousCompound` based on `tagType`.

Please correct me if I'm approaching this entirely wrong.

rom1504:
thanks for the initial work here, we included information you found here and improved on it at https://github.com/PrismarineJS/node-minecraft-protocol/pull/1265

Issue #1258, opened 2023-10-13
support 1.20.2


lastuniverse:
please add support for minecraft 1.20.2

SilkePilon:
Support will be added once you implement it and open a Pull Request.

lastuniverse:
It's a pity. If I succeed, I’ll definitely make a request

kittenvr:
upvote

Arthessia:
upvote here too :)

lastuniverse:
Unfortunately, I won’t be able to deal with this issue for the next two weeks. It's a total mess at work.

lastuniverse:
There is still a problem. Before updating node_minecraf_protocol to 1.20.2 you need to wait for the update https://github.com/PrismarineJS/minecraft-data

lastuniverse:
I looked at https://github.com/PrismarineJS/minecraft-data .
Basically I'm not able to update minecraft-data without diving deep into it (((

lastuniverse:
We are waiting for your reaction here https://github.com/PrismarineJS/minecraft-data/issues/779

SuchBlue:
https://github.com/PrismarineJS/minecraft-data now supports 1.20.2

SuchBlue:
I have created a pull request https://github.com/PrismarineJS/node-minecraft-protocol/pull/1261

7yl4r:
an example error when trying to connect:

```
(base) tylar@laptop:~/repos/mineflayer$ node examples/looker.js localhost 25565
TypeError: Serialization error for login.toServer : Write error for undefined : Cannot read properties of undefined (reading 'length')
    at module.exports.Write.UUID (/home/tylar/repos/mineflayer/node_modules/minecraft-protocol/src/datatypes/compiler-minecraft.js:50:25)
    at Object.packet_login_start (eval at compile (/home/tylar/repos/mineflayer/node_modules/protodef/src/compiler.js:258:12), <anonymous>:708:26)
    at eval (eval at compile (/home/tylar/repos/mineflayer/node_modules/protodef/src/compiler.js:258:12), <anonymous>:754:62)
    at packet (eval at compile (/home/tylar/repos/mineflayer/node_modules/protodef/src/compiler.js:258:12), <anonymous>:760:9)
    at CompiledProtodef.write (/home/tylar/repos/mineflayer/node_modules/protodef/src/compiler.js:76:12)
[...]

PR #1265, opened 2023-12-18
Pc1.20.2

extremeheat:
Fix https://github.com/PrismarineJS/node-minecraft-protocol/issues/1258

~~The configuration state introduces lots of small problems in server/client expectations so there remain some errors in the test due to state desync. Will need to be investigated further~~

Needs audit of protocol.json for 1.20.2 against vanilla source code diff. Also, node-protodef-validator is not working.

After https://github.com/PrismarineJS/prismarine-nbt/pull/81 and https://github.com/PrismarineJS/minecraft-data/pull/810


src/createServer.js
```diff
@@ -45,6 +45,7 @@ function createServer (options = {}) {
   server.onlineModeExceptions = Object.create(null)
   server.favicon = favicon
   server.options = options
+  options.registryCodec = options.registryCodec || mcData.registryCodec || mcData.loginPacket?.dimensionCodec
```
extremeheat, line 48: The dimensionCodec is no longer in loginPacket. So it it needs to be dumped and put into mcdata seperately. It's required before clients are able to join (to exit configuration state)

src/datatypes/minecraft.js
```diff
@@ -43,35 +43,36 @@ function writeUUID (value, buffer, offset) {
   return offset + 16
 }
 
-function readNbt (buffer, offset) {
-  return nbt.proto.read(buffer, offset, 'nbt')
+function readNbt (buffer, offset, { tagType } = { tagType: 'nbt' }) {
```
rom1504, line 46: Are we confident this is only useful for the network serialization?

Would there be any drawback to put this in prismarine-nbt ?

extremeheat: This is already just directly calling prismarine-nbt so it should be possible to just import the prismarine-nbt types directly into the protocol. But the optionalNBT type may actually require a custom type as there is a special case with zero.

extremeheat: Would require importing all the custom types also https://github.com/PrismarineJS/prismarine-nbt/blob/master/nbt.js#L11C1-L20C48

rom1504: It seems to me that would be much cleaner as all nbt stuff would then live in prismarine-nbt

Also important if other stuff than nmp use the same format, for example anvil

src/server/login.js
```diff
@@ -217,6 +208,14 @@ module.exports = function (client, server, options) {
     if (client.supportFeature('signedChat')) chatPlugin(client, server, options)
     server.emit('login', client)
   }
+
+  function onClientLoginAck () {
+    client.state = states.CONFIGURATION
+    client.write('registry_data', { codec: options.registryCodec || {} })
+    client.write('finish_configuration', {})
+    client.state = states.PLAY
+    server.emit('playerJoin', client)
```
extremeheat, line 217: Ah, problem seems due to not waiting for client to send back a finish_config

extremeheat:
ok, tests are now passing, but looks like there may be an issue with `packet_chunk_batch_start`
```
[05:40:13] [Server thread/INFO]: [Not Secure] <Player> hello everyone; I have logged in.
PartialReadError: Read error for undefined : undefined
    at new ExtendableError (/workspace/node-minecraft-protocol/node_modules/protodef/src/utils.js:63:13)
    at new PartialReadError (/workspace/node-minecraft-protocol/node_modules/protodef/src/utils.js:70:5)
    at Object.readVarInt [as varint] (/workspace/node-minecraft-protocol/node_modules/protodef/src/datatypes/utils.js:69:45)
    at Object.packet_chunk_batch_start (/workspace/node-minecraft-protocol/node_modules/protodef/src/g-ReadCompiler104533269d.js:956:67)
    at /workspace/node-minecraft-protocol/node_modules/protodef/src/g-ReadCompiler104533269d.js:3002:74
    at packet (/workspace/node-minecraft-protocol/node_modules/protodef/src/g-ReadCompiler104533269d.js:3104:9)
    at CompiledProtodef.read (/workspace/node-minecraft-protocol/node_modules/protodef/src/compiler.js:70:12)
    at e.message (/workspace/node-minecraft-protocol/node_modules/protodef/src/compiler.js:111:49)
    at tryCatch (/workspace/node-minecraft-protocol/node_modules/protodef/src/utils.js:50:16)
    at CompiledProtodef.parsePacketBuffer (/workspace/node-minecraft-protocol/node_modules/protodef/src/compiler.js:111:29)
    at FullPacketParser.parsePacketBuffer (/workspace/node-minecraft-protocol/node_modules/protodef/src/serializer.js:68:23)
    at FullPacketParser._transform (/workspace/node-minecraft-protocol/node_modules/protodef/src/serializer.js:74:21)
  ...
    at Socket.emit (node:events:514:28)
    ...
      ✔ does not crash for 10000ms (10489ms)
```

extremeheat:
Actually, `packet_chunk_batch_start` is a new packet in 1.20.2, it's incorrect in mc-data. Would be a good idea to check the other types also against the source code, there are probably more unreviewed errors/missing things

wgaylord:
> Actually, `packet_chunk_batch_start` is a new packet in 1.20.2, it's incorrect in mc-data. Would be a good idea to check the other types also against the source code, there are probably more unreviewed errors/missing things

Looks like that may have been a dumb copy paste error on my part when doing all the packet changes in 1.20.2.

extremeheat:
Ok, tests are passing now. Validator issue is fixed, just need to check over the protocol.json now. But anyway I think this can be merged before that, would just need to be checking before working on mineflayer

src/datatypes/compiler-minecraft.js
```diff
@@ -16,8 +16,6 @@ module.exports = {
         size: buffer.length - offset
       }
     }],
-    nbt: ['native', minecraft.nbt[0]],
-    optionalNbt: ['native', minecraft.optionalNbt[0]],
     compressedNbt: ['native', minecraft.compressedNbt[0]],
```
rom1504, line 19: maybe we could also move compressedNbt but not a big deal

src/server/login.js
```diff
@@ -166,37 +166,28 @@ module.exports = function (client, server, options) {
...
       username: client.username,
       properties: []
     })
-    // TODO: find out what properties are on 'success' packet
-    client.state = states.PLAY
+    if (client.supportFeature('hasConfigurationState')) {
+      client.once('login_acknowledged', onClientLoginAck)
+    } else {
+      client.state = states.PLAY
+      server.emit('playerJoin', client)
```
rom1504, line 189: this new event should be documented.

It would be breaking for flying-squid (and other users), why do we change from 'login' ?

rom1504:
looks good, I'm glad we could solve the nbt change with no hack and actually even improving the code, nice work

I think it would be good to fix `PartialReadError: Read error for undefined : undefined` before merging as this is definitely going to break usage of the lib for this version

@wgaylord do you want to check the packets you added against the source? that way you can learn to do it for 1.20.2 and then apply the same method for 1.20.3

src/server/login.js
extremeheat: The server sets the mode to from LOGIN to CONFIG after receiving a `login_ack` from the client. The server emits `login` right after the ack, but the state is still in CONFIG so sending PLAY packets (like previous `login` event would let you do) would fail, so in 1.20.2+ now need to either listen for `state` event for it to turn to PLAY or use the added `playerJoin` event before sending game packets

wgaylord:
Yeah, I am going to go thru all the packets I touched and double check them some time tomorrow.

src/server/login.js
rom1504: Ok makes sense

rom1504:
No value for type soundSource , packedChunkedPos in packet tests

rom1504:
does seem to work fully except for this one, no error anymore for 1.20.2

rom1504:
I think it's good now, merging

PR #1267, opened 2023-12-25
Improve CI setup for per version tests

rom1504:
* move lint out of per version
* bring packet cycle test back to per version
* generate versions to test from js file to avoid having versions in 2 places


rom1504:
not sure about this distinction between client / not client ; par / not par tests
I'm thinking to just do per version again

do you remember the point of this @u9g ? it's not even faster

rom1504:
still not ready

rom1504:
one additional thing I'd like to do here is make sure all the versions in version.js are mentioned in the readme

besides that I think it's ready

rom1504:
do you want to take a look @extremeheat ?

test/serverTest.js
```diff
@@ -94,7 +94,7 @@ for (const supportedVersion of mc.supportedVersions) {
     }
   }
 
-  describe('mc-server ' + version.minecraftVersion, function () {
+  describe('mc-server ' + supportedVersion+'v', function () {
```
extremeheat, line 97: Lint isn't running on test/serverTest since the .gitignore has

`test/server*` instead of `test/server_*` in it https://github.com/PrismarineJS/node-minecraft-protocol/blob/improve_ci_setup/.gitignore#L3

extremeheat (commented):
lgtm

test/serverTest.js
rom1504: good catch thanks

rom1504:
ok merging as is

will follow up in #1269 and #1268 for additional improvements

Issue #1262, opened 2023-11-21
Chat Signing is Forced for Commands


WhoTho:
When sending commands through `client.chat`, chat signing of command arguments is forced despite the client not having profile keys. This causes the "Can't sign message without profile keys, please set valid auth mode" error.

I don't know how chat signing works, but I assume it'll be as easy to fix as changing [the signing line in signaturesForCommand](https://github.com/PrismarineJS/node-minecraft-protocol/blob/827b6cdb61eac9e1b737fa0a9fb8e81385a78d3d/src/client/chat.js#L327) to something like:
```js
signatures.push({
  argumentName: fieldName,
  signature:
    client.profileKeys ?
    client.signMessage(signable, ts, salt, preview, acknowledgements) :
    Buffer.alloc(0)
})
```

frej4189:
Can be fixed by mirroring the normal chat message signing behaviour (e.g. providing no signature when profile keys are not set)

https://github.com/PrismarineJS/node-minecraft-protocol/blob/827b6cdb61eac9e1b737fa0a9fb8e81385a78d3d/src/client/chat.js#L403

extremeheat:
@frej4189 Can you open a PR to fix or show suggested patch?

extremeheat:
Should be fixed with #1277

PR #1277, opened 2024-01-01
Allow commands not to be signed

forester302:
Updated chat.js as suggested in https://github.com/PrismarineJS/node-minecraft-protocol/issues/1262

### DIFF
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -371,7 +371,7 @@ module.exports = function (client, options) {
           command,
           timestamp: options.timestamp,
           salt: options.salt,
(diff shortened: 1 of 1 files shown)


extremeheat (approved):
I don't like that the chat function has so much duplicative code there, it's rather hard to follow. It could do with some simplification. Also `_session` is a bad variable name, it's not clear to me what it has to do with chat.

But anyway LGTM for a temporary fix, cc @frej4189

frej4189:
LGTM.

I don't think this is a temporary fix. This is exactly how it's supposed to work

PR #1094, opened 2023-03-13
Adds support for partial packets

Pandapip1:
Fixes #811

Issue #811, opened 2021-01-22
Chunk size is *** but only 1 was read && client disconnect with DecoderException


ifxory-plugins:
[!] The [FAQ](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - minecraft-protocol: 6.14.11
 - server: minecraft-protcol
 - node: 10.19.0

## Detailed description of a problem
I'm building proxy using minecraft-protocol only. But having 2 troubles. Player everytime kicked with error(without any errors in my proxy process).

![image](https://user-images.githubusercontent.com/63095272/105490045-3c01dd80-5cc5-11eb-80f5-d5beb9d41bff.png)

And 2nd problem is

![image](https://user-images.githubusercontent.com/63095272/105490101-58057f00-5cc5-11eb-81a9-4bb68ff13481.png)

## Current code
```js
var server = utils.createServer(downstream, 1, "My shiny proxy")
let client = null

server.on("login", (socket) => {

    var loginPacket = utils.createLoginPacket(server, socket, false, 0, 10);
    var positionPacket = utils.createPositionPacket(0, 100, 0, 0, 0);

    socket.write('login', loginPacket);
    socket.write('position', positionPacket);

    logger.info('Player ' + socket.username + ' logged in.');

    socket.on('packet', (packet, meta) => {

        logger.downstream('Client => Downstream => Upstream => Server');

        utils.emitEvents(socket, meta, packet);

    });

    socket.on('raw', (data, meta) => {

        if(client === null) {

            client = utils.createClient(upstream, socket.username);
            client.on('raw', (data, meta) => {

                socket.write(meta.name, data);

            });
            client.on('packet', (data, meta) => {

                logger.upstream('Client <= Downstream <= Upstream <= Server');

                utils.emitEvents(client, meta, data);

            });

        }
        
        console.log(data)

        client.write(meta.name, data);

    });

    socket.on("message", (packet, message) => {

        if(message.startsWith('/')) {

            logger.data('Player ' + socket.username + ' issued server command: ' + message);

        }

        else {

            logger.data('Player ' + socket.username + ' > ' + message);

        }

    });

    socket.on('end', (reason) => {

        logger.info('Player ' + socket.username + ' disconnected: ' + reason);

    });

});

```

## Expected behavior
Everytime, no matter what, player will be kicked.

## Additional context
utils.js ->
```js
[...]

PR #1278, opened 2024-01-02
Add support for non-full packets

Pandapip1:
CC #811, #1094

Currently, this project seems extraordinarily dead, in that it isn't being updated.

Admittedly, this PR is a band-aid and not a fix. But I'd rather have the option for hacky non-0% support than forced 0% support, when it comes to any recent version of the game.


extremeheat:
What does this actually do?

Pandapip1:
> What does this actually do?

It adds a client / server option to have `createDeserializer` return a `Parser` object instead of a `FullPacketParser`.

extremeheat:
What does that do?

Pandapip1:
> What does that do?

`FullPacketParser` performs packet validity checks. New minecraft versions extend the existing protocol in ways that break those validitity checks, but where existing code can read 90% of the useful information. `node-minecraft-protocol` does need to be updated to read that remaining 10% that gets cut off, *but that doesn't have to break compatibility if those features aren't yet needed.*

extremeheat:
> extend the existing protocol in ways that break those validitity checks

Where in the code are you seeing this and for what version? Do you have a link to code somewhere?

Pandapip1:
This conversation is already well-documented at #811, except oddly I can't find MC version numbers. IIRC, >1.9.0

https://github.com/ProtoDef-io/node-protodef/blob/e008e688d52478d82926888a59d85d79704ff462/src/serializer.js#L76

extremeheat:
There is no such thing AFAIK of a partial packet. #811 is unrelated, that's caused by error in minecraft-data protocol, or invalid server data.

Pandapip1:
> There is no such thing AFAIK of a partial packet. https://github.com/PrismarineJS/node-minecraft-protocol/issues/811 is unrelated, that's caused by error in minecraft-data protocol, or invalid server data.

Well then, this allows clients to handle invalid server data, and vice versa.

I'll fully admit: I'm not particularly experienced with regards to the MC protocol or protobuf in general. But what I do know is that there was a server that vanilla MC clients were able to connect to and that node-minecraft-protocol (and by extension, mineflayer) clients were not able to connect to.

extremeheat:
I think you misunderstood my last comment. This PR implies there is a thing as a partial packet, like one that has variable length. Or something to do with packet batching. As far as I am aware, this does not exist. Actually I think this seems like a dupe PR of https://github.com/PrismarineJS/node-minecraft-protocol/pull/1094, what's changed here ?

As for #811, you are looking at two different errors. The deserializer (in node-minecraft-protocol) is failing to read something (not fatal) _and_ the serializer is encoding something that is being sent to a vanilla client, causing it to crash (which is fatal). These are two independent errors, and like I mention, caused by legitimate error in data from outside or a wrong definition in minecraft-data

Pandapip1:
> I think you misunderstood my last comment. This PR implies there is a thing as a partial packet, like one that has variable length. Or something to do with packet batching. As far as I am aware, this does not exist. Actually I think this seems like a dupe PR of https://github.com/PrismarineJS/node-minecraft-protocol/pull/1094, what's changed here ?

#1094 and this PR share a git branch, which is why their files are synced. #1094 was closed when https://github.com/PrismarineJS/node-minecraft-protocol/pull/1278/commits/7da26b49285179e7277ca7911706da3e40d05b84 was the latest commit. It fully disabled the `FullPacketParser` all the time.

This PR adds an option to disable them instead of having it disabled 100% of the time.

Pandapip1:
Never mind. I can't find the code that reproduces the error. Closing unless I can repro again.

extremeheat:
Those errors are caused by bad definitions, missing definitions, or illegal packets being sent by the server. This PR wanted to just remove sanity checks which is not good (does not fix anything)

Pandapip1:
> This PR wanted to just remove sanity checks which is not good (does not fix anything)

Correction: this PR added an option that would manually disable the sanity checks for debugging purposes.

rom1504:
The current code https://github.com/Pandapip1/node-minecraft-protocol/blob/master/src/transforms/serializer.js#L5 is always using FullPacketParser and hence always parses partial packets if possible 

So the only only thing this PR would have done is add an option to *not* parse partial packets and instead crash.

PR #1283, opened 2024-01-17
emit noAuth event

zardoy:
...so its possible to interrupt the client sending anything e.g.

```ts
client.on('noCredentials', () => {
    // possibly show message to user
    client.end()
})

```

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -274,6 +274,10 @@ Emitted after the player enters the PLAY protocol state and can send and recieve
 
 Called when an error occurs within the client. Takes an Error as parameter.
 
+### `noCredentials` event
+
+Emitted when client didn't supply credentials and server is in online mode (authenticating is required).
+
 ### `playerChat` event
 
 Called when a chat message from another player arrives. The emitted object contains:
--- a/src/client/encrypt.js
+++ b/src/client/encrypt.js
@@ -24,6 +24,7 @@ module.exports = function (client, options) {
       } else {
         if (packet.serverId !== '-') {
           debug('This server appears to be an online server and you are providing no password, the authentication will probably fail')
+          client.emit('noCredentials')
         }
         sendEncryptionKeyResponse()
       }
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -51,6 +51,7 @@ declare module 'minecraft-protocol' {
 		on(event: `raw.${string}`, handler: (buffer: Buffer, packetMeta: PacketMeta) => PromiseLike): this
 		on(event: 'playerChat', handler: (data: { formattedMessage: string, message: string, type: string, sender: string, senderName: string, senderTeam: string, verified?: boolean }) => PromiseLike): this
 		on(event: 'systemChat', handler: (data: { positionId: number, formattedMessage: string }) => PromiseLike): this
+		on(event: 'noCredentials', handler: () => PromiseLike): this
 		// Emitted after the player enters the PLAY state and can send and recieve game packets
 		on(event: 'playerJoin', handler: () => void): this
 		once(event: 'error', listener: (error: Error) => PromiseLike): this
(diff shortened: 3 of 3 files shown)


src/index.d.ts
```diff
@@ -51,6 +51,7 @@ declare module 'minecraft-protocol' {
 		on(event: `raw.${string}`, handler: (buffer: Buffer, packetMeta: PacketMeta) => PromiseLike): this
 		on(event: 'playerChat', handler: (data: { formattedMessage: string, message: string, type: string, sender: string, senderName: string, senderTeam: string, verified?: boolean }) => PromiseLike): this
 		on(event: 'systemChat', handler: (data: { positionId: number, formattedMessage: string }) => PromiseLike): this
+		on(event: 'noCredentials', handler: () => PromiseLike): this
```
zardoy, line 54: shouldnt it also use typed event emitter as mineflayer does?

src/client/encrypt.js
```diff
@@ -24,6 +24,7 @@ module.exports = function (client, options) {
       } else {
         if (packet.serverId !== '-') {
```
zardoy, line 25: btw looking at the code, what's the point of this check? so the response is sent anyway below...

zardoy:
btw let's add minimal esbuild web example here

src/client/encrypt.js
extremeheat: What is the purpose of your PR ? What does this do?

extremeheat: Should be added to the doc

docs/API.md
```diff
@@ -274,6 +274,10 @@ Emitted after the player enters the PLAY protocol state and can send and recieve
 
 Called when an error occurs within the client. Takes an Error as parameter.
 
+### `noCredentials` event
+
+Emitted when client didn't supply credentials and server is in online mode (authenticating is required).
```
zardoy, line 279: Im not sure if *appears* needs to be added here. Is it possible to encounter a case where `noCredentials` would be emitted and the connection is successful?

src/client/encrypt.js
zardoy: btw it was inspired by the pakkit patch, in my case it was also useful to have in the web app because I was not even getting an answer back from the server, so I was using this event to display an error message to the user

extremeheat: It's not generally a good idea to make a change to code without understanding what's actually happening.

The creation of this event implies there is actually some credential issue here, can you look into what's happening here?

docs/API.md
extremeheat: Do you have a link to nmp/vanilla code or somewhere to doc how this gets triggered? It don't understand how packet.serverId relates to auth.

src/client/encrypt.js
zardoy: > It's not generally a good idea to make a change to code without understanding what's actually happening.

Yes, you are right, I just saw the message in the debug console and decided to also put an event there. But doesn't it mean what it says?

> The creation of this event implies there is actually some credential issue here, can you look into what's happening here?

WDYM?
Do you mean there is an issue with my application instead?

zardoy: My goal is to end the connection as soon as I understand that the server is running online mode

docs/API.md
zardoy: >  I don't understand how packet.serverId relates to auth.

i don't know why there is packet.serverId check, sorry. That's why I raised the question below. Or did you mean something else? I wish we could improve clarity here. This is what I want to get: https://github.com/PrismarineJS/node-minecraft-protocol/pull/1283/files#r1454466037

extremeheat: Yes, I mean we should not be formally emitting something like "no credentials" without being certain there is some credential issue and what's causing this. Usually this is what unit tests are supposed to catch. I don't think we need tests for this change, but an understanding of from where this issue originates, not just some debug comment.

zardoy: > Yes, I mean we should not be formally emitting something like "no credentials" without being certain there

Ok, I see your point. I have only my personal experience: whenever I try to connect to any online server I get it.
Not sure what to do here for now..

rom1504:
the comments seem to indicate this PR is not the right fix. Please re-open if you want to find an other path @zardoy

PR #1284, opened 2024-01-26
Acknowledge returning to configuration state if in play state.

wgaylord:
node-minecraft-protocol fix for part of https://github.com/PrismarineJS/mineflayer/issues/3292 

Seems we missed this detail of the configuration protocol since the server never normally transitions back into the config state.

### DIFF
--- a/src/client/play.js
+++ b/src/client/play.js
@@ -49,7 +49,7 @@ module.exports = function (client, options) {
       if (client.state === states.CONFIGURATION) return
       // If we are returning to the configuration state from the play state, we ahve to acknowledge it.
       if (client.state === states.PLAY) {
-        client.write('configuation_acknowledged', {})
+        client.write('configuration_acknowledged', {})
       }
       client.state = states.CONFIGURATION
       // Server should send finish_configuration on its own right after sending the client a dimension codec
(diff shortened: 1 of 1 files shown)


extremeheat:
Where does `configuation_acknowledged` come from?

wgaylord:
It comes from wiki.vg and the MC source, its also a play state packet.   Also detailed here https://github.com/extremeheat/extracted_minecraft_data/blob/client1.20.2/client/net/minecraft/client/multiplayer/ClientPacketListener.java#L735-L757

extremeheat:
It's not in protocol.json, this will throw

wgaylord:
> It's not in protocol.json, this will throw

[It is in the protocol.json ](https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.20.2/protocol.json#L7001)

extremeheat:
I see, it's in 1.20.2 but not 1.20.3

wgaylord:
> I see, it's in 1.20.2 but not 1.20.3

Ah seems to be another case of a spelling mistake in 1.20.2 protocol.json

wgaylord:
Now have https://github.com/PrismarineJS/minecraft-data/pull/841 open to fix that mistake.

wgaylord:
Could we look into getting this merged, have had multiple people in the discord getting tripped up the the bug it fixes on multiplayer servers that use hubs.

extremeheat (approved):

Issue #1286, opened 2024-02-06
Some event handlers end up being called multiple times after going thru configuration state.


wgaylord:
[ ] The [FAQ](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - minecraft-protocol: #.#.#
 - server: paper 1.20.2
 - node: 20

## Detailed description of a problem

On servers with world switching that use the new configuration state, chat events end up called multiple times (once per number of transitions thru configuration state) this is due to `onReady` https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client/play.js#L64 being called even time we exit configuration state and re-registering the chat handler. 

I am unsure of a proper fix for this.

rom1504:
We need to emit a non ready event and properly unregister all relevant
listeners

PR #1287, opened 2024-02-06
Ensure `onReady` in client is called once

extremeheat:
Fix https://github.com/PrismarineJS/node-minecraft-protocol/issues/1286

Server sending start_config packets will incorrectly cause onReady to call multiple times


wgaylord:
Just tested and this does fix the issue.

PR #1291, opened 2024-03-07
Fix handling of disconnect in versionChecking on 1.20.3+. 

wgaylord:
This fixes the versionChecking code catching disconnects from states other then Handshake and Login, which cause errors on 1.20.3+ where NBT chat is used for the Play and Configuration state.

Since version checking only occurs in Handshake / Login state its best not to do anything when those other states occur.

This also includes a fix for the translate string having changed for when you have a version miss match.

### DIFF
--- a/src/client/versionChecking.js
+++ b/src/client/versionChecking.js
@@ -1,6 +1,9 @@
+const states = require('../states')
+
 module.exports = function (client, options) {
   client.on('disconnect', message => {
     if (!message.reason) { return }
+    if (client.state === states.PLAY | client.state === states.CONFIGURATION) { return }
     let parsed
     try {
       parsed = JSON.parse(message.reason)
@@ -11,7 +14,7 @@ module.exports = function (client, options) {
     let text = parsed.text ? parsed.text : parsed
     let versionRequired
 
-    if (text.translate && text.translate.startsWith('multiplayer.disconnect.outdated_')) { versionRequired = text.with[0] } else {
+    if (text.translate && (text.translate.startsWith('multiplayer.disconnect.outdated_') | text.translate.startsWith('multiplayer.disconnect.incompatible'))) { versionRequired = text.with[0] } else {
       if (text.extra) text = text.extra[0].text
       versionRequired = /(?:Outdated client! Please use|Outdated server! I'm still on) (.+)/.exec(text)
       versionRequired = versionRequired ? versionRequired[1] : null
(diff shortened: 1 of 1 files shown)


src/client/versionChecking.js
```diff
@@ -1,6 +1,9 @@
+const states = require('../states')
+
 module.exports = function (client, options) {
   client.on('disconnect', message => {
     if (!message.reason) { return }
+    if (client.state === states.PLAY | client.state === states.CONFIGURATION) { return }
```
extremeheat, line 6: Can you add a comment here explaining why?

src/client/versionChecking.js
```diff
@@ -11,7 +14,7 @@ module.exports = function (client, options) {
     let text = parsed.text ? parsed.text : parsed
     let versionRequired
 
-    if (text.translate && text.translate.startsWith('multiplayer.disconnect.outdated_')) { versionRequired = text.with[0] } else {
+    if (text.translate && (text.translate.startsWith('multiplayer.disconnect.outdated_') | text.translate.startsWith('multiplayer.disconnect.incompatible'))) { versionRequired = text.with[0] } else {
```
extremeheat, line 17: Can you un-inline this if statement?

src/client/versionChecking.js
wgaylord: Just documented it.

src/client/versionChecking.js
wgaylord: Just un-inlined the if. Not sure why it was inlined to begin with.

src/client/versionChecking.js
rom1504: why use bitwise or here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR

wgaylord: Fixed, didn't realize that was bitwise or.

src/client/versionChecking.js
```diff
@@ -1,6 +1,11 @@
+const states = require('../states')
+
 module.exports = function (client, options) {
   client.on('disconnect', message => {
     if (!message.reason) { return }
+    // Prevent the disconnect packet handler in the versionChecking code from triggering on PLAY or CONFIGURATION state disconnects
+    // Since version checking only happens during that HANDSHAKE / LOGIN state.
```
rom1504, line 7: why don't we want to handle this also in PLAY or CONFIGURATION states ?

rom1504: ok I see #3325

PR #1292, opened 2024-03-14
fix: wrap base64-encoded PEM with 64-char line boundary

littledivy:
According to [RFC7468](https://datatracker.ietf.org/doc/html/rfc7468):

> Generators MUST wrap the base64-encoded lines so that each line
  consists of exactly 64 characters except for the final line, which
  will encode the remainder of the data (within the 64-character line
  boundary), and they MUST NOT emit extraneous whitespace.

Parsers can avoid branching and prevent timing sidechannel attacks. Ref https://arxiv.org/pdf/2108.04600.pdf

Fixes compatibility with Deno as it enforces stricter handling of PEM.

### DIFF
--- a/src/client/encrypt.js
+++ b/src/client/encrypt.js
@@ -79,7 +79,7 @@ module.exports = function (client, options) {
 function mcPubKeyToPem (mcPubKeyBuffer) {
   let pem = '-----BEGIN PUBLIC KEY-----\n'
   let base64PubKey = mcPubKeyBuffer.toString('base64')
-  const maxLineLength = 65
+  const maxLineLength = 64
   while (base64PubKey.length > 0) {
     pem += base64PubKey.substring(0, maxLineLength) + '\n'
     base64PubKey = base64PubKey.substring(maxLineLength)


rom1504:
how did you check this improve something?

rom1504:
Hi @littledivy if nobody can verify that fix solve some issues, I think we should close this PR

ozaner:
Hey @rom1504 this solves Deno compatibility. Without this fix, the node-minecraft-protocol library would only work with the NodeJS runtime and not others that have properly implemented [RFC7468](https://datatracker.ietf.org/doc/html/rfc7468) (i.e. Deno). I've verified that this is a blocker towards that goal.

rom1504:
What example were you able to run with deno?

ozaner:
I see. Honestly, even though I didn't make this PR, I think it should be closed.

The main purpose of this PR was to enable Deno support, but since changing these widths seems to go against the actual Minecraft protocol, and since more work beyond changing these widths would probably need to be done to make it compatible with Deno, it doesn't seem worth it.

A full on proper port of this package is probably necessary for anyone wanting to use it with Deno (or better Deno-Node compatibility) which is definitely not in the scope of this project.

Thanks @extremeheat for helping to investigate this though.

extremeheat:
No, what I said was that only the client PEM encoder method needs to be changed, and that changing it here would have no effect on anything (beyond maybe fixing Deno). The width doesn't matter to nmp as long as it can be read by the crypto call.

On the server code there is nothing to change

ozaner:
As of the release of [Deno 2.0](https://deno.com/blog/v2.0), Deno now has full backwards compatibility with npm packages. I've verified that this package (and thus the packages that depend on it like mineflayer) does indeed work with Deno 2.0.

So I think we should finally close this PR now.

interrrp:
> As of the release of [Deno 2.0](https://deno.com/blog/v2.0), Deno now has full backwards compatibility with npm packages. I've verified that this package (and thus the packages that depend on it like mineflayer) does indeed work with Deno 2.0.
> 
> So I think we should finally close this PR now.

As of Deno 2.0.4, the issue arises yet again:

```sh
$ deno --version
deno 2.0.4 (stable, release, x86_64-pc-windows-msvc)
v8 12.9.202.13-rusty
typescript 5.6.2

$ deno task dev 
Task dev deno run -A --watch src/main.ts
Watcher Process started.
error: Uncaught Error: ASN.1 error: PEM error: PEM Base64 error: invalid Base64 encoding
```

ozaner:
@interrrp can you send your Deno code? I just tried it with Deno 2.0.4 on Minecraft 1.20.4 and it was able to join my local server just fine. Here is my code:

```js
import { createBot } from 'npm:mineflayer'

const bot = createBot({
  host: "localhost",
  port: 25565,
  username: 'Aiko',
});

bot.on('chat', (username, message) => {
  if (username === bot.username) return
  bot.chat(message)
})
```

ozaner:
Yeah honestly not sure what the problem is. I tried running it as a .ts file, using watcher mode instead of running directly, and it still works for me.

All I can think of is that you aren't on Deno 2.0 but you apparently are. Well, in any case, I think your [PR](https://github.com/denoland/deno/issues/26634) is a good change and would probably fix this.

Asguho:
Since the premise for closing this PR was that Deno had implemented full backward compatibility, and this is not the case, would it be possible to reopen it?

Nexus-00:
> i also get the same error, when trying the code in mc 1.20.4 and deno 2.1.1:
> 
> ```shell
> Uncaught Error: ASN.1 error: PEM error: PEM Base64 error: invalid Base64 encoding
>     at Object.publicEncrypt (ext:deno_node/internal/crypto/cipher.ts:221:10)
>     at sendEncryptionKeyResponse (file:///C:/Users/aske/Documents/GitHub/deno-mineflayer/node/node_modules/.deno/minecraft-protocol@1.50.0/node_modules/minecraft-protocol/src/client/encrypt.js:49:52)
>     at gotSharedSecret (file:///C:/Users/aske/Documents/GitHub/deno-mineflayer/node/node_modules/.deno/minecraft-protocol@1.50.0/node_modules/minecraft-protocol/src/client/encrypt.js:28:9)
>     at ext:deno_node/internal/crypto/_randomBytes.ts:42:9
>     at callback (ext:deno_web/02_timers.js:58:7)
>     at eventLoopTick (ext:core/01_core.js:210:13)
> ```

I can confirm that as of Deno version 2.2.4, the error remains.

littledivy:
@extremeheat Sorry for the delay, I have reverted the server change.

LunarLollipop:
`error: Uncaught Error: ASN.1 error: PEM error: PEM Base64 error: invalid Base64 encoding
    at Object.publicEncrypt (ext:deno_node/internal/crypto/cipher.ts:220:22)
    at sendEncryptionKeyResponse (file:///Users/luna/deno/node_modules/minecraft-protocol/src/client/encrypt.js:49:52)`
    
Issue still exists in deno 2.2.11. and minecraft_protocol 1.57.0

I tried manually change from `65 -> 64` in node_modules (locally) and this fixed issue with creating bot.

Is there a specific reason 65 is used? If no, I hope this will be merged soon.

CubeBeveled:
burh please merge this

i couldnt get nodejs to work so am using deno

PR #1294, opened 2024-03-27
Add onLoginPluginRequest callback in client Options

IceTank:


rom1504:
This breaks tests. Should we close it ?

Issue #1281, opened 2024-01-10
Inconsistent `playerChat` event data


Dinhero21:
- [x] The [FAQ](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - minecraft-protocol: 1.26.5

## Detailed description of a problem
[according to index.d.ts](https://github.com/PrismarineJS/node-minecraft-protocol/blob/21240f8ab2fd41c76f50b64e3b3a945f50b25b5e/src/index.d.ts#L52), `playerChat` event's data looks like this:
```ts
interface ServerPlayerMessage {
  formattedMessage: string
  message: string
  type: string
  sender: string
  senderName: string
  senderTeam: string
  verified?: boolean | undefined
}
```
but this is inconsistent [according to API.md](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/API.md#playerchat-event):
```md
Called when a chat message from another player arrives. The emitted object contains:
* formattedMessage -- (JSON) the chat message preformatted, if done on server side
* plainMessage -- (Plaintext) the chat message without formatting (for example no `<username> message` ; instead `message`), on version 1.19+
* unsignedContent -- (JSON) unsigned formatted chat contents ; should only be present when the message is modified and server has chat previews disabled - only on version 1.19 - 1.19.2
* type -- the message type - on 1.19, which format string to use to render message ; below, the place where the message is displayed (for example chat or action bar)
* sender -- the UUID of the player sending the message
* senderTeam -- scoreboard team of the player (pre 1.19)
* senderName -- Name of the sender
* targetName -- Name of the target (for outgoing commands like /tell). Only in 1.19.2+
* verified -- true if message is signed, false if not signed, undefined on versions prior to 1.19
```
from testing (on `1.19.4`), I found that it looks something more like this:
```ts
interface ServerPlayerMessage {
  formattedMessage?: string
  plainMessage?: string
  unsignedContent: string
  type: number
  sender: UUID
  // no senderTeam (since only pre-1.19)
  senderName: string
  targetName: string | undefined
  verified: boolean
}
```

## Expected behavior
[...]

Issue #1290, opened 2024-03-06
versionChecking.js Bug


obseletecode72:
[X] The [FAQ](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - minecraft-protocol: latest
 - server: paper 1.20.4
 - node: #.#.#

## Detailed description of a problem
just tried to join normally in the server and throwed an exception

## Expected behavior
join normally in the server

## Additional context
´´´
SyntaxError: "[object Object]" is not valid JSON
    at JSON.parse (<anonymous>)
    at Client.<anonymous> (C:\Users\Administrator\Desktop\mine\node_modules\minecraft-protocol\src\client\versionChecking.js:6:21)
    at Client.emit (node:events:526:35)
    at emitPacket (C:\Users\Administrator\Desktop\mine\node_modules\minecraft-protocol\src\client.js:83:12)
    at FullPacketParser.<anonymous> (C:\Users\Administrator\Desktop\mine\node_modules\minecraft-protocol\src\client.js:112:9)
    at FullPacketParser.emit (node:events:514:28)
    at addChunk (C:\Users\Administrator\Desktop\mine\node_modules\protodef\node_modules\readable-stream\lib\_stream_readable.js:279:12)
    at readableAddChunk (C:\Users\Administrator\Desktop\mine\node_modules\protodef\node_modules\readable-stream\lib\_stream_readable.js:262:11)
    at Readable.push (C:\Users\Administrator\Desktop\mine\node_modules\protodef\node_modules\readable-stream\lib\_stream_readable.js:228:10)
    at Transform.push (C:\Users\Administrator\Desktop\mine\node_modules\protodef\node_modules\readable-stream\lib\_stream_transform.js:132:32)
SyntaxError: "[object Object]" is not valid JSON
    at JSON.parse (<anonymous>)
    at Client.<anonymous> (C:\Users\Administrator\Desktop\mine\node_modules\minecraft-protocol\src\client\versionChecking.js:6:21)
    at Client.emit (node:events:526:35)
    at emitPacket (C:\Users\Administrator\Desktop\mine\node_modules\minecraft-protocol\src\client.js:83:12)
    at FullPacketParser.<anonymous> (C:\Users\Administrator\Desktop\mine\node_modules\minecraft-protocol\src\client.js:112:9)
    at FullPacketParser.emit (node:events:514:28)
[...]

PR #1303, opened 2024-05-26
Fix realms loading issue due to createClient plugin init order

### LINKED ISSUE #1302: pickRealm silently fails to connect
```
'use strict'

const mineflayer = require('mineflayer')

const [,, username, realmName] = process.argv
if (!realmName) {
  console.log('Usage : node client_realms.js <username/email> <realm_name>')
  process.exit(1)
}

const client = mineflayer.createBot({
  realms: {
    pickRealm: (realms) => realms.find(e => e.name === realmName) // Connect the client to a Realm using a function that returns a Realm
  },
  username,
  auth: 'microsoft' // This option must be present and set to 'microsoft' to  join a Realm.
})

client.on('connect', function () {
  console.info('connected')
})

client.on('disconnect', function (packet) {
  console.log('disconnected: ' + packet.reason)
})
```

prints  connected, but never actually connects to realm.

extremeheat:
* Fix loading plugins before realm host/port have been obtained
  * Fix #1302 
* Fix https://github.com/PrismarineJS/node-minecraft-protocol/issues/1281 types

### DIFF
--- a/src/client/versionChecking.js
+++ b/src/client/versionChecking.js
@@ -10,7 +10,6 @@ module.exports = function (client, options) {
     try {
       parsed = JSON.parse(message.reason)
     } catch (error) {
-      client.emit('error', error)
       return
     }
     let text = parsed.text ? parsed.text : parsed
--- a/src/createClient.js
+++ b/src/createClient.js
@@ -39,17 +39,20 @@ function createClient (options) {
   tcpDns(client, options)
   if (options.auth instanceof Function) {
     options.auth(client, options)
+    onReady()
   } else {
     switch (options.auth) {
       case 'mojang':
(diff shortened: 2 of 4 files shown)


src/client/versionChecking.js
```diff
@@ -10,7 +10,6 @@ module.exports = function (client, options) {
     try {
       parsed = JSON.parse(message.reason)
     } catch (error) {
-      client.emit('error', error)
```
rom1504, line 13: Why do we want to ignore this error?

extremeheat: Related to https://github.com/PrismarineJS/node-minecraft-protocol/issues/1290 

The code below checks if the kick message contains an outdated version message, but it's essential. And since the `reason` field is actually a JSON chat message it's still valid for the server to send a plain string (ie "hello world" as opposed to {"text":"hello world"}) as opposed to JSON, as the chat parsing code will accept it. So to correctly parse it we would have to import prismarine-chat and parse it with that.

PR #1323, opened 2024-08-30
Supply MSA bearer token directly

SinanAkkoyun:
This PR enables one to supply the MSA bearer token directly (which simplifies a lot for users who do not want to rely on the cache system)

Dependent on: https://github.com/PrismarineJS/prismarine-auth/pull/108

Enables:
```js
const bot = mineflayer.createBot({
  ...
  auth: 'microsoft',
  msaToken: 'abcd...'
})
```


SinanAkkoyun:
I just realized that the msa token goes invalid after a day or so, so this PR can be closed. I would, however, be very thankful for help on using the createBot session and access token parameters as I was not able to get that working

SinanAkkoyun:
I implemented the auth flow for refreshing those tokens (Essential Mod auth flow) and would like to publish and use that code. For that, I would appreciate if this PR could be merged

rom1504:
thoughts @extremeheat ?

extremeheat:
This doesn't seem like a good idea to me. The auth system is designed to internally walk up the dep tree of required tokens refreshing and manage the tokens on its own as needed. Trying to inject things into that process is not good. Not using the cache system should be implemented correctly with a custom cacher.

PR #1324, opened 2024-08-30
fix(server): fix for login_acknowledged on sync communication

zardoy:
It works fine but it feels not correct. We could see it was breaking our testing setup along with the client part of minecraft-protocol, since we are using sync communication and the success packet emits `login_acknowledged` on the client side and that event is awaited only after the success event on the server side.


zardoy:
isee no interest

PR #1351, opened 2024-11-27
Fix registryEntryHolderSet read code

GroobleDierne:


rom1504:
Looks correct

rom1504:
We should have tests

PR #1353, opened 2024-12-03
Fix test for 1.21.3

GroobleDierne:
~~Still WIP~~ GTG


GroobleDierne:
Changing test worlds to superflat reduces test duration from 12 minutes to 8 on my machine.

PR #1356, opened 2024-12-11
Fix the import error

kzyqq00-Player:


extremeheat:
is this correct? Have you tested this to work? Doesn't seem right to be returning value of `require('node-rsa')` as a variable type.

PR #1360, opened 2024-12-28
Bumped prismarine-auth dependency

TatoExp:
Bumps prismarine-auth dependency:
- Removes jose dependency
- fixes support for bun

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "minecraft-folder-path": "^1.2.0",
     "node-fetch": "^2.6.1",
     "node-rsa": "^0.4.2",
-    "prismarine-auth": "^2.2.0",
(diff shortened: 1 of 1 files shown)


TatoExp:
Could the 1.21.1 pipeline possibly be reran manually, I'm not sure anything changed here should be causing the error that is happening. Also I don't have the correct JDK installed currently to run it locally

extremeheat:
There's no need to do this, the caret (^) prefix will automatically install the latest version of the package for the specified major version on clean install and when you run `npm update` (which should update all the deps at all the levels)

See https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004

PR #1373, opened 2025-02-12
fix: don't inflate/decompress if compressionThreshold is negative

cxvqo:
follows the notchian client better as if the compressionThreshold is negative, it means that the packets are uncompressed, and it should treat them accordingly

### DIFF
--- a/src/transforms/compression.js
+++ b/src/transforms/compression.js
@@ -51,6 +51,12 @@ class Decompressor extends Transform {
     if (value === 0) {
       this.push(chunk.slice(size))
       return cb()
+    } else if (this.compressionThreshold < 0) {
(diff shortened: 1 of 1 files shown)


extremeheat:
Is this the case on all versions 1.8+?

cxvqo:
the packets for controlling the compressionThreshold were added in 14w28a, in older versions this shouldn't cause any issues because the compressionThreshold should stay the same

cxvqo:
there may be an issue

cxvqo:
there isnt.

extremeheat:
Why close?

cxvqo:
> Why close?

Issue with my implementation of it. (I think)

PR #1383, opened 2025-03-24
Fix select_known_packs registered multiple times

ryanalexander:
select_known_packs was registering a listener each time the client went into configuration state. Causes issues with Bungeecord.

This fixes people complaining about 1.21.1 bungeecord & velocity issues


ryanalexander:
Just noticed this is a duplicate of https://github.com/PrismarineJS/node-minecraft-protocol/pull/1381
Closing

rom1504:
If you have details on why this is correct or even could add a non regression test it would be useful

ryanalexander:
> If you have details on why this is correct or even could add a non regression test it would be useful 

"If multiple sperm fertilize an egg (a condition called polyspermy), it typically leads to a non-viable embryo with an abnormal number of chromosomes, usually resulting in the development halting"

Similar occurs with an event listener. If multiple register, the response fires multiple times halting the connection.

If you need the definition of on or once, google exists

GenerelSchwerz:
If it helps, there was a discussion on this topic [found here](https://discord.com/channels/413438066984747026/743199746876768318/1350930343770980453). (main discord server) https://discord.gg/YKG6fdKYbB

extremeheat:
Issue is enterConfigState gets called multiple times without unregistering the previous `.on` so `.once` does fix

Looks like this was already fixed with #1381

PR #1402, opened 2025-06-13
Update play.js (fix not able to join hypixel)

Henry-Santa:
Add the missing packet. Vanilla client sends it, although optional, many servers block bots via not sending this.


182exe:
merge merge merge merge merge

rom1504:
fails ci and old

PR #1412, opened 2025-07-16
Chatjs: Spread into array

Squidysquid1:
I was using Mineflayer and got the error when using a chat event:
```
TypeError [ERR_INVALID_ARG_TYPE]: The "list" argument must be an instance of Array. Received an instance of Buffer
    at Buffer.concat (node:buffer:580:3)
    at updateAndValidateSession (./node_modules/minecraft-protocol/src/client/chat.js:61:36)
```

this fixed my issue on MC 1.21.4 
Was using latest mineflayer on node 24.4.0

Not too familiar with js/node environment, but hope this (very small) PR helps!

### DIFF
--- a/src/client/chat.js
+++ b/src/client/chat.js
@@ -58,7 +58,7 @@ module.exports = function (client, options) {
 
       if (player.hasChainIntegrity) {
         const length = Buffer.byteLength(message, 'utf8')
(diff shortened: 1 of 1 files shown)


Squidysquid1:
Looks like still having an issue when more than 1 message is sent in chat. Was only having 1 message sent when bot online. Will close until I find whats actually causing it to be broken.

PR #1485, opened 2026-05-02
Fix computeChatChecksum returning unsigned byte for i8 schema (closes #1482)

### LINKED ISSUE #1482: computeChatChecksum returns unsigned byte but schema is i8 → RangeError on MC 1.21.11
### Versions

- `node-minecraft-protocol`: 1.66.0
- `mineflayer`: 4.37.0
- Minecraft server: 1.21.11 (vanilla single-player, opened to LAN)
- Node: 22

### What happens

When the bot attempts to send any slash command whose packet goes through `chat_command_signed` (e.g. `/say hello`, `/tell target msg`), it crashes with:

```
RangeError [ERR_OUT_OF_RANGE]: Write error for undefined :
  The value of "value" is out of range. It must be >= -128 and <= 127. Received 143
    at writeU_Int8 (node:internal/buffer:740:11)
    at Buffer.writeInt8 (node:internal/buffer:871:10)
    at Object.writer [as i8] (.../protodef/src/datatypes/numeric.js:97:25)
    at Object.packet_chat_command_signed ...
  code: 'ERR_OUT_OF_RANGE',
  field: 'play.toServer'
```

The value received varies (143, 212, ...) — it's the chat checksum.

### Root cause

In `src/datatypes/checksums.js`:

```js
function computeChatChecksum (lastSeenMessages) {
  // ...
  const result = checksum & 0xff        // <-- returns 0..255 (unsigned)
  return result === 0 ? 1 : result
}
```

The function returns an **unsigned** byte (`& 0xff` → `0..255`).

The schema for `packet_chat_command_signed` (and `packet_chat_message`) in 1.21.5+ defines this field as a **signed** `i8` (`-128..127`):

```json
{ "name": "checksum", "type": "i8" }
```

Whenever the hash happens to have its top bit set (~50% of the time), the resulting unsigned byte is ≥ 128, and `writeInt8` throws.

### Suggested fix

```js
const unsigned = checksum & 0xff
const signed = unsigned > 127 ? unsigned - 256 : unsigned
return signed === 0 ? 1 : signed
```

or equivalently `new Int8Array([checksum & 0xff])[0]`.

Applied this as a local monkey-patch against `node_modules/minecraft-protocol/src/datatypes/checksums.js` and the bot can now reliably send slash commands to 1.21.11 without the `ERR_OUT_OF_RANGE` crash.

Happy to open a PR if that would help.

mio-rei:
## Summary

The `chat_command_signed` packet schema defines the checksum as `i8` (-128..127), but `computeChatChecksum` was returning the result of `& 0xff` (0..255). When the masked checksum exceeded 127, the protocol writer rejected it with `RangeError: The value of \"value\" is out of range. It must be >= -128 and <= 127. Received N`, breaking any signed slash command on MC 1.21.11.

Closes #1482.

## Fix

Convert the unsigned byte to signed range before returning:

```js
const unsigned = checksum & 0xff
const signed = unsigned > 127 ? unsigned - 256 : unsigned
return signed === 0 ? 1 : signed
```

## Test plan

- [x] Locally patched `node_modules/minecraft-protocol/src/datatypes/checksums.js` with the same conversion against MC 1.21.11 vanilla server (LAN), bot has been sending `/say` / `/tell` for ~14 days without RangeError.
- [ ] Adding a unit test to the repo would be nice — happy to follow up if reviewers want one in this PR.

## Notes

Issue #1482 has the full repro steps (versions, error, root cause). This PR is the minimal fix; the value-zero guard (`return 1`) is preserved.

### DIFF
--- a/src/datatypes/checksums.js
+++ b/src/datatypes/checksums.js
@@ -12,9 +12,11 @@ function computeChatChecksum (lastSeenMessages) {
       checksum = (31 * checksum + sigHash) & 0xffffffff
     }
   }
-  // Convert to byte
-  const result = checksum & 0xff
-  return result === 0 ? 1 : result
+  // Convert to signed byte (i8: -128..127) to match the chat_command_signed packet schema.
+  // The previous `& 0xff` produced 0..255 which causes RangeError when value > 127.
+  const unsigned = checksum & 0xff
+  const signed = unsigned > 127 ? unsigned - 256 : unsigned
+  return signed === 0 ? 1 : signed
 }
 
 module.exports = { computeChatChecksum }


extremeheat (approved):

extremeheat:
This is actually incorrect, checksums should not be signed, the checksum occupies the full integer width

Fixed in https://github.com/PrismarineJS/minecraft-data/pull/1188

PR #1517, opened 2026-08-30
Load auth and server modules only when used

u9g:
## Problem

`createClient.js` requires the Microsoft auth dependency tree (prismarine-auth → @azure/msal-node → jsonwebtoken, plus xboxlive-auth and node-fetch) at module load for every consumer, though only `auth: 'microsoft'` clients ever call into it, and `index.js` loads `createServer` (node-rsa and the server plugins) for client-only consumers.

Profiled with a `Module._load` timing hook: this is ~70ms of the ~275ms first `require('minecraft-protocol')` on an M-series Mac with a cold-ish file cache (~10ms once the OS file cache is warm — msal-node is one large bundled file, so the win is concentrated where startup is already slowest). It also keeps a few MB of never-used code out of memory for offline bots.

## Change

- `createClient`: require `mojangAuth`/`microsoftAuth` inside their `switch` branches, so they load on first authenticated connect.
- `index.js`: export `createServer`/`Server` through getters, so client-only consumers never load the server tree. Destructuring consumers (`const { createServer } = require('minecraft-protocol')`) still work — the getter fires at destructure time.

No behavior change for any auth mode; the modules load at the same points they were first used before.

## Verification

standard clean; packet + cycle tests pass locally; offline-mode client connect exercised against a vanilla server via the mineflayer test suite.


u9g:
Closing: re-measured on an idle machine the win is only ~10ms warm / ~65ms cold-cache — not worth the churn.

PR #1521, opened 2026-09-06
checksum last seen messages in acknowledgement order

Pix3lPirat3:
Use the circular buffer's logical traversal for both acknowledgement bits and the checksum so wrapped chat histories stay synchronized with the server.


test/chatAcknowledgementTest.js
```diff
@@ -0,0 +1,57 @@
+/* eslint-env mocha */
```
extremeheat, line 1: This test would be more useful if it tested against a real server. Or at least a test against node-minecraft-protocol's server and its client

extremeheat (commented):
Cool, did you test against the server side impl also? We implement both sides

Pix3lPirat3:
Added a real server test as you suggested: "accepts valid and rejects invalid wrapped acknowledgements between a protocol client and server" spins up mc.createServer and mc.createClient, has the client send signed chat, and checks the server both accepts a valid wrapped acknowledgement checksum and rejects a tampered one. It runs across 1.21.9, 1.21.11 and 26.1 alongside the existing checksum-ordering unit tests (18 passing). The same follow-up commit also fixed the last-seen ordering logic the unit test caught.

rom1504 (commented):

The updated test now addresses the requested client/server coverage: I ran all 18 acknowledgement tests successfully, including the real local NMP connection accepting wrapped acknowledgements and rejecting a tampered checksum. I also traced the shared acknowledgement order through both command/chat construction and the server's pending-message window. No additional blocker found in this review of c4da904; I did not run a vanilla server or the full repository suite.

Skills used: `prismarine-protocol-data-review` checked both production endpoints and the negative control; `prismarine-review` checked that the later implementation addresses extremeheat's original fixture request instead of repeating it.

PR #1522, opened 2026-09-06
Match the vanilla client in the login/configuration handshake: cookies, brand, per-login chat session, known packs

u9g:
Audit of nmp's client against the vanilla 1.21.4 client (`ClientHandshakePacketListenerImpl`, `ClientCommonPacketListenerImpl`, `ClientPacketListener`, `KnownPacksManager`, `Options`) turned up four places where nmp diverges in ways that servers and proxies notice. One commit each:

- **Answer `cookie_request` (1.20.5+).** nmp never replied, so any server or proxy that sends one (Velocity/Bungee modern forwarding, transfer flows) hangs the connection. A new `src/client/cookies.js` keeps `client._cookies`, fills it from `store_cookie`, and answers `cookie_request` in login, configuration and play with the stored value or an absent one, exactly like vanilla. A `cookies` option seeds the map so a consumer following a `transfer` packet can carry cookies over like vanilla's `TransferState`.
- **Brand then Client Information, once.** Vanilla answers `login_acknowledged` with `custom_payload minecraft:brand` followed by Client Information, and does not repeat either when a server sends it back to configuration (only `configuration_acknowledged`). nmp sent no brand, re-sent the settings on every configuration entry, and used view distance 10 where vanilla's default is 12. New `brand` option (default `'vanilla'`; mineflayer's `options.brand` flows through), `clientSettings` defaults are now the vanilla ones.
- **Fresh chat session on every `login` packet.** After a proxy server switch the second `login` packet left the old session in place; vanilla discards it and announces a new session (new UUID, index 0) with the same key pair. `client.once('login')` → `client.on('login')`.
- **`select_known_packs`.** Vanilla replies with the packs from the server's list it has locally. New `knownPacks` option (`[{ namespace, id, version }]`) answered with the intersection; the default stays `[]` so servers keep sending full registry data, which mineflayer/prismarine-registry rely on.

Docs updated in `docs/API.md`. In-process tests added in `test/serverTest.js` for cookie replies in all three states (including a stored-cookie echo and an absent reply), brand-then-settings sent exactly once across a `start_configuration` round trip, and the known-packs intersection.

Note: minecraft-data's 1.21.8 `protocol.json` declares `cookie_response.value` as a non-optional `ByteArray` (the 1.21.8 client writes it nullable like every other version). The absent reply is still wire-correct there (an empty `ByteArray` and an absent option both serialize as a single `0x00`), but echoing a stored cookie needs a minecraft-data fix; the stored-cookie test is gated on the schema having the option wrapper.


docs/API.md
```diff
@@ -147,16 +147,19 @@ Returns a `Client` instance and perform login.
  * id : a numeric client id used for referring to multiple clients in a server
  * validateChannelProtocol (optional) : whether or not to enable protocol validation for custom protocols using plugin channels. Defaults to true
  * disableChatSigning (optional) : Don't try obtaining chat signing keys from Mojang (1.19+)
- * clientSettings (optional) : Client Information (settings) sent to the server during the configuration phase (1.20.2+). All fields are optional and default to vanilla-safe values:
+ * cookies (optional) : cookies to answer `cookie_request` packets with, as an object or Map of key to Buffer (1.20.5+). Pass the previous connection's `client._cookies` when following a `transfer` packet, like the vanilla client does
```
rom1504, line 150: **Astra agent review — AI-generated, not manually written by the maintainer.**

The newly documented options are missing from `src/index.d.ts`: compiling `createClient({ username: 'bot', brand: 'custom' })` against this head gives TS2353, and the same happens for `cookies` and `knownPacks`. The documented transfer example also gives TS2339 for `previousClient._cookies`. Could this change add those three options to `ClientOptions` and declare the cookie map used by this documented transfer path? This lets TypeScript consumers use the new API without casts.

Skills used: prismarine-code-quality-review helped check the documented public API against its declarations; prismarine-review helped verify the current head and avoid duplicate feedback.

rom1504 (commented):

I checked the current handshake changes and public declarations, ran the 40 focused handshake tests successfully, and reproduced the TypeScript errors below against this head. The existing cookie-schema limitation is already documented and is not repeated here. No vanilla server or full suite was run.

Skills used: prismarine-code-quality-review helped check the documented public API against its declarations; prismarine-review helped verify the current head and avoid duplicate feedback.

docs/API.md
u9g: Added `brand`, `cookies` and `knownPacks` (with a `KnownPack` interface) to `ClientOptions` and declared `client._cookies: Map<string, Buffer>` in 96c837c. Your `createClient({ username: 'bot', brand: 'custom' })` and the `cookies: previousClient._cookies` transfer example now compile under `--strict`.

rom1504 (commented):

The previous declaration finding is fixed: the documented brand, object/Map cookies, knownPacks and previousClient._cookies examples compile with strict TypeScript against this head. I also reran all 40 focused handshake tests successfully; current CI is green. **Still conditional for the full advertised cookie behavior:** the PR body documents that nonempty stored-cookie replies on 1.21.8 require the minecraft-data optional-value schema fix, and that test is skipped there. The type repair is ready, but I would coordinate that data fix/release before treating the whole PR as ready. No new inline defect; this is the existing documented limitation.

Skills used: prismarine-architecture-review verified the real candidate declarations; prismarine-lifecycle-action-review traced configuration re-entry; prismarine-protocol-data-review kept skipped cookie-schema coverage distinct from passing versions. prismarine-review checked current discussions and avoided duplicate findings.

PR #1523, opened 2026-09-06
fix: write the 0 holder id for inline registryEntryHolder entries

u9g:
Constraints on the compiled `registryEntryHolder` writer:

- An inline entry is written as varint `0` followed by the entry; every byte the writer advances over is written.
- A registry entry is written as varint `id + 1`.
- `sizeOf` and `write` agree on the length in both cases, and the reader parses either back to the same value.

Test (`test/registryEntryHolderTest.js`): a 1.21.4 `Slot` whose `consumable` component carries an inline sound is written into a `0xff`-filled buffer; the holder id byte must be `0`, the round trip must parse to the same value with the same size, and the registry case must encode `id + 1`. Fails on master, passes with the change.

### DIFF
--- a/src/datatypes/compiler-minecraft.js
+++ b/src/datatypes/compiler-minecraft.js
@@ -111,7 +111,7 @@ if (n !== 0) {
 if (${baseName} != null) {
   offset = ${compiler.callType(`${baseName} + 1`, 'varint')}
 } else if (${otherwiseName}) {
-  offset += 1
+  offset = ${compiler.callType(0, 'varint')}
   offset = ${compiler.callType(`${otherwiseName}`, opts.otherwise.type)}
 } else {
   throw new Error('registryEntryHolder type requires "${baseName}" or "${otherwiseName}" fields to be set')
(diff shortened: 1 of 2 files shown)


rom1504:
Why varint ? Where does it come from ?

Pix3lPirat3:
The leading holder value is VarInt-encoded. The existing reader already reads a VarInt: 0 means an inline value, while n > 0 means registry ID n - 1. The registry writer also writes id + 1 as a VarInt. The inline branch was only advancing one byte without writing the required 0, so this makes it mirror the reader. [MCProtocolLib uses the same encoding.](https://github.com/GeyserMC/MCProtocolLib/blob/19783c29ece24bc3f07f8ff08628549527e3de20/protocol/src/main/java/org/geysermc/mcprotocollib/protocol/codec/MinecraftTypes.java#L302-L313)

extremeheat:
varint 0 is no different than u8 0. The increment operation by 1 assumes the buffer is already zero initialized, so this could help on buffers made from Buffer.allocUnsafe

PR #1526, opened 2026-09-07
test: select the version under test with MC_VERSION instead of a title grep

u9g:
Invariants of the test run:

- `MC_VERSION` names the one supported version the version-parameterised suites run against; unset runs all of them.
- A `MC_VERSION` that is not in `supportedVersions` fails the run instead of quietly testing nothing.
- Every suite runs in every CI job, whether or not its title carries a version.

CI ran `mochaTest -- -g <version>v`, so a suite only ran if its title carried a Minecraft version. `declare_commands handling`, `lpVec3` and `ping` have never run in any job.

Depends on #1527: `declare_commands handling` is the suite this filter was hiding, and it fails until that tree fix lands.

Checked locally: `MC_VERSION=1.21.4` selects the 1.21.4 suites plus the three unversioned ones, `MC_VERSION=9.9.9` fails with the supported list, and `lpVec3` and `ping` pass.


rom1504:
Change readme to mention this then you can merge

Issue #51, opened 2013-04-06
does it work with node0.10.0 ?


rom1504:
`npm install minecraft-protocol` get stuck in `npm http 304 https://registry.npmjs.org/callsite`
with npm 1.2.14

PR #1528, opened 2026-09-11
Fix custom protocol isolation and 1.7.10 auto-version selection

Pix3lPirat3:
Isolate custom packet schemas between clients without mutating shared protocol data. Validate auto-version candidates against their actual protocol number to correctly select Minecraft 1.7.10. Add regression tests for both fixes. (relative to node-minecraft-protocol-forge PR)

  Related PR:
  - Consumer: https://github.com/PrismarineJS/node-minecraft-protocol-forge/pull/51


rom1504 (commented):

The reviewed changes look ready from this focused review of 633aae1. All 36 added regression cases pass. I additionally exercised an actual local 1.7.10 NMP status/login exchange with `version: false`, which selected 1.7.10 and reached playerJoin. Forge #51's full suite also passes with this candidate dependency (103 tests), including independent registry mappings and repeated configuration cycles.

Validation limit: an additional interpreted-codec control is blocked by the pre-existing `nbt.addTypesToInterperter` call, which is absent from the installed prismarine-nbt export; this PR does not introduce that call. The production compiled path and the targeted tests passed. No vanilla/Forge Java server was run in this review.

Skills used: `prismarine-protocol-data-review` checked selected protocol data and custom-schema isolation through the consumer; `prismarine-architecture-review` checked that shared data stays unchanged while vanilla caching remains available; `prismarine-review` kept the existing interpreter limitation separate from regressions introduced here.

